PR #1, opened 2015-03-23
Add udp support to nfs service definition

zatricky:
- Added udp port
- Adjusted description appropriately.

### DIFF
--- a/config/services/nfs.xml
+++ b/config/services/nfs.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <service>
   <short>NFS4</short>
-  <description>The NFS4 protocol is used to share files via TCP networking. You will need to have the NFS tools installed and properly configure your NFS server for this option to be useful.</description>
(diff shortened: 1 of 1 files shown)


zatricky:
NFS4 uses tcp by default as a client - but it does listen on both udp and tcp as a server (by default).
Server-side, the same applies to NFS3. It doesn't make sense to separate the service definitions when both service versions are in fact listening on both.

I've also noted that related services (mountd, rpc-bind) have both tcp and udp defined in their definitions.

Perhaps instead the service definition needs to change to reflect that it is not specific to NFSv4?

The most sensible courses to resolve the issue would be one of the following:
a) Genericise the current nfs definition (perhaps change short from "NFS4" to "NFS" or "NFS3/4"
b) Specificate the nfs definition into "NFS3/4 for NFSv3 clients" and "NFS3/4 for NFSv4 clients" (I'm not sure if the same should then happen with mountd and rpc-bind).

As the more KISS approach, a) makes more sense to me. But perhaps there are good reasons to consider b) ?

t-woerner:
Closing due to last comment from zatricky.

PR #4, opened 2015-04-04
Add two missing ICMP types

dawud:
The following ICMP types are not included by default in the available
configuration:

Type 13 - Timestamp Request [1]
Type 14 - Timestamp Request [1]

This is useful to address [2].

[1] http://tools.ietf.org/html/rfc792
[2] https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-1999-0524


t-woerner:
Thanks

PR #18, opened 2015-05-13
Travis CI configuration file

jpopelka:
This commit adds .travis.yml - Travis CI configuration file.

There's a problem, that Travis runs Ubuntu-12.04 with autoconf-2.68, while we need autoconf-2.69, so it doesn't work so far. But I'm adding it, so it doesn't get lost.

I tried
http://askubuntu.com/questions/420991/install-autoconf-2-69-on-ubuntu-12-04
to work-around the problem but with no luck.

You can see (unsuccessful) runs here:
https://travis-ci.org/jpopelka/firewalld

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+python:
+  - "2.7"
+  - "3.4"
+before_install:
+#  - "sudo add-apt-repository ppa:costamagnagianfranco/autoconf -y"
+  - "sudo apt-get update -qq"
+# dependencies
+install:
+  - "sudo apt-get install -y intltool desktop-file-utils xsltproc"
+#  - "sudo apt-get install -y python-dbus python3-dbus python-decorator python-gobject python3-gi"
+# command to run tests
+script: ./autogen.sh && make && make check && make distcheck
+notifications:
+  email: false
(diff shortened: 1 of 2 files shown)


t-woerner:
There is a ``AC_PREREQ([2.69])`` in configure.ac. There has been a reason to go to 2.69, but I do not remember what it was. I think something with docs. From a quick look at the changelog of 2.69 I think that autoconf-2.68 should be working, but this needs to be verified.

About .travis.yml: I only see the missing installation of the python3-decorator package.

jpopelka:
Thanks. Yes, it (autogen.sh) seems to work OK with autoconf-2.68.
Another problem I've hit now is failing xsltproc probably due to not having access to outside world:

xsltproc -o ../man/man1/firewall-applet.1 --nonet --xinclude  transform-man.xsl firewall-applet.xml
I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl
compilation error: file transform-man.xsl line 7 element import

https://travis-ci.org/jpopelka/firewalld/jobs/63130534

kiorky:
more an apparmor thing or such, AFAIK net is not blocked from outside to inside on travis (i use it a lot)

jpopelka:
Pulling this in doesn't make much sense until we make it work. Otherwise Travis would claim that the tests failed after each push. So I'm closing this PR for now and I'll open another one if I'm able to make it work later.

PR #22, opened 2015-05-21
add ceph services

ktdreyer:
Ceph's OSD and MDS daemons listen on TCP ports 6800 through 7300 by default.

Ceph's MON daemon listens on port 6789 by default.

This PR adds the "ceph" and "ceph-mon" service definitions for Ceph.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -75,6 +75,8 @@ EXTRA_DIST = \
 	services/amanda-k5-client.xml \
 	services/bacula-client.xml \
 	services/bacula.xml \
+	services/ceph.xml \
+	services/ceph-mon.xml \
 	services/dhcpv6-client.xml \
 	services/dhcpv6.xml \
 	services/dhcp.xml \
--- a/config/services/ceph-mon.xml
+++ b/config/services/ceph-mon.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>ceph-mon</short>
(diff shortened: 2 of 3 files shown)


t-woerner:
The port range 6800 to 7300 is huge, is there a way to split it up for example for use cases?

ktdreyer:
Ceph can run a large number of Object Storage Daemons (the recommendation is to run one per block device). These OSDs can listen on up to three ports each.

Older Ceph docs recommended that a user calculate the number of daemons that are running, and open a number of ports that corresponded to the number of daemons (three for each OSD, one for each MDS).

Unfortunately this doesn't cover all use cases. Users can easily restart daemons in ways that cause the daemons to bind to higher ports. This leads to OSDs or MDSs binding to ports that are firewalled. See [rhbz 1219818](https://bugzilla.redhat.com/1219818) and [rhbz 1219493](https://bugzilla.redhat.com/1219493) for examples.

It is not easy to split these rules up further (based on use-case or something else) because Ceph's OSD and MDS port selection behavior is non-deterministic. Ceph really does expect to be able to bind to any of those ports.

As @linuxkidd mentioned on IRC today, these port openings are only required on the Ceph storage nodes, and not on ceph clients. The Ceph service nodes are typically isolated on their own Storage (public) and Internal (cluster) and Management networks, so the port openings are not usually seen as a huge problem. In an event of compromise to the Ceph services nodes, the user typically has bigger problems already, so, worrying about a few hundred ports being open is generally low on the list.

ktdreyer:
I've rebased this PR onto the tip of master and added the newly-registered IANA-recognized port for ceph-mon (tcp 3300).

ktdreyer:
Thank you!

PR #27, opened 2015-06-24
Fix typos in firewall-cmd helptext

netzvieh:
Fixes two small typos in the firewall-cmd and firewall-offline-cmd help
text.


t-woerner:
Thanks for the patch

PR #39, opened 2015-08-02
add docker registry services

mrjoshuap:
I got tired of manually creating the services for docker and docker registry.

docker 2375/tcp used to remotely control docker servers
docker-registry 5000/tcp used to access docker registry servers

### DIFF
--- a/config/services/docker-registry.xml
+++ b/config/services/docker-registry.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Docker Registry</short>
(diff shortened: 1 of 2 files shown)


rhatdan:
Having docker listen on 2375 right now is opening up your machine to a root attack, without any authentication/authorization protection.

docker run -ti --privileged -v /:/host fedora chroot /root

Connection to the machine gives you full root.  We are working on getting full authentication/authorization into docker but it is not there yet.

We can add this to firewalld but it should be discouraged for now.  The registry one should be fine.

mrjoshuap:
Removed the docker service for 2375, leaving docker-registry service on 5000

PR #40, opened 2015-08-02
add bitcoin related services

mrjoshuap:
The following ports are used by bitcoin to connect to the network and provide full node capabilities.

bitcoin 8333/tcp
bitcoin-rpc 8332/tcp

The following ports are used by bitcoin to connect to the test network and provide full node capabilities.

bitcoin-testnet 18333/tcp
bitcoin-testnet-rpc 18332/tcp


t-woerner:
These port numbers are inofficial and not registered by iana. Do you know why?

PR #67, opened 2016-01-12
config: firewalld.service: Drop Standard{Output,Error}

hwoarang:
Drop the Standard{Output,Error}=null options so we can use the system
logger to find out what firewalld is doing. It also helps when using
--debug in /etc/sysconfig/firewalld as well


hwoarang:
I presume the reason for using Standard{Ouput,Error)=null was that systems running syslog and friends logged firewalld actions twice (journal and /var/log/messages). However, I believe that disabling stdout/stderr all together is probably not optimal since it makes it hard to track what firewalld is doing and renders --debug from /etc/sysconfig/firewalld useless. If double logging is not required, then isn't this something that can be resolved in the syslog config file instead?

t-woerner:
The debug output is logged to /var/log/firewalld, see the firewalld man page. syslog is only used for info,error and warning messages.

PR #68, opened 2016-01-14
smtps.xml: Added SMTPs service

htaira:
I wrote service XML file for SMTPs(SMTP over SSL), TCP port is 465
Currently, firewalld has service XML file for POP3s, IMAPs. But SMTPs is none.
Could you merge my patch?


t-woerner:
Done, thanks

PR #69, opened 2016-01-17
Add ManageSieve service

adrianbroher:
Hello

this pull request adds a service file for the ManageSieve [RFC5804] protocol.  The service file
opens port 4190 for TCP transport as described in the RFC.  It would be great to see this file
added to the project.

[RFC5804]: https://tools.ietf.org/html/rfc5804


t-woerner:
Thanks for the service.

Issue #72, opened 2016-01-25
direct rule gives confusing subnet mask error


hwoarang:
Hi,

I am using 0.3.14.2 and I noticed that the following command leads to a very strange problem

```
~$ firewall-cmd --direct --add-rule ipv4 nat POSTROUTING 1 "-s 192.168.100.0/24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE"
Error: COMMAND_FAILED: '/usr/sbin/iptables -w -t nat -I POSTROUTING_direct 1 -s 192.168.100.0/24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE' failed: iptables v1.4.21: invalid mask `24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE' specified
Try `iptables -h' or 'iptables --help' for more information.
~$
~$ /usr/sbin/iptables -w -t nat -I POSTROUTING_direct 1 -s 192.168.100.0/24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE
~$
```
Running the same commands outside of firewalld does the correct thing though.

I also tried the current master(d7d68f8f79a00f1d9ad66acd0fd1d1c10f130ed7) with the --debug option and the same problem exists

```
2016-01-25 20:26:45 DEBUG1: direct.addRule('ipv4', 'nat', 'POSTROUTING', 1, '-s 192.168.100.0/24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE')
2016-01-25 20:26:45 ERROR: COMMAND_FAILED: '/usr/sbin/iptables -w -t nat -I POSTROUTING_direct 1 -s 192.168.100.0/24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE' failed: iptables v1.4.21: invalid mask `24 -o tun+ -p tcp -m tcp --dport 33:44 -j MASQUERADE' specified
Try `iptables -h' or 'iptables --help' for more information.
```

hwoarang:
It does seem that the problem is with whitespace in args items in execve. https://github.com/t-woerner/firewalld/pull/73 fixes the problem for me

t-woerner:
Arguments containing quotes or double quotes may not be split up into separate words. Please think about comments and log labels which could contain spaces and need to be quoted in this case.

hwoarang:
[...]

PR #73, opened 2016-01-25
firewall: core: prog: Split arguments with spaces into separate words

hwoarang:
Split arguments containing spaces into separate words since execve
normally expects an item in the list of arguments to not contain
any whitespace.

Fixes issue #72: direct rule gives confusing subnet mask error


t-woerner:
Arguments containing quotes or double quotes may not be split up into separate words. Please think about comments and log labels which could contain spaces and need to be quoted in this case.

PR #84, opened 2016-02-12
Add options --save-default-zone, --get-saved-zone, --restore-saved-zone

dougsland:
In order to save the current default zone for later restore this patch
adds the options --save-default-zone, --get-saved-zone, --restore-saved-zone.

An use case as example:
A program ships a zone insinde rpm pkg and before setting it as new default zone
during the instalation (%post) it could get the current default zone and save it.
Later when removing the program (%postun) can restore the previous default zone.

### DIFF
--- a/doc/xml/firewall-cmd.xml
+++ b/doc/xml/firewall-cmd.xml
@@ -218,6 +218,33 @@
 	</varlistentry>
 
 	<varlistentry>
+	  <term><option>--save-default-zone</option></term>
+	  <listitem>
+	    <para>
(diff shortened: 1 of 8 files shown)


dougsland:
The reason I am closing this pull request is because it won't help the use case mentioned above.
Let's say, I have package 'a'.rpm, it saves the current zone during the installation, set the default zone a.xml. However, an external package might be included in my distro and do the same, save current zone and set default zone 'b'.xml. So, there is no guarantee my automate distro creation will ship the zone I expect.

PR #85, opened 2016-02-12
Remove automake generated files

hwoarang:
'install-sh' and 'missing' files depend on the build host so they
shouldn't be part of the git sources. autogen.sh will normally regenerate
these by symlinking them from the system files so they may always appear
as modified when building firewalld from the checked out sources.


t-woerner:
Thanks

PR #87, opened 2016-02-22
src: Makefile.am: Run fix_python_shebang.sh after binaries are in place

hwoarang:
When performing a parallel installation on big systems, there is a race
condition when fix_python_shebang.sh is executed before the binaries are
installed leading to installation failures. Add missing dependencies to
fix this problem.


t-woerner:
Thanks for the patch.

PR #88, opened 2016-02-22
src: core: firewalld_conf: Check if IndividualCalls value is supported

hwoarang:
Checking if the value for the IndividualCalls option is not empty is not
enough. We also need to check if the supplied value is a supported one
otherwise the value will silently be ignored if it is not 'true' or 'yes'
without feedback to the user.
Finally, refactor IndividualCalls code for consistency with the rest of
the options.


t-woerner:
Thanks, for the fix

PR #90, opened 2016-02-26
Only use first part when checking zone name length

AndersBlomdell:
This fixes the error:

INVALID_NAME 'myzone/00_a_very_long_and_descriptive_name' has 42 chars, max is 17

Which is wrong since only the first part of the zone name should be checked (since that is what is used in iptables).

### DIFF
--- a/src/firewall/core/io/zone.py
+++ b/src/firewall/core/io/zone.py
@@ -217,7 +217,7 @@ class Zone(IO_Object):
             raise FirewallError(INVALID_NAME, "'%s' can't end with '/'" % name)
         elif name.count('/') > 1:
             raise FirewallError(INVALID_NAME, "more than one '/' in '%s'" % name)
-        elif len(name) > max_zone_name_len():
+        elif len(name[name.find('/')]) > max_zone_name_len():
             raise FirewallError(INVALID_NAME,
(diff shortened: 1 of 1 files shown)


t-woerner:
The len check is not functional: len(name[name.find('/')]) is always 1

AndersBlomdell:
Right, it should be

  len(name[:name.find('/')])

t-woerner:
If there is no "/" in name, then the size is still not correct as name.find("/") will return -1.

t-woerner:
An extra check `elif "/" in name` should solve this.

AndersBlomdell:
My python begs to differ

>>> name = 'short/with_long_tail'
>>> len(name[:name.find('/')])
5
>>> name = 'long_without_tail'
>>> len(name[:name.find('/')])
16

AndersBlomdell:
Not the clearest code though (since even the author fails to copy his intentions correctly):

t-woerner:
len('long_without_tail') is 17, not 16

You could use this:
elif len(name[:name.find('/')] if "/" in name else name) > max_zone_name_len():

AndersBlomdell:
Whatever is correct, I'm obviously not qualified :-/

PR #91, opened 2016-03-10
Service description

hwoarang:
Hi,

This PR adds --set-{description, short} and --get-{description, short} to manage the short and long descriptions for a service. Furthermore, it also prints such information when requesting all the service info.


t-woerner:
The --list-all* options should not print out the summry and description of a zone, also there is an additional newline before description. The additional lines in the info output are missing in the man pages for the --info-X options.
Please fix this.

hwoarang:
Hi,

Thanks for the review. I pushed the fixes. I move the 'description' field right below the summary one when getting the zone information in order to keep the changes simple. If you think it's better to be at the bottom then I can do that too.

t-woerner:
Thanks for this great improvement.

PR #94, opened 2016-04-12
shell-completion: Add shell completion support for --{get,set}--{desc…

hwoarang:
…ription,short}

Add shell completion for the following options (only available when
--permanent is used): --get-description, --get-short, --set-description,
--set-short


t-woerner:
The --{get,set}-{short,description} options are not only available for zones. Please also add support for ipset, icmptype and service.

hwoarang:
@t-woerner but as far as I can see there is no shell-completion for ipset, icmptype and service

for example ``firewall-cmd --permanent --service=ssh --`` does not propose available completions which is somewhat expected based on the firewall-cmd completion file which seems to only handle  zones and permanent options but not ipset, icmptype and services. That's why I only added support for zone which is inline with the existing code. Adding shell completion for ipset, icmpytype and services needs some infrastructure in place so it probably belongs to a different commit

t-woerner:
Thanks for the patch

PR #96, opened 2016-04-13
Adding IANA-registered sip (SIP) and sips (SIP-TLS) services (RFC3263).

BenjaminLefoul:
This adds two new services:
- SIP (sip) on 5060/tcp and 5060/udp.
- SIP-TLS (sips) on 5061/tcp and 5061/udp.

PS: Obviously, disregard my patch sent directly via the ML earlier this week, but I suppose [this page](http://www.firewalld.org/contribute/) should be clarified.

### DIFF
--- a/config/services/sip.xml
+++ b/config/services/sip.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>SIP</short>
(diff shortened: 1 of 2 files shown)


amessina:
Just asking... Does SIP-TLS need to open 5061/udp?

BenjaminLefoul:
Honestly, I never use SIP-TLS. But 5061/udp is listed here: http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=5061

t-woerner:
According to some documentations 5061/udp seems to be used, therefore it is should be safe to leave it in.

t-woerner:
Thanks for the services

PR #97, opened 2016-04-14
configure.ac: Autodetect the {ip,ip6,eb}tables{,-restore} tools

### LINKED ISSUE #95: ERROR: ipset not usable, disabling ipset usage in firewall.
My system gentoo
4.5.0-gentoo
firewalld - 0.4
systemd

Apr 13 00:24:19 debian /firewalld[3469]: ERROR: ipset not usable, disabling ipset usage in firewall.
Apr 13 00:24:19 debian /firewalld[3469]: ERROR: ebtables not usable, disabling ethernet bridge firewal
Apr 13 00:24:19 debian /firewalld[3469]: FATAL ERROR: No IPv4 and IPv6 firewall.

hwoarang:
Assuming a default path for these tools is error-prone since distributions
install these tools into different places and not all packagers make
use of the --with-* switches to set the correct path. As a result of which,
rework the code to use the standard autoconf macros to detect the path for
these tools during the configure phase. Fixes #95

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,40 +62,43 @@ AC_ARG_WITH([bashcompletiondir],
        [BASHCOMPLETIONDIR=$withval], [BASHCOMPLETIONDIR="${datadir}/bash-completion/completions"])
 AC_SUBST(BASHCOMPLETIONDIR)
 
-AC_ARG_WITH([iptables],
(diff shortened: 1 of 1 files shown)


t-woerner:
Fixed in  da95e48 in combination with AC_ARG_WITH

PR #100, opened 2016-04-22
firewalld: Do not close the Logger file descriptor

hwoarang:
We should not not close the file descriptor used for logging since
that makes firewalld crashing during startup (or later on) with all
sorts of errors.

### DIFF
--- a/src/firewall/core/logger.py
+++ b/src/firewall/core/logger.py
@@ -115,6 +115,10 @@ class FileLog(LogTarget):
         LogTarget.__init__(self)
         self.filename = filename
         self.mode = mode
+        self.fd = None
+
+    def fld(self):
+        return self.fd.fileno()
 
     def open(self):
         if self.fd:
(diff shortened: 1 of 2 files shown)


t-woerner:
Where and how is firewalld crashing during startup because of this?

hwoarang:
Hi,

So I noticed that by executing the following command

```
$ sudo ./src/firewalld --nopid --debug 3
```

Here is the trace (using -m trace)

```
$ sudo python -m trace --trace ./src/firewalld --nopid --debug 3
...
firewalld(98):             if pid > 0:
firewalld(103):             os.chdir("/")
firewalld(104):             os.setsid()
firewalld(105):             os.umask(os.umask(0o077) | 0o022)
firewalld(107):             import resource          # Resource usage information.
trace.py(80):         sys.settrace(None)
firewalld(108):             maxfd = resource.getrlimit(resource.RLIMIT_NOFILE)[1]
firewalld(109):             if (maxfd == resource.RLIM_INFINITY):
firewalld(113):             for fd in range(0, maxfd):
firewalld(114):                 try:
firewalld(115):                     os.close(fd)
firewalld(113):             for fd in range(0, maxfd):
firewalld(114):                 try:
firewalld(115):                     os.close(fd)
$
```

PR #102, opened 2016-04-24
src: firewall-offline-cmd: Use env variable to select a different config file

hwoarang:
Allow users to import their firewall configuration from a different
configuration file using the FW_SYSCONFIG_FILE environmental variable.


t-woerner:
I think a command line option like for example "--system-config-firewall-file="  is better than using an environment variable for this.

hwoarang:
OK no problem I will change it

PR #103, opened 2016-04-25
Fix tests

hwoarang:
This patchset fixes some failures in the python tests.


t-woerner:
I have merged 12cf529, 9ede153 and bda0768.

t-woerner:
I have created other patches for the tests firewalld_rich.py and firewalld_direct.py.

PR #105, opened 2016-04-27
Make sysconfig dir customizable

alucryd:
Some distros don't use /etc/sysconfig but rather /etc/conf.d, this makes it customizable.


t-woerner:
Are there ifcfg files on these distributions in @SYSCONFIGDIR@/network-scripts?

alucryd:
Apologies for the delay. Speaking of Arch Linux, we completely dropped ifconfig since it was deprecated so we don't have any network-scripts directory. Not sure about other distros that might use /etc/conf.d.

erig0:
Closing. This is really old and maybe not relevant anymore. If you disagree please rebase the PR and reopen.

PR #109, opened 2016-05-17
firewall-{,offlind}-cmd: Fix option for setting the destination address

hwoarang:
FirewallD services use --set-destination to set the destination address
for a service instead of --add-destination which is used for icmptypes.
As such, fix the help messages and the related manpages.


t-woerner:
Thanks for this good catch and patch.

PR #116, opened 2016-05-31
src: firewall: core: Drop unneeded python shebangs

hwoarang:
The fw_ifcfg and fw_nm files are not supposed to be executed as
standalone files but rather imported by the main firewalld code so drop
the python shebangs. This also fixes a warning when building firewalld
in openSUSE OBS:

firewalld.noarch: W: non-executable-script
/usr/lib/python2.7/site-packages/firewall/core/fw_nm.py 644 /usr/bin/python
firewalld.noarch: W: non-executable-script
/usr/lib/python2.7/site-packages/firewall/core/fw_ifcfg.py 644 /usr/bin/python

### DIFF
--- a/src/firewall/core/fw_ifcfg.py
+++ b/src/firewall/core/fw_ifcfg.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python
 # -*- coding: utf-8 -*-
 #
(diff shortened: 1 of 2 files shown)


t-woerner:
Thanks, these are left overs from testing

PR #117, opened 2016-05-31
firewall: core: fw_ifcfg: Quickly return if ifcfg directory does not exist

hwoarang:
It's possible for the ifcfg directory to be missing or named differently
so do not try to access it if it does not exist. This avoid warnings
during firewalld start up like the following one:

ERROR: Calling post func <function ifcfg_set_zone_of_interface at
0x7fb2f0d4fc80> (('', 'enp3s0')) failed: [Errno 2] No such file or
directory: '/etc/sysconfig/network-scripts'


t-woerner:
Thanks for discovering this.

PR #120, opened 2016-06-08
Adding nf_conntrack_sip module

BenjaminLefoul:
Similar to what is found in the FTP services.

But I welcome comments on whether or not this is a good idea.


amessina:
I'm not sure adding this to SIP-TLS is necessary.  Is this module inspecting TCP/TLS packets or just UDP packets?

BenjaminLefoul:
You're right. I removed it. The SIP-TLS service itself is unlikely to be used much.

However, can someone with much experience with this module comment on its use for non-TLS?

t-woerner:
As far as I know only helpers that are considered stable make it into the kernel. Therefore I think this could be added.

t-woerner:
Thanks for this.

PR #122, opened 2016-06-10
Added ganglia ports

jcpunk:
This PR adds most of the ganglia services.


t-woerner:
Please do not use daemon names for services. firewalld is using protocol names.
As far as I have seen gmetad is the master node. Therefore I think it should be named ganglia to be compatible to other services. gmond then should be something like ganglia-node (as ganglia-client is not good here).
What do you think?

jcpunk:
Makes sense to me.  I've renamed the items.

t-woerner:
Thanks for the services.

PR #124, opened 2016-06-10
Added RSH service

jcpunk:
My old VXWORKS boxes require RSH access for their boot and setup.  I've added the 'unencrypted' disclaimer from the telnet service to help clarify that this is not an advised service.

### DIFF
--- a/config/services/rsh.xml
+++ b/config/services/rsh.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>rsh</short>
+  <description>Rsh is a protocol for logging into remote machines. It is unencrypted, and provides little security from network snooping attacks. Enabling rsh is not recommended.</description>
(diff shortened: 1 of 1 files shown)


t-woerner:
I am not sure that rsh should be added to the predefined services. The doc string of the service is not used in the gui, therefore the user will not see the warning.

jcpunk:
I am torn on this one, on the one hand rsh really shouldn't be used by anyone.  On the other hand, I've got systems (on an isolated network) that legitimately need the rsh server.

90% of the reason this is in its own PR is so that, if it is rejected there is single place to go for the conversation.

t-woerner:
O.k. as at least you seem to use this still - I am fine with adding. I thought it just might be requested for completeness.
Side note: It would be very good to add the service doc string to firewall-config. But translations are needed then also.

t-woerner:
Thanks for the service.

PR #125, opened 2016-06-17
services ha: Allow corosync-qnetd port

jfriesse:
Corosync is getting new daemon used as quorum network arbiter called
corosync-qnetd. This daemon is running outside of cluster and
comunication is going on tcp port 5403.


t-woerner:
Thanks for the patch.

Issue #75, opened 2016-01-27
RFE: KillMode=mixed


marcosfrm:
firewalld, while does not fork itself on startup (`--nofork`), spawns additional short lived processes (iptables, ebtables, modprobe, ...) during runtime. Default's systemd `KillMode=control-group` (directive from `[Service]` section) sends SIGTERM to all processes present on services' cgroup the moment it is stopped. For daemons that fork things on runtime general recommendation is to use `KillMode=process` (systemd <= 208) or preferably `KillMode=mixed` (systemd >= 209) and let daemon itself deal with its children cleanup.

`process` sends SIGTERM (and SIGKILL after TimeoutStopSec) only to the main daemon process. `mixed` OTOH sends SIGTERM only to the main process, but SIGKILL to all services' cgroup processes after timeout.

I think depend on systemd >= 209 is OK today. Major distros have more recent version: RHEL/CentOS 7.2 (219), Debian 8 (215), SLE/openSUSE 12/13.2 (210), Ubuntu 15.04 (219), Mageia 5 (217). If an older version is used, systemd will emit a warning saying it does not understand `mixed` and will continue with its default value (`control-group`).

http://www.freedesktop.org/software/systemd/man/systemd.kill.html#KillMode=

marcosfrm:
Implemented in v0.4.3.3.

PR #126, opened 2016-06-20
config/firewalld.service.in: use network-pre.target

Siosm:
Firewall services should be ordered before network-pre.target and should pull it, according to : https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

This avoid any specific and fixed 'Before=' dependencies for NetworkManager and libvirtd as those services properly order themselves after network-pre.target or network.target.

### DIFF
--- a/config/firewalld.service.in
+++ b/config/firewalld.service.in
@@ -16,6 +16,7 @@ StandardOutput=null
 StandardError=null
 Type=dbus
 BusName=org.fedoraproject.FirewallD1
+KillMode=mixed
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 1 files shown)


t-woerner:
Is the  network-pre.target available in all systemd versions?

marcosfrm:
Perhaps you can also add `KillMode=mixed` in `[Service]` for #75? `network-pre.target` requires systemd 214 anyway (`mixed` is supported since 209).

Siosm:
Added commit for issue #75.

t-woerner:
Thanks for the patches.

PR #127, opened 2016-06-22
firewallctl manpage

hwoarang:
This pull request add the missing firewallctl.1 manpage to the installed files and also modifies the spec file accordingly.


t-woerner:
Thanks for this

PR #133, opened 2016-06-25
Fix ICMP regression and speedup execution

### LINKED ISSUE #130: ICMP rules not set for some zones
We expect that when a zone (such as `internal`) has interfaces attached to it and firewalld is started, the zone will contain a rule explicitly allowing (or rejecting in case of ICMP block inversion) all ICMP packets.  Such as this:
```
-I IN_external 4 -p ipv6-icmp -j ACCEPT
```

However, In case of non default zones with at least one interface assigned, this does not happen.  This leads to ICMP messages getting rejected in the zones effected by the bug.  See the attached log taken with `firewalld --nofork --debug 10`.  It shows that `external` zone, which is the default, has the rule where as `internal` zone does not.  This happens for both IPv4 and IPv6.
[bug_log.txt](https://github.com/t-woerner/firewalld/files/333024/bug_log.txt)

Adding interfaces later to this zone does not fix the problem. However, in case the zone does not have interfaces assigned during startup and when we add interfaces later, the ICMP rule shows up.

This bug is always reproducible. This behavior is due to the way ICMP block inversion is handled in `core/fw_zone.py:apply_zones()`

### LINKED ISSUE #131: iptables rules are empty after startup
I have noticed on multiple machines that after a successful startup of firewalld, the iptables and ip6tables rules are empty. `iptables -L` and `ip6tables -L` show no rules at all.  However, firewalld log taking using `firewalld --nofork --debug 10` shows that everything went okay.  The iptables-restore command went well.

Taking the input file to this command and manually running `iptables-restore < [tempfile]` creates all the rules.  This hinted to a problem during execution of the iptables-restore command.  Although I didn't see any problem with the exiting mtehod, replacing the `core/prog.py:runProg()` with a method using Python's `subprocess` module fixed the problem.

### LINKED ISSUE #132: Using subprocess module to run programs leads to huge speedup
When `core/prog.py:runProg()` is replaced with a simpler implementation using Python's subprocess module, there is a significant speed up in startup and shutdown time.

Using os.fork() and 1024 os.close() calls:
```
real    0m22.993s
user    0m18.952s
sys     0m1.124s
```
Using subprocess module:
```
real    0m4.323s
user    0m1.000s
sys     0m0.624s
```

SunilMohanAdapa:
### Fix issue with running programs

In some cases, iptables commands are no longer effective.  After
starting firewalld, iptables -L, ip6tables -L are empty.

Using subprocess module to run programs instead of os.fork() fixes this
issue.  (Closes #131)

This also reduced the start and stop time for firewalld on a test
machine from 24 seconds to 4.5 seconds.  (Closes #132)

subprocess.run() is simpler to use but is only available in Python 3.5.

### Fix missing icmp rules for some zones

When a zone has some interfaces and is not the default zone, ICMP
allow/reject rules are not being set on firewalld startup.  This causes
all ICMP packets to be rejected on those zones.  Fix this by making sure
ICMP block inversion rule is always called during startup.
(Closes #130).

### DIFF
--- a/src/firewall/core/fw_zone.py
+++ b/src/firewall/core/fw_zone.py
@@ -171,7 +171,7 @@ class FirewallZone(object):
 
             zone_transaction = transaction.zone_transaction(zone)
 
-            # Do icmp block inversion before setting obj.applied
+            # register icmp block inversion setting but don't apply
             if obj.icmp_block_inversion:
                 self._error2warning(self.add_icmp_block_inversion, obj.name,
                                     use_zone_transaction=zone_transaction)
@@ -211,6 +211,11 @@ class FirewallZone(object):
                                     use_zone_transaction=zone_transaction)
             for args in obj.sources:
                 self._error2warning(self.add_source, obj.name, args,
+                                    use_zone_transaction=zone_transaction)
+            # apply icmp accept/reject rule always
+            if obj.applied:
+                self._error2warning(self.__icmp_block_inversion, True,
+                                    obj.name,
                                     use_zone_transaction=zone_transaction)
 
         if use_transaction is None:
--- a/src/firewall/core/prog.py
+++ b/src/firewall/core/prog.py
@@ -19,61 +19,29 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-__all__ = [ "runProg" ]
+import subprocess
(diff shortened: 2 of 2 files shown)


t-woerner:
Thanks for these fixes.

PR #141, opened 2016-07-29
Add UDP to transmission-client.xml service

Aeyoun:
Added UDP protocol as both DHT and µTP are enabled in Transmission (and essential to modern-day BitTorrent) by default and they depend on UDP. 

Transmission help file says to enable both tcp and udp:
https://www.transmissionbt.com/help/gtk/2.9x/html/pfrouter.html

Removed reference to "GTK+" as there are termainl, Qt, web, and daemon variants too.

### DIFF
--- a/config/services/transmission-client.xml
+++ b/config/services/transmission-client.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <service>
   <short>Transmission</short>
-  <description>Transmission is a lightweight GTK+ BitTorrent client.</description>
+  <description>Transmission is a lightweight BitTorrent client.</description>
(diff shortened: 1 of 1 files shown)


t-woerner:
The UDP port is not listed in https://trac.transmissionbt.com/wiki/PortForwardingGuide though. The documentation does not seem to be consistent.

Aeyoun:
It is nonetheless required for DHT and µTP to work.

Edit: It does actually say on that page to [open “both”](https://trac.transmissionbt.com/wiki/PortForwardingGuide#ForwardmanuallythroughaRouter) (referring to transmission protocols).

t-woerner:
Thanks for the patch.

PR #143, opened 2016-08-10
firewall: core: fw_nm: Instantiate the NM client only once

hwoarang:
If the NM module is available, instantiate the NM client when the module
is imported instead of re-instantiating in every method.
This also fixes a side-effect in firewall-applet and firewall-config
when the [connectivity] section is used in NetworkManager.conf and the
remote end pointed by the 'uri' option is unavailable. In that case,
when a NM client is started, the NetworkManager will send a
PropertiesChanged signal to notify interested subscribers that the
remote end can't be reached. However, such signals are being handled by
the nm_signal_receiver method which instantiates a new NM client because
of the nm_get_connections call. This results in another PropertiesChanged
signal to be sent again so firewall-applet and firewall-config are stuck
in an infinite loop handling the said signal.

Link: https://bugzilla.opensuse.org/show_bug.cgi?id=992082
Signed-off-by: Markos Chandras <mchandras@suse.de>

### DIFF
--- a/src/firewall/core/fw_nm.py
+++ b/src/firewall/core/fw_nm.py
@@ -31,13 +31,16 @@ try:
     gi.require_version('NM', '1.0')
 except ValueError:
     _nm_imported = False
+    _nm_client = None
 else:
     try:
         from gi.repository import NM
         _nm_imported = True
+        _nm_client = NM.Client.new(None)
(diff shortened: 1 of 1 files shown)


t-woerner:
Thanks for the patch

PR #144, opened 2016-08-11
Update AppData configuration file.

jpopelka:
I've been asked by Richard Hughes to update the AppData config file so I ran
`appstream-util upgrade firewall-config.appdata.xml.in`
filled the new metadata fields and here's the result.

### DIFF
--- a/config/firewall-config.appdata.xml.in
+++ b/config/firewall-config.appdata.xml.in
@@ -1,26 +1,30 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright 2014 Jiri Popelka <jpopelka@redhat.com> -->
-<application>
- <id type="desktop">firewall-config.desktop</id>
- <metadata_license>CC0-1.0</metadata_license>
- <project_license>GPL-2.0+</project_license>
- <description>
-  <p>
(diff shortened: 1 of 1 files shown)


t-woerner:
Is this also working with older appstream versions?

jpopelka:
Not sure, but I'd expect that the unknown metadata fields are ignored.

t-woerner:
Please use https://raw.githubusercontent.com/t-woerner/firewalld/master/doc/firewall-config.png for the firewall-config screenshot.

t-woerner:
According to https://people.freedesktop.org/~hughsient/appdata/ it should be working also with the old version.

jpopelka:
Rebased with the new screenshot link.

t-woerner:
The image is new with different size: 870 x 600
Is the image size needed there at all? If the image gets replaces sometime in the future for example the size might not match again.

jpopelka:
No, I don't think it's needed. Rebased.

t-woerner:
Thanks for the patch

hughsie:
FWIW, the "new" metadata format is supported even in Fedora 21, so updating now is completely safe.

PR #145, opened 2016-08-22
Add condor collector service

jcpunk:
Added service for the central condor master.


t-woerner:
Thanks for the new service.

Issue #142, opened 2016-08-07
Please add submission (SMTP over port 587)


SpareSimian:
Essentially a duplicate of smtp.xml, but with port changed to 587. Used for authenticated submission of email to the SMTP server.

hspaans:
I prepared a service file for SMTP-Submission, but I wait for #150 to be correctly processed as some checks fail currently.

[https://github.com/hspaans/firewalld/commit/069c5717ee6590d3396e12140ae24d817ab9751d](url)

t-woerner:
Fixed with commit 958d9c7

t-woerner:
And 069c571

da2x:
Why was this called `smtp-submission`? The standard service name this is referred to elsewhere is always just `submission`? It caused a discoverability and confusion-issue over on [Ask Fedora](https://ask.fedoraproject.org/en/question/107382/).

Can it be renamed with the common service name?

viniciusferrao:
It makes no sense to be called `smtp-submission`.

```
[root@mx ~]# firewall-cmd --add-service=submission --permanent
Error: INVALID_SERVICE: Zone 'public': 'submission' not among existing services
[root@mx ~]# firewall-cmd --add-service=smtp-submission --permanent
success
```

SpareSimian:
Would a symlink work?

viniciusferrao:
I took the liberty of writing a pull request (PR) to better describe the SMTP submission service, that can be found here: https://github.com/firewalld/firewalld/pull/1208

viniciusferrao:
> Why was this called `smtp-submission`? The standard service name this is referred to elsewhere is always just `submission`? It caused a discoverability and confusion-issue over on [Ask Fedora](https://ask.fedoraproject.org/en/question/107382/).
> 
> Can it be renamed with the common service name?

The fix has been merged. I hope we get it on the next releases.

PR #150, opened 2016-08-28
Add CFEngine service.

hspaans:
used for the CFEngine...

add 5308/tcp for cfengine service


hspaans:
Accidentally also pushed an update for #142 to my repo.

t-woerner:
Thanks for the two services.

Issue #147, opened 2016-08-27
Please call syslog.openlog() to work-around Python issue 27875


canvon:
When using syslog.syslog() without calling openlog() before, it gets called automatically with no arguments, which then should set the logging identifier "to sys.argv[0] with leading path components stripped". It does not strip the final slash, though. So firewalld ends up logging, e.g.:

```
# firewall-cmd --permanent --delete-zone=test
Error: INVALID_ZONE: test

Aug 27 18:24:47 blackbox /firewalld: ERROR: INVALID_ZONE: test
```

The additional slash appears as most unusual, and may lead to the impression that there would be a bug in firewalld which causes this (which there isn't), or a chroot would be involved (which appears is not the case). So I consider the additional slash a bug and filed https://bugs.python.org/issue27875

Until that gets fixed (if at all), a work-around by calling syslog.openlog() with an explicit identifier would be nice. (While at it, that could include the PID as well (via logoption=syslog.LOG_PID?) -- when viewing the systemd journal, the PID seems to get inserted automatically anyhow, but in the plain rsyslogd log files, it is missing for firewalld.) The documentation is at https://docs.python.org/3/library/syslog.html

t-woerner:
Fixed in https://github.com/t-woerner/firewalld/commit/adb927065989a305d1a6a412c73218c67f86cf11

t-woerner:
Thanks for the patch, I used it as a base with reference to your patch.

canvon:
Thank you. :)

That renders PR #153 unnecessary.

PR #153, opened 2016-09-04
firewall.core.logger: Use syslog.openlog()

canvon:
Fixes: #147

* Work around Python issue 27875, "Syslogs /usr/sbin/foo as /foo
  instead of as foo".  https://bugs.python.org/issue27875

* Set syslog logging ident (to basename), include daemon PID
  and log to facility "daemon" (instead of the default, "user").

---

Before, firewalld logs like this:

```
# firewall-cmd --permanent --delete-zone=test
Error: INVALID_ZONE: test

Aug 27 18:24:47 blackbox /firewalld: ERROR: INVALID_ZONE: test
```

After, it logs like this:

```
Sep  3 18:17:52 blackbox firewalld[1353]: ERROR: INVALID_ZONE: test
```

The code is supposed to be called at module loading time / once.

### DIFF
--- a/src/firewall/core/logger.py
+++ b/src/firewall/core/logger.py
@@ -83,6 +83,12 @@ class _StderrLog(_StdoutLog):
 class _SyslogLog(_StdoutLog):
     fd = None
 
+    # Work around Python issue 27875, "Syslogs /usr/sbin/foo as /foo instead of as foo"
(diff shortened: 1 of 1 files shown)


canvon:
Fixed differently in adb9270 in #147 .

PR #164, opened 2016-11-06
Include reminder that zones must be configured

da2x:
It kind of sounds like the zones are auto-configuring. Include reminder that that isn't the case.

I stumbled across a few tutorials for firewalld that all explain that zones are great, but all either glance over them or make it sound like they're automatically configured. Kind of a dangerous misunderstanding, so I thought it was a good idea to clarify.


erig0:
Closing old PRs.

PR #171, opened 2016-11-16
firewall.core.fw_nm: create NMClient lazily

lkundrak:
This actually fixes a couple of issues:

NetworkManager.service depends on firewalld.service, so that it doesn't
proceed activating services before it can set the firewall zones. That
means that fw_nm's attempt to create a NMClient always fails.

Secondly, the failure to construct a NMClient is silently discarded in
the except branch. Not cool.

Lastly, asynchronous NMClient creation could be rather expensive if
there's a lot of devices or connections. (Thankfully, it doesn't
succeed, right?). Let's not delay the service startup with it.

### DIFF
--- a/src/firewall/core/fw_nm.py
+++ b/src/firewall/core/fw_nm.py
@@ -32,15 +32,13 @@ try:
     gi.require_version('NM', '1.0')
 except ValueError:
     _nm_imported = False
-    _nm_client = None
 else:
     try:
         from gi.repository import NM
         _nm_imported = True
-        _nm_client = NM.Client.new(None)
     except (ImportError, ValueError, GLib.Error):
         _nm_imported = False
-        _nm_client = None
+_nm_client = None
 
 from firewall import errors
 from firewall.errors import FirewallError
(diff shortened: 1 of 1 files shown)


t-woerner:
You are right, the NMClient creation right at import time is bad.

But with this change it is needed to add `global _nm_imported` and `global _nm_client` to the functions using `_nm_imported` and `_nm_client`. Somehow this is not needed with the original code.

src/firewall/core/fw_nm.py
```diff
@@ -52,6 +50,7 @@ def check_nm_imported():
     """
     if not _nm_imported:
         raise FirewallError(errors.MISSING_IMPORT, "gi.repository.NM = 1.0")
+        _nm_client = None
```
lkundrak: Yeah, dropped.

lkundrak:
> But with this change it is needed to add global _nm_imported and global _nm_client to the functions using _nm_imported and _nm_client. Somehow this is not needed with the original code.

I think it's only the places, that are *assigning* the names, not merely using them? If I got it correctly how does "global" keyword work; I am not too familiar with Python and had to look into the manual.

(updated the branch)

t-woerner:
Yes, `global _nm_client` is only needed in `nm_get_client` where `_nm_client` is assigned.

t-woerner:
Thanks for the patch

PR #180, opened 2016-12-08
Run interface rules even when NetworkManager is active

AndersBlomdell:
Later versions of firewalld does not run all needed commands when NetworkManager is active


erig0:
Perhaps this was fixed by, 99e1c79249445 ?

erig0:
No feedback, so I will assume this was fixed by 99e1c79. If that's wrong then please reopen.

Issue #89, opened 2016-02-26
INVALID_NAME 'myzone/00_a_very_long_and_descriptive_name' has 42 chars, max is 17


AndersBlomdell:
Which is wrong since only the first part of the zone name should be checked (since that is what is used in iptables).

From dc34f7b8912536b3fc9aa6aa6422089ccd63b618 Mon Sep 17 00:00:00 2001
From: Anders Blomdell anders.blomdell@control.lth.se
Date: Fri, 26 Feb 2016 13:19:22 +0100
Subject: [PATCH] Only use first part when checking zone name length

---

 src/firewall/core/io/zone.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/firewall/core/io/zone.py b/src/firewall/core/io/zone.py
index 3b72a2d..5dcf042 100644
--- a/src/firewall/core/io/zone.py
+++ b/src/firewall/core/io/zone.py
@@ -217,7 +217,7 @@ class Zone(IO_Object):
             raise FirewallError(INVALID_NAME, "'%s' can't end with '/'" % name)
         elif name.count('/') > 1:
             raise FirewallError(INVALID_NAME, "more than one '/' in '%s'" % name)
-        elif len(name) > max_zone_name_len():
-        elif len(name[name.find('/')]) > max_zone_name_len():
           raise FirewallError(INVALID_NAME,
                               "'%s' has %d chars, max is %d" % (name, len(name), max_zone_name_len()))
## 

2.5.0

/

AndersBlomdell:
Changed to a merge request

AndersBlomdell:
Still an issue :-(

AndersBlomdell:
Rebased version:

https://github.com/t-woerner/firewalld/pull/181

PR #181, opened 2016-12-08
check_name fix for combined zones

AndersBlomdell:
Rebase of https://github.com/t-woerner/firewalld/pull/154


AndersBlomdell:
Any reason for not merging?

src/firewall/core/io/zone.py
```diff
@@ -245,11 +245,19 @@ def check_name(self, name):
         elif name.count('/') > 1:
             raise FirewallError(errors.INVALID_NAME,
                                 "more than one '/' in '%s'" % name)
-        elif len(name) > max_zone_name_len():
-            raise FirewallError(
-                errors.INVALID_NAME,
-                "'%s' has %d chars, max is %d" % (name, len(name),
-                                                  max_zone_name_len()))
+        else:
+            if "/" in name:
```
erig0: AFAICS, this change is unnecessary since you added no_check_name  to zone_reader() for the combined case. We do check_name() for the combined_zone with the modified name.

erig0 (changes requested):

AndersBlomdell:
If memory serves me right, that part i needed for check_name(...) calls from firewalld-cmd, but some time has passed since the merge request...

src/firewall/core/io/zone.py
AndersBlomdell: If memory serves me right, that part i needed for check_name(...) calls from firewalld-cmd, but some time has passed since the merge request...

erig0: Alright. I'll try to create a test case for this #89 to verify. I hope to get to it soon, but I have some other things on my plate first.

erig0 (commented):
> Higly disturbing that code is duplicated in fw.py and fw_test.py.

100% agree. It's on my TODO list to remove it. Right now the only thing using it is firewall-offline-cmd. I've since redone the test framework so I think fw_test.py is not needed anymore.

PR #196, opened 2017-01-20
Add InfluxDB service

QuentinMoss:
Add InfluxDB to provided services.

https://docs.influxdata.com/influxdb/v1.1/administration/ports/


t-woerner:
The ports seem to have partly changed with version 1.2. 8083 seemed to be the admin port, that might not need to be opened for normal client operation. New is 8088 for backup and restore.

I my opinion it might be good to ship the firewalld service file with the InfluxDB package as the ports are depending on the version of InfluxDB till they are stable again.

QuentinMoss:
👍

PR #205, opened 2017-02-11
add bgp service to predefined services

pronix:
using name simplified automation and integration with puppet modules.


pronix:
@erig0 thanks for comment. config/Makefile.am updated.

PR #209, opened 2017-03-02
Add mssql services

htaira:
I added mssql services for Microsoft SQL Server on Linux. It use 1433/tcp.


erig0:
mssql is already an included service.

PR #213, opened 2017-03-27
firewalld.spec: Add missing autotools dependencies

hwoarang:
Commit 0ab7673fd666 ("Fix build from spec without fedorahosted.org
archives") added 'autogen.sh' to the %prep phase but this script calls
'autoreconf' so we need to add the necessary autotools dependencies to
the spec file as well.

Fixes: 0ab7673fd666 ("Fix build from spec without fedorahosted.org archives")
Signed-off-by: Markos Chandras <mchandras@suse.de>


t-woerner:
Thanks for the fix

PR #215, opened 2017-04-10
add new service IRCd

sassmann:
Common ports for IRC daemons are TCP 6665-6669.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -153,6 +153,8 @@ CONFIG_FILES = \
 	services/ipp-client.xml \
 	services/ipp.xml \
 	services/ipsec.xml \
+	services/ircs.xml \
+	services/irc.xml \
 	services/iscsi-target.xml \
 	services/kadmin.xml \
 	services/kerberos.xml \
--- a/config/services/irc.xml
+++ b/config/services/irc.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>IRC</short>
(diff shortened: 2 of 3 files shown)


teadur:
6697 is common ssl port for ircd
Aswell 7000 is used on some networks

sassmann:
right, should the ssl port be a separate feature or put everything into one?

sassmann:
updated with all ports 6665-6669, 6697,7000.

t-woerner:
I would like to keep the port range that will be opened as small as possible. As far as I know 6667 is the default port that is used for most irc servers. Wouldn't it be sufficient to open this port only and if the admin wants to use another port either the firewalld service could be adapted to the needs or the used port could be opened only?
If you open all these ports with the service, then any user service that uses one of these unprivileged ports to be accessible, which would be unexpected.

teadur:
I would follow standards as much as possible, users can run arbitary services on different ports.
6665-6669 is defined by https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=6665

tls port is defined by https://tools.ietf.org/html/rfc7194#section-4

sassmann:
I can understand both sides. We don't want to punch too many holes into the firewall, OTOH the ports are standardized. Another idea would be to add 2 service files. The first being a single port 6667, which is probably enough for most people. The second including all the standardized ports.
Thoughts?

teadur:
I would include 6697 (most modern networks offer tls to clients) with 6667 aswell.
But from standards perspective it would make more sense to have irc 6665-6669 and ircs 6697

PR #216, opened 2017-04-28
Fix misspelled word in documentation

tobiasvl:
Change "protcol" to "protocol" in documentation


t-woerner:
Thanks for the patch

PR #217, opened 2017-05-02
Adding ovirt-vmconsole service file

leongold:


t-woerner:
This is the same port as in pull request #212 for ovirt-imageio - is this intentional?

leongold:
No it isn't, thanks for noticing!

PR #230, opened 2017-06-12
Add docker swarm service

tdobrovolny:
Docker swarm service allow communication between docker engines in
swarm mode.

### DIFF
--- a/config/services/docker-swarm.xml
+++ b/config/services/docker-swarm.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <service>
-  <short>Docker swarm</short>
-  <description>Ports and protocol required by docker in swarm mode</description>
+  <short>Docker integrated swarm mode</short>
+  <description>Natively managed cluster of Docker Engines (>=1.12.0), where you deploy services.</description>
   <port port="2377" protocol="tcp"/>
   <port port="7946" protocol="tcp"/>
   <port port="7946" protocol="udp"/>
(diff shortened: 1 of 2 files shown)


erig0:
You also need to update config/Makefile.am

config/services/docker-swarm.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Docker swarm</short>
+  <description>Ports and protocol required by docker in swarm mode</description>
```
erig0: This should be a one line description of Docker Swarm not a description of the service file.

tdobrovolny: TCP port 2376 is not required by docker engines running in swarm mode. See https://docs.docker.com/engine/swarm/swarm-tutorial/#open-protocols-and-ports-between-the-hosts

Your link describes legacy standalone swarm used on docker prior the version 1.12. See https://docs.docker.com/swarm/provision-with-machine/

So if you want also open port 2376 needed for docker-machine and legacy swarm, I recommend have two firewalld services. One for "Docker legacy Swarm" also with port 2376 and one for "Docker integrated Swarm" without 2376 port.

tdobrovolny:
TCP port 2376 is not required by docker engines running in swarm mode. See https://docs.docker.com/engine/swarm/swarm-tutorial/#open-protocols-and-ports-between-the-hosts

Your link describes legacy standalone swarm used on docker prior the version 1.12. See https://docs.docker.com/swarm/provision-with-machine/

So if you want also open port 2376 needed for docker-machine and legacy swarm, I recommend have two firewalld services. One for "Docker legacy Swarm" also with port 2376 and one for "Docker integrated Swarm" without 2376 port.

FruityWelsh:
This TCP port is also used by Rancher nodes as per the [Rancher req docs](https://rancher.com/docs/rancher/v2.x/en/installation/requirements/). For "Docker daemon TLS port used by Docker Machine (only needed when using Node Driver/Templates)".

PR #231, opened 2017-06-14
Add git service

jcpunk:
It may be helpful to have the git-daemon pre-defined as a service.  I know I'd use it.


erig0:
Hi jcpunk,
This also needs to update config/Makefile.am.

jcpunk:
Fixed

PR #233, opened 2017-06-26
Add kprop service

jcpunk:
This service is used by MIT Kerberos for KDC slave sync.


erig0:
Please also update config/Makefile.am

jcpunk:
fixed

PR #236, opened 2017-07-09
minidlna definitions

sghosh151:
Added Service definition for MiniDLNA


erig0:
Hi sghosh151,
I think you also need to add this new file to CONFIG_FILES in config/Makefile.am.

sghosh151:
@erig0 - thank for the hint
added

Issue #185, opened 2016-12-11
RFE: Support client side conntrack helpers


aalba6675:
Issue: --add-service tftp-client doesn't seem to help TFTP connections **from** this device as the rule goes into PREROUTING instead of OUTPUT.

Question: Is this the intended behaviour? Perhaps the documentation could be clearer that this is meant for the device acting as a NAT+router not TFTP client? Then won't we need nf_nat_tftp as well?

For "--add-service tftp-client" the rule

-j CT -p udp -m udp --dport 69 --helper tftp 

goes into PRE_$INZONE_allow. This helps TFTP connections **through** this device but not **from** this device, right? This still won't work without enabling nf_nat_tftp (if device is doing NAT as well as routing).

If I wanted to enable tftp-client **from** this device, I would need

firewall-cmd --direct --add-rule -j CT ipv4 raw OUTPUT 0 -o $OFACE -p udp -m udp --dport 69 --helper tftp 

which is not obvious.

aalba6675:
Clean Fedora 25 install with firewalld-0.4.4.2. Sure had a hard time getting the TFTP client to work because I didn't  realise that "--add-service tftp-client" was inserting a rule in PREROUTING not OUTPUT (as per the documentation). The return TFTP connection was still being blocked.

Mismatch in expectation because "--add-service tftp-client" sounds like "enables TFTP client from this machine through this zone".

erig0:
In order to support this, firewalld first needs real OUTPUT support. Once that is done we can attach the helpers to the OUTPUT chains.

Issue #25, opened 2015-06-07
Add SSDP/UPnP service?


bigon:
Hello,

Wouldn't be nice to add a "SSDP" or "UPnP" service?

The "announcements" messages are sent over udp using port 1900 with destination address 239.255.255.250 (multicast), that would be a good start already I guess?

There is a [Fedora bug](https://bugzilla.redhat.com/show_bug.cgi?id=892801) already opened about this.

aboe76:
something like:
ssdp.xml:
```
<?xml version="1.0" encoding="utf-8"?>
<service version="1">
  <short>Multicast ssdp (upnp)</short>
  <description>Simple Service Discovery Protocol  for advertisement and discovery of network services and presence information.</description>
  <port port="1900" protocol="udp" />
  <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
</service>
```

Issue #32, opened 2015-07-25
RFE: Add a way to filter outgoing traffic


genodeftest:
Currently there is no easy way to filter outgoing traffic. It would be nice to add this functionality. I would like to block outgoing traffic by default except for some critical services like dns, web browser, email client, package manager.

edgimar:
I also need a way to change OUTPUT rules for zones.  In particular, I have some wifi connections which are tethered / metered data connections, and for these, I would prefer to add some output rules that block all traffic except traffic from a special "internet" gid (using the iptables 'owner' module).  Furthermore, I want the rules only to be active when I am using a connection which is designated as being in the 'metered' zone.

By the way, I notice that this issue is named [GUI]... -- does this mean that what I describe here is already possible using firewall-cmd?  Direct rules seem not to be the right thing, since I want the rules to only be active when connected to certain connections.

genodeftest:
I didn't test it on firewall-cmd.

t-woerner:
Output filtering is planned for one of the next releases. There is already support for output filtering with direct rules, though. Since 0.4.1.2 direct rules can be reliably added to zones.

Magissia:
Does this milestone tag means we will get outgoing rules support for version 0.5 ?
Firewalld is an interesting project to initiate people to firewall, but without proper outgoing support it doesn't work for that purpose as direct rules would be too hardcore.

Thanks

t-woerner:
Yes, that is the plan.

t-woerner:
It might be good to get information about the needs of users, administrators and projects for outgoing filtering.

BenjaminLefoul:
I think most users would like "firewall-cmd --add-interface=lo" or "echo 'ZONE=public' >> ifcfg-lo" to JUST-WORK™ :).

t-woerner:
@BenjaminLefoul Please explain what you want to achieve with that.

BenjaminLefoul:
Most sysadmins I work with typically have a web service of sorts running on a non-standard port and do a port forward with firewalld. By adding "lo" to the zone, they expect a "curl http://localhost/somewebapp" to work (without specifying the port). If that is impossible, I would be glad to refer them to your explanation.

t-woerner:
O.k. you want to have support for transparent proxies. The current implementation of port forwarding is not allowing this. There is an other request to support the REDIRECT target that will enable this. This is #78

teadur:
Outgoing filtering is definetly must have for servers.

[...]

PR #237, opened 2017-07-14
Output filtering

ashh87:
This set of patches extends support for helpers by adding an option to specify whether they are added to the raw OUTPUT chain or the raw PREROUTING chain. I'm not sure if there is a situation where both might be required for a helper, but I can update if so. If unspecified, the default is PREROUTING, so existing helper xml files are currently unchanged, and custom helpers outside of this repo won't break.

This might help with resolving #185 .The SSDP conntrack helper needs to go on raw OUTPUT, so this is needed for #25.


t-woerner:
Please append new fields always to the end of a D-Bus signature. Inserting them in the middle is not recommended as it will break existing code.
I think that a field as `direction` (`input`|`output`) might be better as naming the chain.

src/firewall/core/io/helper.py
```diff
@@ -42,17 +42,18 @@ class Helper(IO_Object):
         ( "description", "" ),                # s
         ( "family", "", ),                    # s
         ( "module", "", ),                    # s
+        ( "chain", "", ),                     # s
         ( "ports", [ ( "", "" ), ], ),        # a(ss)
         )
-    DBUS_SIGNATURE = '(sssssa(ss))'
+    DBUS_SIGNATURE = '(ssssssa(ss))'
```
t-woerner: Please add the new element to the end of the signature.

ashh87:
I've updated so that UI is setting the 'direction' property to either 'input' or 'output'.

ashh87:
Would this be more useful if I merged it with master?

erig0:
It would be helpful if you could squash some commits together. There are quite a few instances where something changes, then changes again in later patches. This makes it hard to review. For example, the dbus signature changes 3 times in separate commits.

ashh87:
No problem, I'll get that done soon.

ashh87:
Hopefully that should be a bit simpler, although github pull request shows the commits in the wrong order. I've not squashed the renaming of 'chain' to 'direction' as suggested, because it seems like a separate issue to me, and because it required a new dictionary to be added to convert one string to another. Please let me know if it needs further improvements!

erig0:
@ashh87, sorry for the long delay. Thanks again for your effort on these patches.

I have been giving this quite a bit of thought. There are also many requests to support OUTPUT filtering in other areas of firewalld. As such, I think this could be better implemented by adding a generic "--direction={input, output}" flag. This could support all aspects of a service; rules, helper, etc. It also be used for --add-protocol, --add-port, etc. The direction flag could also be extended to support "forward" in the future.

I've begun working on this, but it is not a small task and likely means introducing new DBUS methods. I will update this issue when I have something more to report.

erig0:
Closing this PR. We need to implement general OUTPUT/FORWARD filtering support not helper specific extensions.

henryptung:
@erig0 Is there a general issue tracking this?

erig0:
> Is there a general issue tracking this? 

Yes, #32.

> FORWARD filtering is its own concept, which involves a lot of new design

Indeed. Awhile ago I put together proposed changes that cover all the points you mention. I'll have to clean that up and push it to the git repo.

> OUTPUT filtering could be useful, but I think one of its features will be exactly the "helper-specific extension" you mention above.

There are other use cases outside of helpers; e.g. notify of new traffic from applications.

mwilck:
I just became aware of the issue here. Let me summarize shortly, correct me if I'm wrong.

The `samba-client` service of firewalld is non-functional. It opens port 137/udp and 138/udp, which is not necessary for SMB client functionality, but blocks responses from UDP port 137 (see #266, #267). There are two possible workarounds:

 1. open source port 137
 2. enable "automatic helpers" with `samba-client`, or enable the `net.netfilter.nf_conntrack_helper` sysctl system-wide.

*Both workarounds are discouraged for security reasons*. (2) is probably less critical, so it would be recommended for end users who need SMB client functionality (?)

The reason these workarounds are necessary is that firewalld implements the conntrack filtering logic wrongly, at least for `netbios-ns`. It inserts a `-j CT --helper netbios-ns` rule in the `PREROUTING` chain, but that's wrong - the rule belongs in the `OUTPUT` chain. The rule in `PREROUTING` is *not* necessary to make NetBIOS name lookups work.

This problem has been known for at least 3 years, and patches have been proposed e.g. in this PR, but have been rejected in favor of a more generalized approach which is tracked in #32. The latter issue is tracked as an RFE, as it's mainly seen as an additional feature. Very obviously it has been tracked with priority, which is appropriate for an RFE, but not so much for a bug. 

Unless I'm missing something essential, the `samba-client` service in firewalld is broken, and I would call that a bug. I believe that some short-term solution should be targeted for this problem, as the generic "output filtering" functionatlity is apparently still quite far away. There are likely other conntrack helpers that are affected by this the same way as `netbios-ns`.

PR #247, opened 2017-09-23
Add llmnr service

rcorrear:
https://en.wikipedia.org/wiki/Link-Local_Multicast_Name_Resolution. Useful for Windows interop.

### DIFF
--- a/config/services/llmnr.xml
+++ b/config/services/llmnr.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Link-Local Multicast Name Resolution (LLMNR)</short>
+  <description>LLMNR allows both IPv4 and IPv6 hosts to perform name resolution for hosts on the same local link.</description>
+  <port protocol="tcp" port="5355"/>
+  <port protocol="udp" port="5355"/>
+</service>
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -171,6 +171,7 @@ CONFIG_FILES = \
 	services/ldap.xml \
 	services/libvirt-tls.xml \
 	services/libvirt.xml \
+	services/llmnr.xml \
 	services/managesieve.xml \
 	services/mdns.xml \
 	services/minidlna.xml \


rcorrear:
@erig0 sorry this update took so long. You're correct, systemd-resolved also listens on TCP. Fixed the Makefile as well, thanks for the review.

config/services/llmnr.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Link-Local Multicast Name Resolution (LLMNR)</short>
+  <description>LLMNR allows both IPv4 and IPv6 hosts to perform name resolution for hosts on the same local link.</description>
+  <port protocol="tcp" port="5355"/>
+  <port protocol="udp" port="5355"/>
+  <destination ipv4="224.0.0.252" ipv6="ff02::1:3"/>
```
erig0: Using destination means it's part of the match criteria for the TCP port as well (see below). So I think you have to omit the destination in the service definition to allow the unicast queries.

e.g.
```
Chain IN_public_allow (1 references)
target     prot opt source               destination         
...
ACCEPT     tcp  --  0.0.0.0/0            224.0.0.252          tcp dpt:5355 ctstate NEW
ACCEPT     udp  --  0.0.0.0/0            224.0.0.252          udp dpt:5355 ctstate NEW
```

erig0: @rcorrear, Any comment?

rcorrear: @erig0, in this case, the only possibility I see is to have two different service files, one for the UDP portion with the destination tag, and another with just the TCP port (which seems a little overkill) or having just this file without the destination part. Which one is preferable to you?

erig0: @rcorrear, I prefer the later. Only use the <port> lines.

rcorrear: @erig0 please have another look.

rcorrear:
@erig0 please have another look

PR #253, opened 2017-10-09
Adding FirewallD mongodb service

hos7ein:
opening tcp port 27017 for mongodb


config/services/mongodb.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>mongodb</short>
+  <description>MongoDB is a free and open-source cross-platform document-oriented database program.</description>
+  <port protocol="tcp" port="27017"/>
```
erig0: mongo also uses other ports: https://docs.mongodb.com/manual/reference/default-mongodb-port/
Should these also be included? Are they commonly used?

hos7ein: the other port use when "sharding" or "web status page" is enable in mongodb config file.
by default  "sharding" or "web status page" is disable in mongodb and user must be configure and enable manual.

PR #260, opened 2017-11-03
Add new UPnP client service

da2x:
Important for many desktop programs, especially for home users.

This really really shouldn’t be enabled on public zones, so included
suggestion to only use it in trusted zones in the service description.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -230,6 +230,7 @@ CONFIG_FILES = \
 	services/tinc.xml \
 	services/tor-socks.xml \
 	services/transmission-client.xml \
+	services/upnp-client.xml \
 	services/vdsm.xml \
 	services/vnc-server.xml \
 	services/wbem-https.xml \
--- a/config/services/upnp-client.xml
+++ b/config/services/upnp-client.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>UPnP Client</short>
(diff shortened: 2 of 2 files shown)


erig0:
Please update config/Makefile.am

ashh87:
I think this will need something like #237 to work properly (UPnP is why I did that originally). Also #25 is an issue covering the same thing.

erig0:
I believe this would be sufficient to let the replies pass the INPUT filter. It's filtering based on source-port which should be 1900. But I agree it would be better if conntrack could handle this, otherwise it will let unsolicited packets through.

I'm be willing to take this under the assumption that we'll fix up the *-client services when OUTPUT filtering and helpers are supported. Will wait to hear @ashh87's feedback.

da2x:
> Please update config/Makefile.am

I’d done that locally but forgot to push. Sorry, I’ll fix that right away!

UPnP requests are broadcast from an arbitrary local port, and receive unicast responses from port 1900 from the nearest router. I’m not sure how else  to allow UPnP than to allow requests from source port 1900. Issue #25 covers a UPnP server, whereas this adds a UPnP client.

I included a note in the service description about only enabling this on trusted zones. It’s not a technical mitigation against unsolicited packets but that isn’t really an option with UPnP anyway. If I remember correctly, an UPnP server can send the client an unsolicited packet at a later time informing it that it has closed the port-mapping so the client will need to take action to re-open it.

ashh87:
Ah sorry, I was indeed thinking about sending from 1900 and receiving on another port, rather than opening 1900 so that M-Search packets can be received.

PR #267, opened 2017-11-30
samba-client: Add source-port 137/udp

### LINKED ISSUE #266: samba-client service needs source-port 137/udp
(With default-reject) I can't browse Windows shares even with samba-client service enabled unless I also add source-port 137/udp.

From log-denied:

```
FINAL_REJECT: IN=wlp1s0 OUT= MAC=[...] SRC=192.168.1.253 DST=192.168.1.104 LEN=90 TOS=0x00 PREC=0x00 TTL=64 ID=0 DF PROTO=UDP SPT=137 DPT=46872 LEN=70 
```

(Not sure MAC is sensitive, so I stripped it.)

dag:
Fixes #266 

I don't know if source-port 138/udp is also needed. For me browsing Windows shares in Nautilus works without it, and in the logs, any time I see SPT=138 I also see DPT=138 which is already covered in the current samba-client.service. I'm no expert at SMB/CIFS and I haven't tested any other case like printers. I opted not to add any rule I haven't confirmed as necessary.

I also don't know if any of this is needed for samba.service as well. My only observation is a `FINAL_REJECT` that blocks browsing remote Windows shares in Nautilus even with samba-client.service enabled, which seems like something it should allow.

### DIFF
--- a/config/services/samba-client.xml
+++ b/config/services/samba-client.xml
@@ -4,5 +4,6 @@
   <description>This option allows you to access Windows file and printer sharing networks. You need the samba-client package installed for this option to be useful.</description>
   <port protocol="udp" port="137"/>
   <port protocol="udp" port="138"/>
+  <source-port protocol="udp" port="137"/>
   <module name="nf_conntrack_netbios_ns"/>
 </service>


dag:
Although, doesn't matching on source-port mean that the source can access any (UDP) port on the destination by simply setting their source port to 137? If so, perhaps a warning in the description is in order, and make sure samba-client isn't enabled by default in any non-trusted zone (I guess it already isn't). You can also use rich rules to restrict samba-client further, for example to `192.168.1.1/24`. The new upnp-client.service would seem to suffer the same issue, and does include a warning in the description.

Of note is that samba-client is enabled by default in the FedoraWorkstation zone, and with this patch you could side-step the default-reject for ports 1-1024/udp, unless I've misunderstood how source-port works. An attacker does have to know to set their source port to 137, but security through obscurity isn't security and all that. Is there a way to express that if the source port is 137, the destination port has to be in the range >1024? Rich rules in firewall-config doesn't seem expressive enough; maybe with IPSets or with the CLI tools...

Is there a better way to fix #266, and the same for upnp-client?

dag:
So when I tried the IPSet `192.168.1.1/24,udp:32768-60999` I got "Set lan-ephemeral-udp is full, maxelem 65536 reached"; I guess IPSets are not the solution here.

(Side note: I think 32768-60999 is the range FedoraWorkstation should open instead of the wider 1025-65535, based on `/proc/sys/net/ipv4/ip_local_port_range` which IIUC is the dynamic range the kernel works with and what we'd need to open to make the Sharing settings in GNOME work. I could certainly be wrong.)

dag:
Of course with the FedoraWorkstation zone, none of this is necessary because the REJECT in #266 is let through by the rule for 1025-65535, so adding a rule for SPT=137 does nothing except to further open up the firewall. But at the same time, that's all incidental, and the fact remains that samba-client alone is insufficient to actually allow samba client traffic: you need either source-port 137 or port 32768-60999 (or 1025-65535) in addition to what's currently in samba-client.service.

erig0:
This would be better fixed by implementing OUTPUT filtering, see #237. Then the netbios_ns conntrack helper will allow the return traffic (source-port 137). Opening up the client to all traffic with source-port 137 seems like a bad idea.

erig0:
Closing as per my previous comment.

PR #272, opened 2017-12-26
services/amqp: Add AMQP and AMQPS services

jcpunk:
This commit adds service definitions for AMQP and AMQPS.


erig0:
@jcpunk, these look good. IANA also has the same numbered UDP port assigned as well - should we add UDP to the services? I don't know if it's common to use UDP with AMQP.

Issue #31, opened 2015-07-25
[GUI] firewall-config does not display zone description


genodeftest:
In firewall-config the zone description does not show up.

PR #275, opened 2017-12-28
Display selected zone description in firewall-config (issue #31)

MichaelAnckaert:
This commit displays the description of the selected zone in firewall-config

### DIFF
--- a/src/firewall-config
+++ b/src/firewall-config
@@ -399,6 +399,7 @@ class FirewallConfig(object):
         self.runtime_view = True
 
         self.zoneView = builder.get_object("zoneView")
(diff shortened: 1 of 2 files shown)


thom311:
Hi @MichaelAnckaert. Sorry for the very long time without reply.

I tried this patch, but it seems to me not nice to take so much prominent space in the GUI for the description. Also, the descriptions are of different lengths, which means that other widgets are moved around, depending on how much space the description takes.

I think, it's nicer to just show a tooltip in the zone list.

This would be done by https://github.com/firewalld/firewalld/pull/1197

I would close this PR and replace it by https://github.com/firewalld/firewalld/pull/1197.

PR #284, opened 2018-01-19
Add nmea-0183 service

da2x:
GPS data service offered by GPSD, Gypsy, gps-share, etc. and used by clients like GeoClue.

Port 10110 is registered for the nmea-0183 service with IANA.

### DIFF
--- a/config/services/nmea-0183.xml
+++ b/config/services/nmea-0183.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>nmea-0183</short>
+  <description>NMEA-0183 Navigational Data server for use with Global Navigation Satellite System (GNSS) devices.</description>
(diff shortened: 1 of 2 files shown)


config/services/nmea-0183.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>nmea-0183</short>
```
erig0: nmea-0183 is a serial communications protocol so I don't follow what it has to do with firewalld. Are you trying to open up ports for a server that uses nmea-0183? If so use the name of the server/daemon/product. nmea-0183 is nonsensical to firewalld.

da2x: This is used for servers sharing nmea-0183 data on the network (for sharing one GPS location or time devices with multiple clients, or for submitting sensor data)

PR #288, opened 2018-02-01
firewall/core/fw_transaction.py: Remove deduplication in add_rule (RH…

t-woerner:
…BZ#1534571)

Loading services from permanent configuration containing the same port numbers
results in desuplication of the rules. This then results in an error if the
second service gets removed from the zone.

This fixes RHBZ#1534571


hwoarang:
s/desuplication/deduplication/ :)

And I think that the RHBZ reference shouldn't be on the commit title since it's already references in the commit body.

erig0:
Thanks @t-woerner, I made the typo changes that @hwoarang pointed out and pushed to master. I also added a test case.

Issue #44, opened 2015-08-23
Can we move INPUT_direct and FORWARD_direct to the top of INPUT/FORWARD


divereigh:
I am running [rtpengine](https://github.com/sipwise/rtpengine) which runs a kernel module that essentially passes packets back out without it having to go through the user-space stack (for performance). However it needs to get in before the conntrack rule (otherwise the conntrack rule just accepts the packet and passes it into userspace, bypassing the kernel module).

I have tried using the --direct interface to add in a rule above the conntrack (in the INPUT chain), however it just gets added to the INPUT_direct chain instead, which comes after conntrack. I have had to resort to creating a new chain (INPUT_prefilter) using the --direct --passthrough. However this seems like a hack. Also anything done with --passthrough has to be --permanent otherwise it won't survive a --reload (is that a bug?).

So I see two solutions:
1) Move INPUT_direct so that it is the first in the INPUT chain
2) Create a INPUT_prefilter chain (as I have done) that is the first in the INPUT chain.

Then for completeness we should do the same with FORWARD.

I am happy to produce a patch to do either.

Thoughts?

t-woerner:
I thought about addition something similar as the prefiter flag also some time ago. But his would require to change the D-Bus interface of the direct interface in firewalld.

There is a solution without the need to change the interface: Use a passthrough rule to add a rule in the first line of the INPUT, .. chain. In this rule you can also jump into chains you have created with the --add-chain option in the same ipv and table.

divereigh:
I have to admit I don't enough about the D-Bus interface, but I'm guessing that rules out option 1. However is there any reason we can't do 2 - i.e. create a standard chain that runs first up before any conntrack.

I understand that you can use --passthrough - that is essentially what I say in the second paragraph above. The issue is that this would be hard to do in any sort of automatic install - for example in an init script. The --passthrough option only works if use --permanent (if you don't, your changes will be lost when you do a --reload), it is also hard to do a remove (essentially you have to do it by rule number - hoping that it is still number 1 in the INPUT chain).

It would be better to do it as a standard chain with a known (pre-defined) name. What happens if there are multiple kernel modules that wan't to get ahead of conntrack?

mstrhh:
There is another issue with the conntrack entry before any other chain:

I cannot block intruders while they are holding a connection! For example a drop entry in IN_public_deny will only take effect once the hacker starts a new connection.

 This looks like a major security problem to me.

marcosfrm:
[...]

PR #291, opened 2018-02-07
allow --delete rule via direct passthrough

josephtingiris:
Workaround for #44 as described by @josephtingiris.


erig0:
As of v0.6.0 this is no longer necessary. With the new nftables backend direct rules have higher precedence. For more info please read the blog post: https://firewalld.org/2018/07/nftables-backend

PR #301, opened 2018-02-28
firewall-config: Print error message when firewalld is not running

### LINKED ISSUE #299: firewall-config: Infinite wait loop when firewalld is stopped
When the `firewalld` service is stopped and we try to run the `firewall-config` UI, then it keeps waiting for connection with the `firewalld` daemon indefinitely. This is slightly annoying and it should really just timeout after a while or perhaps offer to start the `firewalld` service as an option.

hwoarang:
Fixes #299

### DIFF
--- a/src/firewall-config
+++ b/src/firewall-config
@@ -1580,7 +1580,7 @@ class FirewallConfig(object):
 
     def _dialog(self, text, msg=None, title=None,
                 message_type=Gtk.MessageType.INFO,
-                buttons=("gtk-close", 1)):
+                buttons=[("gtk-close", 1)]):
         dialog = Gtk.MessageDialog(parent=None, flags=0,
                                    message_type=message_type)
         dialog.set_markup(text)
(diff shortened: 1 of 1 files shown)


src/firewall-config
```diff
@@ -1614,7 +1618,15 @@ class FirewallConfig(object):
...
 
+    def connection_failed(self, msg):
+        if self._dialog("<b>"+_("Error")+"</b>",
+                        message_type=Gtk.MessageType.ERROR, msg=msg,
+                        buttons=[("gtk-quit", 1)]) == 1:
+            self.onQuit()
+
     def connection_changed(self):
+        if self.connection_timer:
+                GLib.source_remove(self.connection_timer)
```
erig0: Do we need to set self.connection_timer to None so we don't re-trigger source_remove() ?

hwoarang:
@erig0 yeah good point. I will fix it

erig0:
LGTM. Thanks for the fix @hwoarang.

PR #306, opened 2018-03-06
Mark some labels and headings as translatable

hwoarang:
Some of the UI labels and headings are not marked as translatable but there is probably no good reason for that so mark them as available for translation.

### DIFF
--- a/src/firewall-config
+++ b/src/firewall-config
@@ -184,7 +184,7 @@ class FirewallConfig(object):
         self.lockdownContextView = builder.get_object("lockdownContextView")
         self.lockdownContextStore = Gtk.ListStore(GObject.TYPE_STRING)
         self.lockdownContextView.append_column(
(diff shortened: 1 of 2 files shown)


src/firewall-config
```diff
@@ -279,11 +279,11 @@ class FirewallConfig(object):
                                               GObject.TYPE_STRING, # table
                                               GObject.TYPE_STRING) # chain
         self.directChainView.append_column(
-            Gtk.TreeViewColumn("ipv", Gtk.CellRendererText(), text=0))
+            Gtk.TreeViewColumn(_("ipv"), Gtk.CellRendererText(), text=0))
```
erig0: I don't think there in any way to translate "ipv" as it's not really a word, but more of a firewalld created term. What do you think?

hwoarang:
@erig0 you are right of course. I've updated the PR

PR #316, opened 2018-03-26
New samba-dc service

monnerat:
IP port reference: https://wiki.samba.org/index.php/Samba_AD_DC_Port_Usage


monnerat:
Thanks for merging :-)

PR #320, opened 2018-04-02
services: Add lightning-network related services

dougsland:
The following ports are used from lightning-network network.

lightning-network 9735/tcp
lightning-network-testnet 19735/tcp


cdecker:
Testnet does not have its own port since the lightning network is mostly agnostic of the underlying blockchain. We use the same port for bitcoin mainnet, bitcoin regtest, bitcoin testnet, and litecoin so far.

Issue #263, opened 2017-11-08
Module/Helper loading broken in Centos 7.4 for conntrack_proto_gre


sll552:
Hi,

I encountered an error in a previous release (CentOS 7.3) documented in #242. As recommended in this Issue, I created a service and it worked until I upgraded to CentOS 7.4.

I have the following service defined:
```xml
<?xml version="1.0" encoding="utf-8"?>
<service>
  <protocol value="gre"/>
  <module name="nf_conntrack_proto_gre"/>
</service>
```
This was enough to get firewalld to load the required module. 
After the upgrade I received the following error when trying to add the service to a zone:

> Error: INVALID_HELPER: proto-gre

I figured that I need to create a helper module which looks like this:
```xml
<?xml version="1.0" encoding="utf-8"?>
<helper module="nf_conntrack_proto_gre">
</helper>
```
But now I get different error:

> Error: INVALID_HELPER: 'nf_conntrack_proto_gre' is not available

By looking at https://github.com/firewalld/firewalld/blob/master/src/firewall/functions.py and the `get_nf_conntrack_helpers()` function, it seems like only conntrack modules which have an alias defined are considered. 
But unfortunately the conntrack module for gre doesn't have an alias as `modinfo` shows:
```
filename:       /lib/modules/3.10.0-693.2.2.el7.x86_64/kernel/net/netfilter/nf_conntrack_proto_gre.ko.xz
license:        GPL
rhelversion:    7.4
srcversion:     916FE0F82D2357B5A8B68BC
depends:        nf_conntrack
intree:         Y
vermagic:       3.10.0-693.2.2.el7.x86_64 SMP mod_unload modversions
signer:         CentOS Linux kernel signing key
sig_key:        C0:8B:35:53:41:76:0F:03:0A:64:E3:C5:50:E1:13:CD:47:77:1E:A9
sig_hashalgo:   sha256
```

Is it possible to work around this issue somehow, or is this a Centos issue?

Best regards

erig0:
IIIRC, all the other nf_conntrack_proto_* (e.g. sctp) are now built directly into the nf_conntrack module for newer kernels. I don't think firewalld currently handles nf_conntrack_proto_gre properly. I believe it's checking for the "helper" string in the modinfo output, but that does not present for nf_conntrack_proto_gre.

Try any of the following as a workaround:

1. manually load nf_conntrack_proto_gre before firewalld starts
[...]

PR #323, opened 2018-04-07
FIX: Module/Helper loading broken in Centos 7.4 for conntrack_proto_g…

kvishnivetsky:
FIX: Module/Helper loading broken in Centos 7.4 for conntrack_proto_gre #263

### DIFF
--- a/src/firewall/functions.py
+++ b/src/firewall/functions.py
@@ -343,9 +343,21 @@ def get_nf_conntrack_helpers():
             (status, ret) = runProg(COMMANDS["modinfo"], [ module, ])
             if status != 0:
                 continue
-            helper = filename.split(".")[0].strip()
-            helper = helper.replace("_", "-")
-            helpers.setdefault(module, [ ]).append(helper)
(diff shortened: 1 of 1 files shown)


kvishnivetsky:
Hi, Eric.
Ok - I agree, that maybe this is some kind of **workaround**, but what if we add a **configureation flag**, that will regulate this behaviour? Kernel fixup is rather long process, and this issue is rather actual - GRE is a popular proto.

erig0:
After speaking with colleagues, I think our best path forward is below.

In get_nf_conntrack_helpers():

- module name matches "nf_conntrack_proto_*"

or

- "-helper" in the "alias:" line of modinfo

Similarly for get_nf_nat_helpers():

- module name matches "nf_nat_proto_*"

or

- "NAT helper" in "description:" line

kvishnivetsky:
Ok. I'll make corrections in patch.

erig0:
Thanks @kvishnivetsky. I took your PR, squashed the commit, and added a couple fixes on top. I hope to push it soon - just waiting on testsuite to finish.

PR #328, opened 2018-04-13
Add matrix service

ykgmfq:
See https://github.com/matrix-org/synapse#using-a-reverse-proxy-with-synapse


ykgmfq:
I actually thought about that as well, I guess you're right.

Issue #80, opened 2016-02-05
rich rule masquerade needs to allow masqueraded packets on the zone


hwoarang:
Hi,

I am trying to use the rich rule masquerade functionality to only allow a specific subnet to see the inet via masquerade. `--add-masquerade` is not suitable because it does not do source based filtering so I use the following rich rule in the external zone (the one connected to inet)

`rule family=ipv4 source address=192.168.123.0/24 masquerade`

This seems to generate the following rule in the nat table

`-A POST_external_allow -s 192.168.123.0/24 ! -o lo -j MASQUERADE`

but there is no matching rule in the filter table. Here is a diff when i run `firewall-cmd --zone=external --add-masquerade` to the already running instance

```
 -A FWDO_external -j FWDO_external_deny
 -A FWDO_external -j FWDO_external_allow
 -A FWDO_external_allow -d 192.168.123.0/24 -m conntrack --ctstate NEW -j ACCEPT
+-A FWDO_external_allow -j ACCEPT
 -A FWDO_internal -j FWDO_internal_log
 -A FWDO_internal -j FWDO_internal_deny
 -A FWDO_internal -j FWDO_internal_allow
```

As you see, when you set `--add-masquerade` to the zone, a suitable ACCEPT target is created in the `FWDO_external_allow` zone which actually allows the forwarded packets.

I believe the previously mentioned rich rule should do the same thing since you want masquerade for the specified <source network> so a matching ACCEPT target should be created.

Using the following direct rule workarounds the problem and makes masquerading for the said <source network> to work as expected

`--add-rule ipv4 filter FWDO_external_allow 1 -s 192.168.123.0/24 -j ACCEPT`

What are your thoughts?

t-woerner:
The general ACCEPT rule is accepting everything. This is not correct and should not be needed. Is the traffic for 192.168.123.0/24 handled by the interface that is bound to the external interface? If you want to masquerade only 192.168.123.0/24, then bind this source to the external interface instead.

olebowle:
Hi,

I basically have the same question as hwoarang. I only want to masquerade 100.64.0.0/10. Currently it looks like this:
```
cpe_priv
  interfaces:
  sources: 100.64.0.0/10
  services: dhcp dns
  ports:
  masquerade: no
  forward-ports:
  icmp-blocks:
  rich rules:

external (active)
  interfaces: enp11s0
  sources:
  services: http https
  ports:
  masquerade: yes
  forward-ports:
[...]

PR #329, opened 2018-04-17
fw_zone: fix erroneous source/destination reversal in rich rule masqu…

erig0:
…erade

Specifying "source .." in rich rule masquerade was mistakenly reversed
causing it to match the destination instead. This is not what users
expect. It's expected that "source .." match something internal.

This allows "rule family=ipv4 source address=192.168.123.0/24
masquerade" to masquerade only a subset of the internal network.

Fixes: #80


erig0:
Code has changed quite a bit since I submitted this PR. The newer PR #352 fixes it for current code.

PR #339, opened 2018-05-16
Add Steam service

ykgmfq:
See https://support.steampowered.com/kb_article.php?ref=3629-RIAV-1617#networkports

@scaronni this is your XML file. Is it ok for me to submit this?

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -241,6 +241,7 @@ CONFIG_FILES = \
 	services/spideroak-lansync.xml \
 	services/squid.xml \
 	services/ssh.xml \
+	services/steam.xml \
 	services/svdrp.xml \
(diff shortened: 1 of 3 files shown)


scaronni:
Sure, but it's a proprietary product, don't know if it makes sense to add to the basic firewalld set of rules. It might change and nobody would notice.

config/Makefile.am
```diff
@@ -241,6 +241,7 @@ CONFIG_FILES = \
 	services/spideroak-lansync.xml \
 	services/squid.xml \
 	services/ssh.xml \
+	services/steam.xml \
```
erig0: I think "steam-streaming" is a better name as there are other ports used by steam for file downloads, etc.

PR #345, opened 2018-06-18
Add Zsh completion function

okapia:
This adds a completion function for zsh to complement the existing bash function.

I expect more can be done with respect to marking mutually exclusive options. Note that in order for zsh to pick this up when compinit is run, it should be installed into /usr/share/zsh/site-functions.

A second commit on the branch is a copy/paste of the bash function installation to install the function. I'm no expert on automake and could not test this properly because I get some errors about the AM_PO_SUBDIRS macro when running autogen.sh so I'm probably missing some dependency. Feel free to squash the commits or you may want to ignore the second one.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -69,6 +69,11 @@ AC_ARG_WITH([bashcompletiondir],
        [BASHCOMPLETIONDIR=$withval], [BASHCOMPLETIONDIR="${datadir}/bash-completion/completions"])
 AC_SUBST(BASHCOMPLETIONDIR)
 
+AC_ARG_WITH([zshcompletiondir],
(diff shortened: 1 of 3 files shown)


okapia:
I've pushed amended commits.

Naming it _firewalld is correct. I hadn't been aware of firewall-offline-cmd. I've adapted the function to complete for both commands. This is normal for zsh completion functions: they are usually named after the software rather than the command (e.g _subversion for svn and svnadmin) and switch on $service to handle different commands.

Sorry for the error in the original install script. These needed correction in exactly the manner you described and this time I did manage to test them after tracking down missing dependencies for the autogen.sh script to run.

PR #348, opened 2018-07-03
core: Replace 'icmp' with ipv6-icmp for IPv6 iptables rules

hwoarang:
When we select the protocol from the UI, the value is being taken from
the drop-down combobox. The combobox only lists 'icmp' to cover both
icmp and ipv6-icmp protocols but this results to icmp being injected
into the IPv6 chains so it does not work as expected. The solution to
that would be to scan the rule and replace 'icmp' instances with the
correct protocol for our IP version.

Link: https://bugzilla.suse.com/show_bug.cgi?id=1099698


src/firewall/core/ipXtables.py
```diff
@@ -401,7 +401,8 @@ def set_rules(self, rules, flush=False, log_denied="off"):
             self._rule_replace(rule, "%%REJECT%%", \
                     ["REJECT", "--reject-with", DEFAULT_REJECT_TYPE[self.ipv]])
 
-            # replace %%ICMP%%
+            # replace %%ICMP%% or icmp if injected by the UI
+            self._rule_replace(rule, "icmp", [ICMP[self.ipv]])
```
erig0: This would break a rich rule that was actually trying to match/drop icmpv4 in ipv6 (malicious packet). I'd rather see us introduce a keyword in the UIs to allow both, e.g. "icmpx". Then in build_zone_protocol_rules() we could just s/icmpx/%%ICMP%%/ and let the normal rule replace take care of it. I'm open for better ideas. :)

hwoarang: OK I see what you mean. So if I understand this correctly, you still want the ability to add 'icmp' to ipv6 chains to block bad packets. Do we also need an icmpv6 entry too for the ipv4 case? Maybe we can have the following entries instead

icmpv4 -> replaced by 'icmp' during rule build
icmpv6 -> no replacement
icmp -> replaced by 'icmpv6' or 'icmp' during rule build

make this will make it more clear from the UI perspective? What do you think?

erig0: > Do we also need an icmpv6 entry too for the ipv4 case?

It should already exist an "ipv6-icmp". Check /etc/protocols. It may not be pre-populated in firewall-config, but it should still be accepted by all the UIs.

hwoarang: So I thought about this a little bit more and what I did was to simply add 'ipv6-icmp' to the drop down menu to cover the IPv6 case. My understanding is that the 'Protocol' tab from the UI is not really appropriate to block the invalid icmp packets in the IPv6 case and users probably are going to use the Rich Rule for that. As such, the new ipv6-icmp protocol alone should get the job done. What do you think?

erig0: Sounds good to me!

hwoarang:
@erig0 oops yes I missed that too. Thank you for fixing it.

PR #349, opened 2018-07-04
core: logger: Remove world-readable bit from logfile

hwoarang:
The firewalld logfile was created with the 'read' bit set for all users
which could potentially reveal information about the firewall
configuration to unauthorised users. As such, we need to make use of
the low-level I/O facilities of the 'os' module to drop that bit during
file opening and creation.

Co-Authored-By: Matthias Gerstner <matthias.gerstner@suse.com>

### DIFF
--- a/src/firewall/core/logger.py
+++ b/src/firewall/core/logger.py
@@ -30,6 +30,7 @@ import syslog
 import traceback
 import fcntl
 import os.path
+import os
 
 # ---------------------------------------------------------------------------
 
@@ -134,7 +135,14 @@ class FileLog(LogTarget):
     def open(self):
         if self.fd:
             return
-        self.fd = open(self.filename, self.mode)
+        flags = os.O_CREAT | os.O_WRONLY
+        if self.mode.startswith('a'):
+            flags |= os.O_APPEND
+        self.fd = os.open(self.filename, flags, 0o640)
+        # Make sure that existing file has correct perms
+        os.fchmod(self.fd, 0o640)
+        # Make it an object
+        self.fd = os.fdopen(self.fd, self.mode)
         fcntl.fcntl(self.fd, fcntl.F_SETFD, fcntl.FD_CLOEXEC)
 
     def write(self, data, level, logger, is_debug=0):


src/firewall/core/logger.py
```diff
@@ -134,25 +135,30 @@ def __init__(self, filename, mode="w"):
     def open(self):
         if self.fd:
             return
-        self.fd = open(self.filename, self.mode)
+        flags = os.O_CREAT|os.O_WRONLY
+        if self.mode.startswith('a'):
+            flags |= os.O_APPEND
+        self.fd = os.open(self.filename, flags, 0o640)
+        # Make sure that existing file has correct perms
+        os.chmod(self.fd, 0o640)
```
erig0: You need to use fchmod() with an fd. chmod()'s first argument is a string/path. This is what is causing all the tests to fail.

hwoarang: oops you are right of course. I wonder why I missed it during testing. Thank you

src/firewall/core/logger.py
```diff
@@ -134,25 +135,30 @@ def __init__(self, filename, mode="w"):
...
+            flags |= os.O_APPEND
+        self.fd = os.open(self.filename, flags, 0o640)
+        # Make sure that existing file has correct perms
+        os.fchmod(self.fd, 0o640)
         fcntl.fcntl(self.fd, fcntl.F_SETFD, fcntl.FD_CLOEXEC)
 
     def write(self, data, level, logger, is_debug=0):
         if not self.fd:
             self.open()
-        self.fd.write(data)
```
erig0: Can you also avoid these changes by using os.fdopen() right after os.open()? I think it would also fix the remaining test failure triggered by the data.encode().

hwoarang:
Should be OK now. Thank you for the review!

erig0:
> Should be OK now. Thank you for the review!

Looks good now! Tests pass, merging. Thanks for working through it with me.

atgreen:
Any objection to making this 0600 instead of 0640?  We are working with security standards that are expecting 0600 on /var/log contents.

erig0:
> Any objection to making this 0600 instead of 0640? We are working with security standards that are expecting 0600 on /var/log contents.

I don't think there is any reason it couldn't be `0600`, but it's typically owned by `root` with group `root`.

```
# ls -al /var/log/firewalld 
-rw-r-----. 1 root root 21913 Oct 19 08:54 /var/log/firewalld
```

Any thoughts @hwoarang ?

hwoarang:
@erig0 I do not have any real thoughts on this. I suppose it's OK

erig0:
@atgreen, I would merge a PR that changes it to `0600`.

PR #351, opened 2018-07-16
Avoid telling NetworkManager to modify generated connections

lkundrak:
Preceded by some NetworkManager API improvements.


erig0:
Thanks Lubomir. I did some testing and this looks good! With this the interface to zone assignment is no longer delegated to NetworkManager. Instead firewalld handles it itself.

PR #352, opened 2018-07-17
Rich Rule Masquerade inverted source-destination in Forward Chain #278

### LINKED ISSUE #278: Rich Rule Masquerade inverted source-destination in Forward Chain
A rich-rule masquerade:

```
firewall-cmd --permanent --zone=public \
  --add-rich-rule 'rule family="ipv4" source address="192.168.2.0/24" masquerade'
```

```
<zone>
  <short>Public</short>
  ...
  <rule family="ipv4">
    <source address="192.168.2.0/24"/>
    <masquerade/>
  </rule>
</zone>
```

currently is generating a reversed source-destination rule in the FWDO chain:

```
-A POST_public_allow -s 192.168.2.0/24 ! -o lo -j MASQUERADE
-A FWDO_public_allow -d 192.168.2.0/24 -m conntrack --ctstate NEW -j ACCEPT
```

I expected instead that the resulting FWDO rule would have used the same source and destination of the POSTROUTING MASQUERADE rule:

```
-A FWDO_public_allow -s 192.168.2.0/24 -m conntrack --ctstate NEW -j ACCEPT
```

The [relevant code](https://github.com/firewalld/firewalld/blob/b20345ad5db13cf9a8ca8f5cb036ef526ab6693b/src/firewall/core/fw_zone.py#L1306-L1311) even has a comment on the need for source and destination inversion:

```python
    command = [ ]
    # reverse source/destination !
    self.__rule_source(rule.destination, command)
    self.__rule_destination(rule.source, command)
    command += [ "-m", "conntrack", "--ctstate", "NEW",
                 "-j", "ACCEPT" ]
```

I cannot understand why this inversion is needed though.
Even in a real-case scenario this is not allowing the relevant traffic to pass correctly.
On the other hand the same rule with non-reversed source and destination works as expected.

cc @t-woerner

sapensadler:
Fixes #278

### DIFF
--- a/src/firewall/core/ipXtables.py
+++ b/src/firewall/core/ipXtables.py
@@ -992,9 +992,8 @@ class ip4tables(object):
                                             zone=zone)
         rule_fragment = []
         if rich_rule:
-            # note the reversed source/destination
-            rule_fragment += self._rich_rule_destination_fragment(rich_rule.source)
-            rule_fragment += self._rich_rule_source_fragment(rich_rule.destination)
+            rule_fragment += self._rich_rule_destination_fragment(rich_rule.destination)
+            rule_fragment += self._rich_rule_source_fragment(rich_rule.source)
 
         rules.append([ add_del, "%s_allow" % (target), "-t", "filter"]
                      + rule_fragment +


sapensadler:
Worth noting this appears to be the same as https://github.com/firewalld/firewalld/issues/80 

fixes #80

erig0:
Thanks! I had fixed this is #329, but was waiting on someone to test. Seems like you've done that. :)

erig0:
Also worth noting that is only affected the iptables backend. The nftables backend is correct.

Issue #360, opened 2018-08-04
appstream-metadata-in-legacy-location usr/share/appdata/firewall-config.appdata.xml


mbiebl:
When building firewalld on Debian, I get the following warning from lintian, our package compliance checker:
```
W: firewall-config: appstream-metadata-in-legacy-location usr/share/appdata/firewall-config.appdata.xml
N: 
N:    AppStream metadata file was found in /usr/share/appdata/. The AppStream
N:    XML files should be placed in /usr/share/metainfo/.
N:    
N:    Refer to https://wiki.debian.org/AppStream/Guidelines for details.
N:    
N:    Severity: minor, Certainty: certain
N:    
N:    Check: appstream-metadata, Type: binary
N: 
```

PR #361, opened 2018-08-04
Use preferred location for AppData files

mbiebl:
The AppStream metadata files should be placed in /usr/share/metainfo/.

Refer to https://wiki.debian.org/AppStream/Guidelines for details.

Fixes: #360


erig0:
Merged. Thanks!

PR #368, opened 2018-08-08
Add Spotify local files sync service

ngdio:
Includes a service file which opens the ports needed by the Spotify client to sync local music files with the Android (maybe iOS too?) app.

### DIFF
--- a/config/services/spotify-sync.xml
+++ b/config/services/spotify-sync.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Spotify Client Sync</short>
+  <description>The Spotify Client allows you to sync local music files with your phone.</description>
+  <port port="57621" protocol="tcp"/>
+</service>
--- a/config/services/spotify-sync.xml
+++ b/config/services/spotify-sync.xml
@@ -2,5 +2,6 @@
 <service>
   <short>Spotify Client Sync</short>
   <description>The Spotify Client allows you to sync local music files with your phone.</description>
+  <port port="57621" protocol="udp"/>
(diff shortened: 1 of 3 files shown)


config/services/spotify-sync.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Spotify Client Sync</short>
+  <description>The Spotify Client allows you to sync local music files with your phone.</description>
+  <port port="57621" protocol="tcp"/>
```
erig0: There is a [negativo17/spotify-client](https://github.com/negativo17/spotify-client) project that provides a spotify service definition. It has the following for ports:
```
<port protocol="tcp" port="57621-57622"/>
<port protocol="udp" port="57621-57622"/>
```

This jives with other sources, e.g. https://mrlithium.blogspot.com/2011/10/spotify-and-opting-out-of-spotify-peer.html.

Unfortunately I could not find an official port list from Spotify.

ngdio: The description of the project seems to be inaccurate - I only needed tcp/57621 (not sure where exactly I found that) to make the local file sync work. Even with their service definition, my client seems to be unable to find the one Spotify Connect device I have in my network (as opposed to the Android client).

erig0: I see my spotify client with a UDP socket open on the same port. I think we should open UDP as well.

```
$ lsof -p 5297  |awk '/57621/{ print $8 " " $9 " " $10}'
UDP *:57621 
TCP *:57621 (LISTEN)
```

ngdio: Done.

Issue #371, opened 2018-08-17
firewalld doesn't fully support netroot=iscsi


wenjianhn:
Backend: iptables

How to reproduce:
1. boot a kernel with netroot=iscsi
2. firewall-cmd --add-masquerade --permanent 
3. firewall-cmd --reload

Expected result:  It prints success
Actual: It blocks forever.

Root cause:
    Currently, while reloading the policy for built-in chains is set to
    DROP for all the tables and is then reset to ACCEPT at the end.
    Because the policy of the raw/mangle/security table is set to DROP
    and lack of ACCEPT rules, connection interruption is possible during
    the reload. libipt_MASQUERADE.so was not in page cache, so iscsid
need to fetch it using the network. But both its inbound and outbound
 traffic and dropped.

Workaround:
Instead of reloading firewalld, make a command both persistent and take effect immediately.
e.g.
` firewall-cmd --add-masquerade --permanent` 
 `firewall-cmd --add-masquerade`

erig0:
I think this would work as expected if a reload were to set policy to drop, but at the same time permit existing connections (i.e. ct state == EST). I think this is reasonable on a reload.

What do you think?

wenjianhn:
The rules that are used to allow existing connections are added in the filter table.
Inbound traffic will never hit the rules because the raw table, which has higher priority, 
will drop the packets first.

I think even the localhost traffic are dropped during reloading.

erig0:
@wenjianhn, right. I was suggesting replacing the DROP policy on reload with something that _only_ allows existing connections. We would not drop existing ones at the raw table.

wenjianhn:
Got it.

To do that, I think we should only update the policy of the filter table as
DROP.
And add new default rules (state ESTABLISHED and maybe RELATED) for the
INPUT, FORWARD and OUTPUT chains of the filter table to only allow existing
connections.

And for performance reason we may need to delete the rules after reloading.
For the most cases we only need ACCEPT for the OUTPUT chain. The ESTABLISHED
rule is kind of a waste.

The risk is that raw/mangle/security table will ACCEPT packets by default.

I don't know if it makes everybody happy. We need an config option.
What do you think?

erig0:
Fixed by 67aa1ce6a965f617774f5af7b8a28782a41a4cbd.

erig0:
> The risk is that raw/mangle/security table will ACCEPT packets by default.

Packets are still subject to the later occurring netfilter hooks.

PR #372, opened 2018-08-17
Opt drop on reload

wenjianhn:
issue https://github.com/firewalld/firewalld/issues/371


src/firewall/core/fw.py
```diff
@@ -428,6 +428,7 @@ def _start(self, reload=False, complete_reload=False):
         # exsting ipsets and the configuration in firewalld.
         if (reload and complete_reload) or \
            (self.ipset_enabled and self.ipset.has_ipsets()):
+            # FIXME(security issue): if self.drop_on_reload, all traffic are ACCEPT now
```
wenjianhn: The new opt doesn't work with ipset_enabled.

erig0:
As discussed in #371, it would be better to allow existing connections during a reload and only drop new connections. I think this is the correct approach instead of "featuring" DropOnReload. As such, I'm closing this PR and leaving issue #371 open.

Issue #370, opened 2018-08-16
firewall/core/fw_nm: nm_get_zone_of_connection returns False


Farenjihn:
Hello,

I use NetworkManager which automatically creates a connection for the `docker0` interface. In its current state, `nm_get_zone_of_connection` will return `False` for a generated connection and thus trigger a `ValueError` when running `firewall-config` (in `update_active_zones`).

It seems to me that `nm_get_zone_of_connection` should either not return a Boolean in case of error or each call should check whether it has succeeded before carrying along with string comparisons.

Here is the trace for reference:

```
Traceback (most recent call last):
  File "/bin/firewall-config", line 8093, in <module>
    app = FirewallConfig()
  File "/bin/firewall-config", line 1386, in __init__
    self.connection_changed()
  File "/bin/firewall-config", line 1692, in connection_changed
    self.update_active_zones()
  File "/bin/firewall-config", line 2464, in update_active_zones
    connection, zone ])
  File "/usr/lib/python3.7/site-packages/gi/overrides/Gtk.py", line 1245, in append
    return self._do_insert(parent, -1, row)
  File "/usr/lib/python3.7/site-packages/gi/overrides/Gtk.py", line 1237, in _do_insert
    row, columns = self._convert_row(row)
  File "/usr/lib/python3.7/site-packages/gi/overrides/Gtk.py", line 876, in _convert_row
    result.append(self._convert_value(cur_col, value))
  File "/usr/lib/python3.7/site-packages/gi/overrides/Gtk.py", line 894, in _convert_value
    return GObject.Value(self.get_column_type(column), value)
  File "/usr/lib/python3.7/site-packages/gi/overrides/GObject.py", line 218, in __init__
    self.set_value(py_value)
  File "/usr/lib/python3.7/site-packages/gi/overrides/GObject.py", line 274, in set_value
    (py_value, type(py_value)))
ValueError: Expected string but got False<class 'bool'>
```

Am I missing something?

Thanks

erig0:
Fixed by #373.

PR #373, opened 2018-08-17
firewall/core/fw_nm: nm_get_zone_of_connection return type fix

Farenjihn:
Hello,

As explained in #370, when using NetworkManager the automatically generated connections (such as `docker0` for the interface with the same name) make `nm_get_zone_of_connection` return False which in turn makes `firewall-config` crash.

This pull request basically changes the return values of `nm_get_zone_of_connection` so that it matches the old behavior in version 0.6.0 by returning None, except for a generated connection in which case it returns the empty string so the default zone is used by firewalld.

Let me know if this is useful.

Thanks

### DIFF
--- a/src/firewall/core/fw_nm.py
+++ b/src/firewall/core/fw_nm.py
@@ -75,21 +75,21 @@ def nm_get_zone_of_connection(connection):
 
     con = nm_get_client().get_connection_by_uuid(connection)
     if con is None:
-        return False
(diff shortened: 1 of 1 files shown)


erig0:
> This fix should be cherry-picked for the stable-0.6 branch

Also done.

mbiebl:
> Also done.

thanks!

PR #379, opened 2018-08-29
Fix Neighbor Discovery when using drop zone

hwoarang:
This adds two commits needed to fix the ND when an interface belongs to the drop zone

- Patch#1 Accepts neighbour-advertisement message to the 'raw' table
- Patch#2 Accepts all needed ND packets in the filter table which is needed if the default policy on the filter table is 'DROP/REJECT' (eg drop zone)

### DIFF
--- a/src/firewall/core/ipXtables.py
+++ b/src/firewall/core/ipXtables.py
@@ -1140,9 +1140,13 @@ class ip6tables(ip4tables):
                        "--icmpv6-type=router-advertisement",
                        "-j", "ACCEPT" ]) # RHBZ#1058505
         rules.append([ "-I", "PREROUTING", "3", "-t", "raw",
+                       "-p", "ipv6-icmp",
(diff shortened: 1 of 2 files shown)


herrold:
This needs more thought

We have noticed that in a VM environment, and with people using third party IPv6 tunnelling, that they (perhaps unwillingly) also turn on their radvd, and so pollute the local and accurate advertisements from an authoritative radvd

We need to be able to * and under iptables / ebtables DO * filter out such 'rogue' advertisements

hwoarang:
Back you can still add your rules on the `raw_prerouting` chain to actually drop these packets right? Dropping something in any table which drop the packet entirely, whereas accepting it in a table, will still make it traverse the rest of them.

erig0:
@hwoarang, please see PR #380 for an alternative solution.

erig0:
Closed in favor of alternative fix #380.

PR #380, opened 2018-09-05
*tables: For opened ports/protocols/etc match ct state new,untracked

erig0:
Some valid packets may be untracked and as such "ct state new" won't
match them. A prime example of this is IPv6 neighbor-advertisements. Now
adding the "icmp" service to a zone will match and accept these
untracked packets.

Note: We don't want to fully remove ct state match because we don't want
to match invalid states. This is especially important for the forward
chain.

Fixes: suse bz 1105821


erig0:
@hwoarang, please review.

hwoarang:
@erig0 I think it looks reasonable

Issue #344, opened 2018-06-08
firewall-config - Do not add rules in the Rich Rules tab, do not release the OK button.


clayzanfolin:
Hello!

Firewalld does not create the rules in the Rich Rules tab in graphical interface in a user account and putting the root password when requested or using the 'firewall-config' command logged in root just with 'su'. Now if you add 'su -' or log in to a GUI as root, everything works. Example of problem:

In the graphical interface in zones >> public, in the Rich Rules tab click + Add, ipv4 rule check Action boxes with the accept and source ip 192.168.0.1 item. Does not release the OK button.

Even marking the Element box and adopting a service does not work.

It was reported in bugzilla.opensuse because the problem is in openSUSE Leap 15.0:
https://bugzilla.opensuse.org/show_bug.cgi?id=1096542

I also did a video demonstrating the problem:
https://youtu.be/nKziaIjN3wI

I hope I have provided all the information you need!
Up until!

hwoarang:
After some investigation this seems to only affect non-English and non-French translations. Not quite sure why a translation would make the UI misbehave to be honest.

clayzanfolin:
Hello!

Unfortunately I'm not a programmer, and I do not know how to help, but casso has to do some testing or something like that, I'm ready to help.

Up until!

erig0:
Hi @clayzanfolin,

Thanks for the video. It was very helpful.

I think I found the cause. There is untranslated string used when editing existing rich rules. Later on this untranslated string is compared against a translated string to see if the rich rule is valid. I think we can quickly fix this instance by translating the strings.

However, keying off these translated strings is fragile at best. We should look at real fix for this issue.

Note: It works for some locales because the language translations haven't been done for some of these strings, e.g. in the French case "accept" is translated to "accept".

Untranslated strings:
https://github.com/firewalld/firewalld/blob/6a3ac5835e26980098b1226313984d667cd84918/src/firewall-config.in#L3198-L3201

Comparison to translated string to test if rich rule is valid:
https://github.com/firewalld/firewalld/blob/6a3ac5835e26980098b1226313984d667cd84918/src/firewall-config.in#L3692-L3693

PR #392, opened 2018-09-20
Fix translating labels

MeggyCal:
Fix of https://github.com/firewalld/firewalld/issues/344 was incomplete, the "flags" were not translating and the reported bug was still active. I am proposing a patch which solves the issue for me.


erig0:
Thanks @MeggyCal !

PR #393, opened 2018-09-20
NM backported fixes

hwoarang:
Hello,

These backported patches are needed in order for firewalld-0.5.X to work better with NetworkManager


erig0:
Hi @hwoarang, do you have an downstream bug for these? I'm curious what the report is.

hwoarang:
@erig0 yeah this is the bug report http://bugzilla.suse.com/show_bug.cgi?id=1109074

Issue #303, opened 2018-03-02
Avoid UnicodeDecoderError: use binary stream for xml files, let SAX determine the encoding


StefanBruens:
Currently  zone_reader from zone.py:
https://github.com/firewalld/firewalld/blob/10bcda63d348d1cc8d930a07c7edaf88e0536ed5/src/firewall/core/io/zone.py#L699

and *_reader from the other io classes open the file as character stream, which fails as soon as e.g. the "C" locale is used and the xml files contains non-ASCII character.

SAX allows to feed a binary stream into the parser, and let SAX determine the encoding from the XML encoding pseudo attribute.

The following code works with both Python2 and Python3:
```
#! /usr/bin/python3
# -*- coding: utf-8 -*-

import xml.sax as sax
import codecs

name = "./foo.xml"
parser = sax.make_parser()

with open(name, "rb") as f:
  source = sax.InputSource(None)
  source.setByteStream(f)
  try:
    parser.parse(source)
  except Exception as e:
    print(e)
```

This is able to parse the following XML file:
```
<?xml version="1.0" encoding="utf-8"?>
<foo>
  <a>
    asdf
  </a>
  <b>
    äöü
  </b>
</foo>
```

erig0:
Can you tell me what version of python you're using? I can reproduce this on 3.5, but not 3.7

StefanBruens:
Currently 3.6.5, not sure which version I used in march.

On 3.7 it probably works cause of https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep538

erig0:
> On 3.7 it probably works cause of https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep538

Indeed. Thanks for the pointer.

PR #395, opened 2018-09-21
firewall/core/io/*.py: Let SAX handle the encoding of XML files

StefanBruens:
SAX is able to determine the encoding of XML files itself if the file
contains a correct "encoding" pseudo attribute, e.g.:
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>

For this to work, the file stream has to be opened in binary mode, and
the parser has to read the stream using a SAX InputStream, which
autodetects the encoding.

Fix for #303.

### DIFF
--- a/src/firewall/core/io/direct.py
+++ b/src/firewall/core/io/direct.py
@@ -360,9 +360,11 @@ class Direct(IO_Object):
         handler = direct_ContentHandler(self)
         parser = sax.make_parser()
         parser.setContentHandler(handler)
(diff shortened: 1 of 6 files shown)


StefanBruens:
> This looks good. Can you also make the change in `src/firewall/core/io/lockdown_whitelist.py` ?

`lockdown_whitelist.py` uses `parser.parse(filename)`, not `parser.parse(stream)`, and thus is already working correctly.

PR #400, opened 2018-10-01
Add BOOTP services

jcpunk:
For RH Sat6 install guide it recommends opening these ports for some hosts.  I've added service definitions here so that they can be called out by name.

### DIFF
--- a/config/services/bootp-client.xml
+++ b/config/services/bootp-client.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>bootp-client</short>
+  <description>The Bootstrap Protocol (BOOTP) is a computer networking protocol used in Internet Protocol networks to automatically assign an IP address to network devices from a configuration server.</description>
+  <port protocol="udp" port="68"/>
+  <port protocol="tcp" port="68"/>
+</service>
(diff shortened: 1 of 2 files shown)


config/services/bootp-client.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>bootp-client</short>
+  <description>The Bootstrap Protocol (BOOTP) is a computer networking protocol used in Internet Protocol networks to automatically assign an IP address to network devices from a configuration server.</description>
+  <port protocol="udp" port="68"/>
+  <port protocol="tcp" port="68"/>
```
erig0: I don't think adding TCP here makes sense.

config/services/bootp-server.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
erig0: Note, this is the same as `config/services/dhcp.xml`. Just noting, I think it's fine to have both.

config/services/bootp-server.xml
```diff
@@ -0,0 +1,7 @@
...
+<service>
+  <short>bootp-server</short>
+  <description>The Bootstrap Protocol (BOOTP) is a computer networking protocol used in Internet Protocol networks to automatically assign an IP address to network devices from a configuration server.</description>
+  <port protocol="udp" port="67"/>
+  <port protocol="tcp" port="67"/>
```
erig0: Again, I don't think TCP makes sense here.

jcpunk:
The TCP bits look to be a CISCO specific extension, don't merge

PR #401, opened 2018-10-01
Modern TFTP servers can also use TCP transport.

jcpunk:
I've noticed some modern TFTP servers also support use of TCP.  This is also noted in /etc/services on RHEL7.


erig0:
Can you give an example tftp server that also listens on TCP?

jcpunk:
The TCP bits look to be a CISCO specific extension, don't merge

PR #403, opened 2018-10-05
services: renamed RH-Satellite-6 to foreman

lzap:
We like to reference our project rather than product. Also, I've added missing
8443 port which is required:

https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#ports_prerequisites

### DIFF
--- a/config/services/foreman.xml
+++ b/config/services/foreman.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <service>
-  <short>Red Hat Satellite 6</short>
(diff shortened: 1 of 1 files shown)


erig0:
> Will do. Is there any kind of alias mechanism so I could possibly prevent braking user commands?

Unfortunately not. Probably best to retain `RH-Satellite-6` and add `foreman` as a completely new service.

lzap:
Looks like the port is really incorrect, so the XML is all correct. I am gonna go ahead and close this since just renaming for nothing is really not what I want to do. Thanks!

PR #410, opened 2018-10-20
ipXtables/nftables: Fix "object has no attribute '_log_denied'"

fedux:
This fixes nftables and ipXtables (when IndividualCalls=yes),
as _log_denied is not an attribute of the class but a param.


erig0:
Thanks for the fix @fedux. I'm also pushing test coverage for this.

PR #412, opened 2018-10-25
firewall/core/fw_nm: nm_get_zone_of_connection should return None or …

hwoarang:
…empty string instead of False

(cherry picked from commit 5a59a90f449a8bf836e62e2d9ad486301b1aa2bb)


erig0:
Ah. Looks like the inital cherry-pick of this was a partial. Can you note that in the commit message?

hwoarang:
@erig0 I can't see any previous cherry-pick of this patch so I am not exactly sure what you mean :)

PR #415, opened 2018-10-27
minimize modinfo calls

maage:
Firewalld runs modinfo per interesting module. With this patch it calls modinfo with multiple modules. This should lower firewalld startup time slightly.

To call get_nf_conntrack_helpers and get_nf_nat_helpers 100 times each without my patch takes:

real	0m10.087s
user	0m5.008s
sys	0m5.895s

and with my patch:

real	0m3.319s
user	0m2.621s
sys	0m0.784s

It would be possible to have only one modinfo call, but that would require to modify other modules.

### DIFF
--- a/src/firewall/functions.py
+++ b/src/firewall/functions.py
@@ -36,6 +36,7 @@ import os
 import os.path
 import shlex
 import pipes
+import re
 import string
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


erig0:
Thanks for the quality PR, and sorry for the delay.

PR #434, opened 2018-11-27
Move D-Bus policy file to /usr/share/dbus-1/system.d/

mbiebl:
This is supported since dbus 1.9.18.
The old location in /etc/dbus-1/system.d/ has been deprecated.

https://lists.freedesktop.org/archives/dbus/2015-July/016746.html

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -22,7 +22,7 @@ polkit1_action_FILES = org.fedoraproject.FirewallD1.server.policy.in \
 polkit1_actiondir = $(datadir)/polkit-1/actions
 polkit1_action_DATA = $(polkit1_action_FILES:.in=)
 
-dbus_policydir = $(sysconfdir)/dbus-1/system.d
+dbus_policydir = $(datadir)/dbus-1/system.d
 dist_dbus_policy_DATA = FirewallD.conf
 
 gsettings_in_file = org.fedoraproject.FirewallConfig.gschema.xml.in


mbiebl:
dbus 1.9.18 has been released 2015 and is available in stable distros like Debian Stretch or Ubuntu 16.04.

mbiebl:
I filed this bug report after talking to D-Bus upstream, @smcv (I hope he actively uses that handle)
Let's bring im into this, if you have specific concerns about this change, maybe he can answer them.

smcv:
Since 1.10, the intention is that after a transitional period, software will install its defaults in `${datadir}`, reserving `${sysconfdir}` for local sysadmin overrides (the same as the way packages like systemd use /lib and /etc). https://dbus.freedesktop.org/doc/dbus-api-design.html#security-policies mentions this recommendation, although the NEWS doesn't yet.

There are two ways distro packages could work during that transitional period:

* Upstream packages continue to install in `${sysconfdir}` by default, and distros that know they have dbus 1.10 or later can move those files into `/usr/share` before packing the built software into a package (for example Debian packages could put `etc/dbus-1/system.d usr/share/dbus-1` in `debian/firewalld.install`)

* Upstream packages install to `${datadir}` by default, and distros that still support (backports to) branches with dbus versions older than 1.10 can move those files into `/etc` before packing the built software into a package (for example Debian packages could have put `usr/share/dbus-1/system.d etc/dbus-1` in `debian/firewalld.install`)

In the past we've had the first of those (more by inertia than anything else), but now that dbus 1.10 is widely available in long-term-supported/"enterprise" distros (Debian >= 9, Ubuntu LTS >= 16.04, and apparently RHEL/CentOS >= 7.5), I think it would make sense to move towards the second.

smcv:
I think dbus 1.10 is now sufficiently widespread that it would make sense to say that third-party software installing dbus policies into `${sysconfdir}/dbus-1/*.d` is considered to be deprecated, leaving those directories reserved for the sysadmin.

(dbus 1.10 isn't the current stable branch any more - that's dbus 1.12, released in October 2017.)

erig0:
Merged. Thanks @mbiebl and @smcv.

Issue #457, opened 2019-02-11
In firewall-config cannot edit rich rule added with firewall-cmd


dalanicolai:
I added a rich rule to whitelist a single ip-address with firewall-cmd using the following command:

sudo firewall-cmd --zone=home --add-rich-rule="rule family="ipv4" source address="242.110.155.196" port protocol="tcp" port="4567" log prefix="test-firewalld-log" level="info" accept"

When I subsequently tried to edit the rich rule in firewall-config I got the following traceback message

Traceback (most recent call last):
  File "/usr/bin/firewall-config", line 3065, in onEditRichRule
    self.add_edit_rich_rule(False)
  File "/usr/bin/firewall-config", line 3247, in add_edit_rich_rule
    old_obj.source.invert)
TypeError: Argument 1 does not allow None as a value`

erig0:
Fixed by 7da05ef, please remember to use the `Fixes:` tag on PRs. Althought I forgot to check for it in my review. :)

klemenkobetic:
I get the exact same error when trying to edit a rich rule in firewall-config  :+1: 
```
Traceback (most recent call last):
  File "/usr/bin/firewall-config", line 3074, in onEditRichRule
    self.add_edit_rich_rule(False)
  File "/usr/bin/firewall-config", line 3265, in add_edit_rich_rule
    old_obj.destination.invert)
TypeError: Argument 1 does not allow None as a value
```

CentOS Stream release 8
firewalld-0.9.3-13.el8.noarch
firewall-config-0.9.3-13.el8.noarch

dfdario:
In firewalld-config by trying to edit existing rich-rule 'rule family=ipv4 source not address=192.168.5.0/24 drop' error arisen is:

 File "/usr/lib/python3.12/site-packages/firewall/core/rich.py", line 660, in __post_init__
    self._import_from_string(rule_str)
  File "/usr/lib/python3.12/site-packages/firewall/core/rich.py", line 817, in _import_from_string
    raise FirewallError(
firewall.errors.FirewallError: INVALID_RULE: unknown element ipv4

Furthermore there is no way to add a new rich-rule via the interface but the console.

PR #458, opened 2019-02-11
fix issue #457

dalanicolai:
I found out I did not set a value for invert when adding the rich rule via firewall-cmd. Then I got the error as mentioned in issue #457 because the invert attribute was given a default value None. I corrected it here so that it gets the default value False. This fixed the issue for me.


erig0:
Thanks @dalanicolai !

PR #460, opened 2019-02-13
Add pandorafms-server port

giannakopoulosj:
PandoraFMS server uses the 41121 port to accept data from its clients


erig0:
We prefer to use the protocol names if they're given by IANA. In this case it's "tentacle".

erig0:
I manually merged this with the following changes:
- renamed to tentacle
- updated `config/Makefile.am` and `po/POTFILES.in`

PR #464, opened 2019-02-19
Add Prometheus service

hos7ein:
Open tcp port 9090 for Prometheus service.


erig0:
According to 
  https://github.com/prometheus/prometheus/wiki/Default-port-allocations
there are other ports for "pushgateway" and "alertmanager". Do those also need to be opened. I also don't know about the many "exporters".

hos7ein:
In this case other ports is not need.
"exporters" is Separate of Prometheus server.

PR #465, opened 2019-02-19
Add Grafana service

hos7ein:
Open tcp port 3000 for Grafana service.


erig0:
Manually merged so I could update `po/POTFILES.in`.

PR #468, opened 2019-02-27
adds memcached service at tcp/udp port 11211

### LINKED ISSUE #467: add memcached service
I found myself missing a `memcache` service, and (since there's no protocol number for memcache)` firewall-cmd --add-protocol=memcache` (at least on CentOS 7) gives:

`Error: INVALID_PROTOCOL: memcache `

Therefore I'm submitting a PR adding a`memcache` service.

joshrabinowitz:
Closes #467

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -186,6 +186,7 @@ CONFIG_FILES = \
 	services/managesieve.xml \
 	services/matrix.xml \
 	services/mdns.xml \
+	services/memcached.xml \
 	services/minidlna.xml \
 	services/mongodb.xml \
 	services/mosh.xml \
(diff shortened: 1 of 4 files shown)


erig0:
I don't use the project, but it seems like it's always refered to as memcache**d**.

joshrabinowitz:
@erig0 I wasn't sure which way to go, I had originally called it memcache, but `/etc/services` calls it `memcache`, so I went with that.

Switching it back to 'memcached'

erig0:
> @erig0 I wasn't sure which way to go, I had originally called it memcached, but `/etc/services` calls it `memcache`, so I went with that.
> 
> Switching it back to 'memcached'

Ah, I didn't realize it was listed in `/etc/services`. I suppose we should go with what's there. Don't worry about changing it yet again. I'll take care of it.

Thanks for adding a new service!

Issue #435, opened 2018-12-03
FreeIPA service definitions are outdated


tiran:
firewalld comes with four pre-defined service definitions for FreeIPA:

* freeipa-ldap with HTTP, HTTPS, Kerberos, LDAP and NTP
* freeipa-ldaps with HTTP, HTTPS, Kerberos, LDAPS, and NTP
* freeipa-replication with 7389/TCP
* freeipa-trust with Samba and AD ports

However the service definitions and documentation are outdated:

* Port 7389 and service ``freeipa-replication`` are no longer needed since RHEL 7.0 and Fedora 18. Port 7389/TCP was used for Dogtag's LDAP server. FreeIPA 3.1 consolidated the 389-DS databases and now Dogtag and IPA share the same LDAP server instance.
* FreeIPA requires both LDAP and LDAPS to work properly. Iit doesn't make any sense to have separate services ``freeipa-ldap`` and ``freeipa-ldaps``. FreeIPA clients and master-master replication both require port 389/TCP (LDAP) to work. The LDAP protocol is not a security problem, because FreeIPA doesn't use plaintext LDAP but STARTTLS and/or GSSAPI to authenticate and protect. Dogtag PKI also uses LDAPS.
* FreeIPA 4.7 and newer no longer install a NTP instance on each master.

I propose a single freeipa service ``freeipa.xml`` that combines LDAP and LDAPS but leaves out NTP:

```freeipa.xml
<?xml version="1.0" encoding="utf-8"?>
<service>
  <short>FreeIPA server</short>
  <description>FreeIPA is an integrated identity and authentication solution with Kerberos, LDAP, PKI, and web UI. Enable this option if you plan to provide a FreeIPA server. Enable the 'dns' service if this FreeIPA server provides DNS services and 'ntp' service if this FreeIPA server provides NTP services.</description>
  <!-- CRL and OCSP -->
  <port protocol="tcp" port="80"/>
  <!-- API and web UI -->
  <port protocol="tcp" port="443"/>
  <!-- Kerberos and kpasswd -->
  <port protocol="tcp" port="88"/>
  <port protocol="udp" port="88"/>
  <port protocol="tcp" port="464"/>
  <port protocol="udp" port="464"/>
  <!-- LDAP and LDAPS -->
  <port protocol="tcp" port="389"/>
  <port protocol="tcp" port="636"/>
</service>
```

erig0:
[...]

PR #471, opened 2019-03-01
Add firewall rules for FreeIPA 4.x

tiran:
The firewall services ``freeipa-ldap``, ``freeipa-ldaps``, and
``freeipa-replication`` are outdated. FreeIPA 4.x requires a different set
of rules.

* Port 7389 and service ``freeipa-replication`` are no longer needed since
  RHEL 7.0 and Fedora 18. Port 7389/TCP was used for Dogtag's LDAP server.
  FreeIPA 3.1 consolidated the 389-DS databases and now Dogtag and IPA
  share the same LDAP server instance.
* FreeIPA requires both LDAP and LDAPS to work properly. Iit doesn't make
  any sense to have separate services ``freeipa-ldap`` and
  ``freeipa-ldaps``. FreeIPA clients and master-master replication both
  require port 389/TCP (LDAP) to work. The LDAP protocol is not a security
  problem, because FreeIPA doesn't use plaintext LDAP but STARTTLS and/or
  GSSAPI to authenticate and protect. Dogtag PKI also uses LDAPS.
* FreeIPA 4.7 and newer no longer install a NTP instance on each master.

Fixes: https://github.com/firewalld/firewalld/issues/435
Signed-off-by: Christian Heimes <cheimes@redhat.com>

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -147,6 +147,7 @@ CONFIG_FILES = \
 	services/etcd-client.xml \
 	services/etcd-server.xml \
 	services/finger.xml \
+	services/freeipa-4.xml \
 	services/freeipa-ldaps.xml \
 	services/freeipa-ldap.xml \
 	services/freeipa-replication.xml \
--- a/config/services/freeipa-4.xml
+++ b/config/services/freeipa-4.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>FreeIPA 4 server</short>
+  <description>FreeIPA is an integrated identity and authentication solution with Kerberos, LDAP, PKI, and web UI. Enable this option if you plan to provide a FreeIPA server. Enable the 'dns' service if this FreeIPA server provides DNS services, 'ntp' service if this FreeIPA server provides NTP services, and 'freeipa-trust' for cross-forest trusts with Active Directory.</description>
+  <!-- CRL and OCSP -->
+  <port protocol="tcp" port="80"/>
+  <!-- API and web UI -->
+  <port protocol="tcp" port="443"/>
+  <!-- Kerberos and kpasswd -->
+  <port protocol="tcp" port="88"/>
+  <port protocol="udp" port="88"/>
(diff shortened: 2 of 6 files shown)


erig0:
@tiran, this PR looks fine. What about freeip-trust? Is that still relevant?

tiran:
@erig0 freeipa-trust is still relevant. It's used to establish trust between an IPA system and a Windows AD system.

@abbra could you please verify the rules in https://github.com/firewalld/firewalld/blob/master/config/services/freeipa-trust.xml ?

abbra:
`freeipa-trust.xml` has outdated references to the dynamically opened RPC ports.
Current defaults in Samba 4.7+:
```
rpc server dynamic port range = 49152-65535
```

erig0:
Does it make sense for `freeipa-trust` to carry a major version number as well, e.g. `freeipa-4-trust`?

Adding the ports that @abbra mentions, I think the new `-trust` service should look like:
```
<service>
  <short>FreeIPA-4 trust setup</short>
  <description>FreeIPA is an LDAP and Kerberos domain controller for Linux systems. Enable this option of you plan to deploy cross-forest trusts with FreeIPA and Active Directory</description>
  <port protocol="tcp" port="135"/>
  <port protocol="tcp" port="138-139"/>
  <port protocol="udp" port="138-139"/>
  <include service="ldap" />
  <port protocol="tcp" port="445"/>
  <port protocol="udp" port="445"/>
  <port protocol="tcp" port="3268"/>
  <port protocol="tcp" port="49152-65535"/>
</service>
```

I assume we should remove `<port protocol="tcp" port="1024-1300"/>`, is that correct?

abbra:
Correct, you need to remove the old port range.
As for the number, I don't think it is needed. We are not planning to remove any of these ports from communication any time soon because these are the standard ports for various protocols

abbra:
To your last questions: both are correct, these can be dropped according to section 6 of RFC1002.

erig0:
> Correct, you need to remove the old port range.

Is removing the old range going to break users of older samba releases?

PR #475, opened 2019-03-09
Add Chromecast service

guoyunhe:
Chromecast requires several ports open to send video or audio to TV. It is a popular device so I think it worth a service configuration.

### DIFF
--- a/config/services/chromecast.xml
+++ b/config/services/chromecast.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>chromecast</short>
+  <description>Chromecast devices allow you to play videos or audios in your computer on TV without cables. You also need to enable mdns service.</description>
+  <port protocol="udp" port="1900"/>
+</service>


erig0:
Can you clarify why you think this service is needed? firewalld allows outgoing connections. Why would the client need to open these incoming ports?

guoyunhe:
> Can you clarify why you think this service is needed? firewalld allows outgoing connections. Why would the client need to open these incoming ports?

When projecting video/audio to TV, PC serves as a HTTP server and Chromecast is client. So here is incoming connections. Detailed explanation can be found here https://blog.g3rt.nl/allow-google-chromecast-host-firewall-iptables.html

erig0:
>  Detailed explanation can be found here https://blog.g3rt.nl/allow-google-chromecast-host-firewall-iptables.html

This post also mentions it's only needed if the OUTPUT policy is set to DROP. That is not the case for firewalld.
Have you tested using a chromecast from a machine running firewalld? AFAICS, it should work without this service.

guoyunhe:
I have tested Chromecast with default Firewalld configuration and it doesn't work. Once I do `systemctl stop firewalld`, it starts to work immediately.

erig0:
I still don't think all these ports are necessary, can you try the following:
```
# firewall-cmd --add-service=mdns
# firewall-cmd --add-port=1900/udp # SSDP
```

guoyunhe:
Sorry for late reply! I tried the commands you said and they worked.

erig0:
I just pushed a new SSDP service to master. For chromecast clients you should enable the following services:
```
# firewall-cmd --add-service=mdns
# firewall-cmd --add-service=ssdp
```

As such, this PR is no longer valid. I'm closing it.

PR #478, opened 2019-03-23
Fix Rich_Mark logic

kaechele:
We are looking to compare the type, not the object.
Without this fix ipXtables will only mark the very first packet of a connection.

Signed-off-by: Felix Kaechele <heffer@fedoraproject.org>


erig0:
Thanks @kaechele. I also fixed this is the nftables backend and added test coverage.

PR #481, opened 2019-04-03
Fix several issues with special characters in rich rules

### LINKED ISSUE #480: Issues with special characters in rich rules
There are several issues related to special characters in rich rules. I'll group them in this issue since they have similar solutions and I have a PR that I'll submit in a moment.

* When \n, \r or some other special character is used, this breaks the call to iptables as explained in https://bugzilla.redhat.com/show_bug.cgi?id=1596304 . This can be reproduced with:

```
> for i in $(curl "https://www.cloudflare.com/ips-v4" | unix2dos); do sudo firewall-cmd --permanent --zone=public --add-rich-rule 'rule family="ipv4" source address="'$i'" port port=80 protocol=tcp accept'; done

Error: COMMAND_FAILED: '/usr/sbin/iptables-restore -w -n' failed: iptables-restore v1.8.2 (legacy): option "-s" requires an argument
Error occurred at line: 2                                                                                                                                                                                                                  
Try `iptables-restore -h' or 'iptables-restore --help' for more information.                                                                                                                                                               
```

* When firewalld is using the iptables backend and the user enters some text into the log prefix widget in firewall-config, then the actual log prefix used includes single quotes around the prefix, but those single quotes shouldn't be there.

* When firewalld uses the nftables backend and the user enters some text with double quotes (one or more of them) into the log prefix widget in firewall-config, this breaks the netfilter parser.

antlarr-suse:
This PR fixes several issues with special characters in rich rules:

* Remove nonprintable characters when parsing a rich rule from a string

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1596304

* Remove single and double quotes from log prefixes.

That way we don't allow invalid nftables rules with broken syntax.

Also, since prefixes don't have quotes anymore, there's no need to
quote them in _rich_rule_log (since it will be quoted correctly anyway
later with shlex) so this fixes the double quoting issue with the
iptables backend that always added single quotes around log prefixes.

* Add removeCharacters function and CONTROL_CHARS set

Used by the fixes above.

Fixes #480

### DIFF
--- a/src/firewall/core/ipXtables.py
+++ b/src/firewall/core/ipXtables.py
@@ -901,7 +901,7 @@ class ip4tables(object):
         rule += self._rich_rule_priority_fragment(rich_rule)
         rule += rule_fragment + [ "-j", "LOG" ]
         if rich_rule.log.prefix:
-            rule += [ "--log-prefix", "'%s'" % rich_rule.log.prefix ]
+            rule += [ "--log-prefix", "%s" % rich_rule.log.prefix ]
         if rich_rule.log.level:
             rule += [ "--log-level", "%s" % rich_rule.log.level ]
         rule += self._rule_limit(rich_rule.log.limit)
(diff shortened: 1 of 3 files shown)


erig0:
Any luck fixing the testsuite failures? Does this behave correctly if the log prefix contains spaces?

erig0:
Superseded by #714.

PR #485, opened 2019-04-22
chore: config/services/RH-Satellite-6: use service includes

erig0:
explicit port number.


erig0:
@jcpunk, please review this.

jcpunk:
This looks right to me.

erig0:
> This looks right to me.

Thanks. Merged.

PR #486, opened 2019-04-22
chore: config/services/freeipa-4: use service includes

erig0:
explicit port number.


erig0:
@tiran, please review.

tiran:
@erig0 +1

I like the new approach more than the original approach.

abbra:
LGTM. This is much better -- even though we do not support relocating ports for freeipa instances to something else, reusing common names is better also from an educational point of view.

Issue #488, opened 2019-04-30
XSD for services XML


traek:
This isn't an issue per se but it would be nice to have the ability to ensure custom services conform to the documentation found at https://firewalld.org/documentation/man-pages/firewalld.service.

erig0:
This can be verified by using the `--check-config` option.

```
# firewall-cmd --help |grep check-config
  --check-config       Check permanent configuration for errors

# firewall-offline-cmd --help |grep check-config
      --check-config   Check system and default configuration
```

traek:
Sorry I wasn't more explicit in my request. The reason behind wanting an XSD is so I can author properly in the XML editor and have confidence it will pass said validation. I have a lot of these to do and often pull in XSDs when I'm doing large efforts like this. Since the XML spec already provides a mechanism for validating, I don't think this is an unreasonable request. In fact, I think it'd be an elegant solution if the check-config simply used an XSD (embedded or on disk) to do that check-config.

ciez:
.xsd schema for any .xml document is very useful indeed as it helps to write correct .xml document rather than go through an iterative try & guess process.

PR #492, opened 2019-06-04
Service XML Schema

traek:
Ensures XML author basic conformity to 'man firewalld.service'

### DIFF
--- a/config/services/services.xsd
+++ b/config/services/services.xsd
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+    <xs:simpleType name="transport-layer-protocol">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="tcp"/>
+            <xs:enumeration value="udp"/>
+            <xs:enumeration value="sctp"/>
+            <xs:enumeration value="dccp"/>
+        </xs:restriction>
+    </xs:simpleType>
+    <xs:element name="service">
+        <xs:complexType>
+            <xs:sequence>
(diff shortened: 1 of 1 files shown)


traek:
Addresses my request for #488

erig0:
@traek, see commit 81d3d6a7d84b and the handful of commits that precede it. A schema previously existed, but was removed in favor of the built in `--check-config` option. What is your use case for XSD that can't be served by `firewall-offline-cmd --check-config` ?

charlie1968:
If you have a bunch of XML config files to edit, I can import the XSD and support in my favorite IDE for writing the xml files would be great.

ciez:
thank you very much for the effort, @traek !

Much appreciated

Tristan971:
Seconding charlie in this case. The builtin CLI checker is great if you need a last syntax+semantics check at the time of deployment of the configuration, but it doesn't entirely overlap with XML schemas.

With those, one's IDE can handle silly typos and syntactically invalid things very early on, and even provide autocomplete suggestions that helps find options one might not have missed while skimming through the documentation, etc.

Overall I don't quite see why it would be one or the other rather than both?

thom311:
I agree that schemas are useful and should be (re) added.

Since schemas exist in the past, I think we should instead revert their removal, and not add a (different) schema from this pull request. Also, this pull-request only covers services, while the revert covers also `check`, `icmptype`, `ipset`, and `zone`, and adds utility script. Also, the `services.xsd` schema from this pull request does not validate the currently existing schemas (which would be fixable).

Instead, I think it's better to close this PR and instead do a revert (pull request 1198).

I am sorry for the long time it took, and then preferring another approach in the end. Thank you for your efforts though!

Issue #499, opened 2019-07-22
'Active bindings' part of UI is not working


yrro:
Under "active bindings" I see an indented entry for my currently active
network conection. On the row beneath is a duplicate of this entry, but
clicking on it doesn't do anything. And if I move the mouse into this
entry from beneath, it changes to say "Sources" instead.

At the same time, `firewall-config` outputs the following:

    (firwall-config:12867): Gtk-WARNING **: 14:36:45.228: Failed to set
    text from markup due to error parsing markup: Unknown tag
    'interface' on line 1 char 20

In the Options → Change Zones of Connections... menu, there's a menu
entry called `<interface>` in the same position as the faulty item in
"active bindings".

I guess `<interface>` is a string replacement that isn't happening, and
Gtk is trying to interpret that as Pango markup.

I've seen this with 0.6.3 and 0.7.0 on Debian and 0.6.4 on Fedora.

thom311:
> (firwall-config:12867): Gtk-WARNING **: 14:36:45.228: Failed to set
text from markup due to error parsing markup: Unknown tag
'interface' on line 1 char 20

this is an actual bug, not a translation problem. Whatever text there is, it must not be interpreted as markup.

The PR #500 addresses the symptom, not the cause.

yrro:
I agree, all text added to the bindings store must be escaped, since the cell renderer is told to interpret column 0 as markup, not as text.

PR #500, opened 2019-07-22
Fix bad translation for "Interfaces" in several translations

yrro:
Fixes: #499

### DIFF
--- a/po/as.po
+++ b/po/as.po
@@ -147,7 +147,7 @@ msgstr "সংযোগসমূহ"
 #: ../src/firewall-config.in:2642 ../src/firewall-config.in:2862
 #: ../src/firewall-config.in:2889 ../src/firewall-config.glade.h:130
 msgid "Interfaces"
-msgstr "<interface>"
+msgstr ""
 
 #: ../src/firewall-applet.in:558 ../src/firewall-config.in:647
 #: ../src/firewall-config.in:2682 ../src/firewall-config.in:2865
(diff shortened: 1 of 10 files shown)


thom311:
Sorry for no reply for such a long time...

There are changes like
```diff
 msgid "Interfaces"
-msgstr "<interface>"
+msgstr ""
``` 
which seems not a correct translation update.

I cannot fix this, because I am not (sufficiently) familiar with the languages.


I think this pull request should be closed (or reworked).

yrro:
> There are changes like
> 
> ```diff
>  msgid "Interfaces"
> -msgstr "<interface>"
> +msgstr ""
> ```
> 
> which seems not a correct translation update.
> 
> I cannot fix this, because I am not (sufficiently) familiar with the languages.
> 
> I think this pull request should be closed (or reworked).

I think this is how you indicate that there is no translation, i.e., fall back to `msgid`. This lets translators know that the string needs to be translated.

thom311:
Hi. You are correct. Patch makes sense to me!

The branch should be rebased to latest `master` (I don't have permission to do that). Rebasing will result in conflicts, due to the changes to the "PO-Revision-Date:".

yrro:
Done, thanks

yrro:
Although GitHub shows that latest push as making loads of changes, it's just a rebase onto the latest `main` branch that doesn't change anything.

erig0 (approved):

PR #513, opened 2019-09-14
config/services/matrix: use service include

ykgmfq:
explicit port number.


erig0:
Thanks for using the new service include feature. :)

Issue #517, opened 2019-09-21
kernel builltins: ERROR: INVALID_HELPER: 'nf_conntrack_netbios_ns' is not available


djlucas:
When using monolithic kernel (no modules), unable to load the helper, and it enters a failed state. Obviously killing off samba in internal gets past the error.

> dj@lfsdt2 [ ~ ]$ grep "nf_conntrack" /lib/modules/$(uname -r)/modules.builtin
kernel/net/netfilter/nf_conntrack.ko
kernel/net/netfilter/nf_conntrack_netlink.ko
kernel/net/netfilter/nf_conntrack_amanda.ko
kernel/net/netfilter/nf_conntrack_ftp.ko
kernel/net/netfilter/nf_conntrack_h323.ko
kernel/net/netfilter/nf_conntrack_irc.ko
kernel/net/netfilter/nf_conntrack_broadcast.ko
kernel/net/netfilter/nf_conntrack_netbios_ns.ko
kernel/net/netfilter/nf_conntrack_snmp.ko
kernel/net/netfilter/nf_conntrack_pptp.ko
kernel/net/netfilter/nf_conntrack_sane.ko
kernel/net/netfilter/nf_conntrack_sip.ko
kernel/net/netfilter/nf_conntrack_tftp.ko

I _think_ this is ultimately in the modinfo calls, but I couldn't follow the code well enough to even verify that. Assuming I'm in the correct function, (get_modinfos()), I attempted to separate out the builtins inline with a separate list  - that part was easy enough, I didn't bother attaching a patch, I just stripped out (/lib/modules/%s, kver) and used re.match on modules.builtin in that same prefix,, and then put it back for the second half of the loop (the original code that populates modules[]), but I've no idea how to handle the binary file. The best I could come up with at CLI was the following (yes it's ugly):

>dj@lfsdt2 [ ~ ]$ grep -z "nf_conntrack_netbios_ns" /lib/modules/5.2.16/modules.builtin.modinfo | sed 's@nf_conntrack_netbios_ns@\nnf_conntrack_netbios_ns@g'
nf_conntrack_netbios_ns.parm=timeout:timeout for master connection/replies in seconds
nf_conntrack_netbios_ns.parmtype=timeout:uint
nf_conntrack_netbios_ns.alias=nfct-helper-netbios_ns
nf_conntrack_netbios_ns.alias=ip_conntrack_netbios_ns
nf_conntrack_netbios_ns.license=GPL
[...]

PR #518, opened 2019-09-25
fix: allow use of kernel builtins for nf_conntrack helpers

djlucas:
Tested only on SysV with all modules built in. Works for me. Somebody more competent in Python can probably clean that up a bit. I'm concerned about the duplicate lines, but wasn't sure if 'ret' could be reused in that context, and this _should_ still work for mixed modules and built-ins (though I did not test that). This should tie in with #517 (opened also by me).

HTH

### DIFF
--- a/src/firewall/functions.py
+++ b/src/firewall/functions.py
@@ -351,12 +351,27 @@ def get_nf_conntrack_short_name(module):
 
 def get_modinfos(path_templates, prefix):
     kver = os.uname()[2]
+    builtinmods = []
     modules = []
-    for path in (t % kver for t in path_templates):
-        if os.path.isdir(path):
-            for filename in sorted(os.listdir(path)):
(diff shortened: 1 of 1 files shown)


djlucas:
Oops, forget to remove the extraneous print statements used while testing.

djlucas:
Checking status, for whatever reason, I guess I borked the squash. The commit above (without the green checkmark) is the squashed commit (15bfa3b), and effectively what was tested. If needs be, I'll close, destroy that fork, and resubmit a proper commit, just say so.

Thanks.

djlucas:
Tested with mixed builtins and modules. Checks OK

erig0:
This PR was closed. Instead of the fixes here modinfo usage was completely removed in commit adbf3476b7533ef3a2c002db62f76614a9f0f6c1.

PR #522, opened 2019-10-06
feat(service): Add CrateDB ports

hos7ein:
Open tcp ports 4200, 4300 and 5432 for CrateDB service.

### DIFF
--- a/config/services/cratedb.xml
+++ b/config/services/cratedb.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>CrateDB</short>
+  <description>CrateDB is a distributed SQL database management system that integrates a fully searchable document oriented data store.</description>
+  <port protocol="tcp" port="4200"/>
+  <port protocol="tcp" port="4300"/>
+  <include service="postgresql"/>
+</service>
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -139,6 +139,7 @@ CONFIG_FILES = \
 	services/cockpit.xml \
 	services/collectd.xml \
 	services/condor-collector.xml \
+	services/cratedb.xml \
 	services/ctdb.xml \
 	services/dhcpv6-client.xml \
 	services/dhcpv6.xml \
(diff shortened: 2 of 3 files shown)


config/services/cratedb.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>ceph-mon</short>
+  <description>CrateDB is a distributed SQL database management system that integrates a fully searchable document oriented data store.</description>
+  <port protocol="tcp" port="4200"/>
+  <port protocol="tcp" port="4300"/>
+  <port protocol="tcp" port="5432"/>
```
TorontoMedia, line 7: I believe port `5432` should be moved to a separate service file such as `cratedb-postgresql.xml` or even better, mention in the description to also use the `postgresql` service if PostgreSQL wire support is enabled.

TorontoMedia:
Could you rebase and amend/squash your commits into one so that we have a cleaner git history? Maybe add the type reference to the commit message, so like this `feat(service): Add CrateDB ports`

TorontoMedia (changes requested):

TorontoMedia:
@hos7ein Is the postgresql wire protocol option enabled by default? Also could you please rebase/squash these commits into a single commit using the message `feat(service): Add CrateDB service` and force push these changes so that we have a cleaner git history?

config/services/cratedb.xml
erig0: If CrateDB requires this port by default, then we should include it here. Since it's the same port, we could just include the postgresql service, e.g. `<include service="postgresql" />`.

erig0 (changes requested):
Otherwise, lgtm.

config/services/cratedb.xml
TorontoMedia: > If CrateDB requires this port by default, then we should include it here. Since it's the same port, we could just include the postgresql service, e.g. `<include service="postgresql" />`.

It's undocumented but the postgresql wire protocol is enabled by default, so the `postgresql` service file should be include as you mentioned.

erig0 (approved):
I squashed the commits and added an `include` for `postgresql`.

I will merge after CI.

PR #524, opened 2019-10-08
improvement: added shell-completion on bash to firewall-offline-cmd

snifbr:
This commit add bash_completion to firewall-offline-cmd command.

Signed-off-by: Danilo Figueiredo Rocha <drocha.figueiredo@gmail.com>

### DIFF
--- a/shell-completion/bash/firewall-offline-cmd
+++ b/shell-completion/bash/firewall-offline-cmd
@@ -1,0 +1,228 @@
+# bash completion for firewall-offline-cmd                         -*- shell-script -*-
+
+# Authors:
+# Danilo Figueiredo Rocha <drocha.figueiredo@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# TODO: find a way how to get the following options from firewall-offline-cmd
+
+## Revisado todas as opções --DFR-- 30-07-2018
+OPTIONS_LOCKDOWN="--add-lockdown-whitelist-command= --remove-lockdown-whitelist-command= \
+       --query-lockdown-whitelist-command= --list-lockdown-whitelist-commands \
+       --add-lockdown-whitelist-context= --remove-lockdown-whitelist-context= \
+       --query-lockdown-whitelist-context= --list-lockdown-whitelist-contexts \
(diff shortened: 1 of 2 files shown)


shell-completion/bash/firewall-offline-cmd
```diff
@@ -0,0 +1,228 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# TODO: find a way how to get the following options from firewall-offline-cmd
```
erig0: Maybe you can parse `--help` output?

shell-completion/bash/firewall-offline-cmd
```diff
@@ -0,0 +1,228 @@
...
+
+## Revisado todas as opções --DFR-- 30-07-2018
```
erig0: Please use English comments. These look stale anyway. Maybe just remove them.

snifbr: Sorry, I just forgot to remove, this comment is useless.

Can you remove this? or what I have to do to fix my PR?

shell-completion/bash/firewall-offline-cmd
snifbr: Sorry, it's just old comments I forgot remove from firewall-cmd completion file.

Can you remove this? or what I have to do to fix my PR?

snifbr:
> In general, the options available to firewall-offline-cmd are mostly the same to firewall-cmd. It'd be nice if the completion scripts could be common.

Would you like that I'll put this scripts in the same completion file of firewall-cmd? Does it make sense for you?

shell-completion/bash/firewall-offline-cmd
erig0: You can remove these comments. You'll have to push other changes anyway.

erig0:
> > In general, the options available to firewall-offline-cmd are mostly the same to firewall-cmd. It'd be nice if the completion scripts could be common.
> 
> Would you like that I'll put this scripts in the same completion file of firewall-cmd? Does it make sense for you?

That would work. Any way you can share between `firewall-cmd` and `firewall-offline-cmd` is good.

shell-completion/bash/firewall-offline-cmd
```diff
@@ -0,0 +1,228 @@
...
+                 --direct"
+
+_firewall_offline_cmd()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+    firewall-cmd --state &> /dev/null
+    if [[ "${?}" -eq 0 ]]; then
+        echo -e "\nfirewalld is running, please use firewall-cmd."
+        return
```
thom311, line 133: this doesn't seem right. `firewall-offline-cmd` may be for daemon-less use, but it should not fail because the daemon is running.

thom311 (approved):
Optimally, I think implementing bash completion via a shell-script, always leads to suboptimal results. A command should implement it's own completion and not be reimplemented in shell.

Anyway, since currently no completion exists, I think this patch is an improvement and should be merged. It can be improved later.

PR #542, opened 2019-11-20
Fix - Move /var/log/firewalld creation into debug arg handling

tcunningham07:
Documentation specifies that /var/log/firewalld contains debug output
written when the --debug option is included. Currently, this file will
be opened and created regardless of whether the debug option has been
included. This also removes some dead code.


erig0:
I think this is obsoleted by #964.

PR #559, opened 2019-12-27
RHEL7 doesn't have nftables

jcpunk:
This adds some additional logic to the sample spec file as RHEL7 doesn't have nftables.

### DIFF
--- a/firewalld.spec
+++ b/firewalld.spec
@@ -51,7 +51,9 @@ Requires: dbus-python
 Requires: python-slip-dbus
 Requires: python-decorator
 Requires: pygobject3-base
+%if 0%{?rhel} >= 8
 Requires: python-nftables
+%endif
 
 %description -n python-firewall
 Python2 bindings for firewalld.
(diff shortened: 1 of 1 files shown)


firewalld.spec
```diff
@@ -119,15 +121,24 @@ firewalld.
 
 %prep
 %autosetup
+%if 0%{?rhel} < 8
```
erig0: This change is not necessary as of 3f7419dcb50a5bdcf800512d24919c33a413c71b. We only rely on python3-nftables and no longer check for the `nft` binary.

firewalld.spec
```diff
@@ -119,15 +121,24 @@ firewalld.
...
+./autogen.sh --without-nft
```
erig0: This change is not necessary as of 3f7419d. We only rely on python3-nftables and no longer check for the nft binary.

erig0:
@jcpunk , not sure if you meant to close this PR. The guards around python3-nftables were fine, but the changes to `./configure` invocations are not necessary.

I opened issue #563 to track this.

Issue #2, opened 2015-03-25
Accept/Reject/Drop Forwarding packets


jonramc:
It would be a great deal if there was the option of creating a rule for forwarding packets within any zone.

For example: firewall-cmd --zone=internal --add-rich-rule='rule family="ipv4" source address="192.168.100.0/24" service name="http" type="forward" drop'

This way you can control any forwarding traffic coming from your LAN to the Internet or any other Interface.

t-woerner:
Yes, this needs to be implemented. A limitation of rules to INPUT or FORWARD would be good to have.

mailinglists35:
https://bugzilla.redhat.com/show_bug.cgi?id=929426

PR #561, opened 2020-01-02
improvement: Select target chain in rich rules

adamrushad:
Adds needed syntax, implementation in ipXtables and nftables,
documentation, and tests to extend the rich rule syntax to allow
rules to be added to chains other than INPUT. Currently limited
to INPUT, FORWARD_IN, and FORWARD_OUT. OUTPUT could be added easily as
soon as firewalld creates the needed per-zone OUTPUT chains.

Currently, firewalld lacks the ability to be effectively used on a
router due to rules only getting added to the INPUT chain. This commit
seeks to resolve this.

Fixes: #2

### DIFF
--- a/doc/xml/firewalld.richlanguage.xml
+++ b/doc/xml/firewalld.richlanguage.xml
@@ -99,7 +99,7 @@ rule
       <title>Rule</title>
       <para>
 	  <programlisting>
-rule [family="ipv4|ipv6"] [priority="priority"]
+rule [family="ipv4|ipv6"] [priority="priority"] [chain="INPUT|FORWARD_IN|FOWARD_OUT"]
 	  </programlisting>
       </para>
       <para>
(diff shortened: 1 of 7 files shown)


erig0:
This is an interesting change, but we're currently in the process of adding full support for forward/output filtering in a way that will let us use all features (services, helpers, ..) not just rich rules. See https://github.com/orgs/firewalld/projects/1#card-25963208.

As such, I don't see us merging this PR.

adamrushad:
Not a problem. I had seen the card, but hadn't seen any specific work around the forward/output filtering, so decided to take a stab at it. I had chosen to make the patch rich-rule only because adding generic support involves a lot more design work to adjust output for displaying non-rich forward/output rules, and doing that kind of UX work is not something I am generally skilled with.

If a more generic patch like that is desired, I could probably work on it over the weekend. I would likely refactor what I have here to make it a bit more generic, and I don't know how the interface should be modified to display the non-rich rules. That aside, adding zones as policy objects in the future should "just work," since I shouldn't need to muck around with that particular code. I would want to know if you would prefer commits to be squashed or separated for adding rich vs generic rules. Adding support for generic rules via firewall-cmd and services shouldn't be much of an issue, though.

erig0:
> Not a problem. I had seen the card, but hadn't seen any specific work around the forward/output filtering, so decided to take a stab at it. I had chosen to make the patch rich-rule only because adding generic support involves a lot more design work to adjust output for displaying non-rich forward/output rules, and doing that kind of UX work is not something I am generally skilled with.

That makes sense. When I first started thinking about adding forward/output filtering doing it via rich-rules was my first approach as well. But after more thought we came up with policy objects as they'll help us address some other short-comings with the current zone implementation.

> If a more generic patch like that is desired, I could probably work on it over the weekend. I would likely refactor what I have here to make it a bit more generic, and I don't know how the interface should be modified to display the non-rich rules.

It's a little more complex than that due to the dbus interface. e.g. "addService()" would need a new argument to specify the target chain. But the dbus API is stable and we can't change the method signature. We'd have to introduce an entirely new method.. for each feature (service, ports, etc).

> That aside, adding zones as policy objects in the future should "just work," since I shouldn't need to muck around with that particular code.

Agreed.

> I would want to know if you would prefer commits to be squashed or separated for adding rich vs generic rules. Adding support for generic rules via firewall-cmd and services shouldn't be much of an issue, though.

I prefer smaller, more digestible commits. It makes them easier to review.

erig0:
Closing as this is obsoleted by [policy objects](https://firewalld.org/2020/09/policy-objects-introduction).

PR #596, opened 2020-03-29
Add an option to disable building documentation

aduskett:
For embedded applications such as Buildroot or Yocto, the man pages may not be
desired or even capable of being built.

On line 45 of configure.ac there is the line:
JH_CHECK_XML_CATALOG([http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl], [DocBook XSL Stylesheets])

There are three issues with this:
  - It requires building the xml-catalog package.
  - It automatically defaults to the host systems xml-catalog
    instead of the cross environments.
  - It isn't necessary to have a functioning firewalld.

Create a new option: --disable-docs. By default, build the documentation, but
if a user passes --disable-docs documentation will no longer be built as
desired.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,7 +42,14 @@ GLIB_GSETTINGS
 
 #############################################################
 
-JH_CHECK_XML_CATALOG([http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl], [DocBook XSL Stylesheets])
+AC_ARG_ENABLE([docs],
+    [AS_HELP_STRING([--enable-docs], [Enable building documentation])],
+    [ENABLE_DOCS="${enableval}"], [ENABLE_DOCS='yes'])
+AM_CONDITIONAL(ENABLE_DOCS, [test x$ENABLE_DOCS = xyes])
+if test "x$ENABLE_DOCS" = "xyes"; then
+    JH_CHECK_XML_CATALOG([http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl], [DocBook XSL Stylesheets])
+fi
+AC_SUBST(ENABLE_DOCS)
 
 #############################################################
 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -184,6 +184,15 @@ AC_CONFIG_FILES([Makefile
 		 src/tests/atlocal
 		 src/icons/Makefile])
 
+if test "x$ENABLE_DOCS" = "xyes"; then
+AC_CONFIG_COMMANDS([xsl-cleanup],,[rm -f doc/xml/transform-*.xsl])
+AC_CONFIG_FILES([doc/Makefile
```
yann-morin-1998: As per the [autoconf manual](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/autoconf.html#Configuration-Actions), it is indeed legit to call `AC_CONFIG_FILES()` multiple times:

```
This section describes [...] AC_CONFIG_FILES [...]
All these macros can be called multiple times, [...]
```

configure.ac
```diff
@@ -42,7 +42,14 @@ GLIB_GSETTINGS
...
 #############################################################
 
-JH_CHECK_XML_CATALOG([http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl], [DocBook XSL Stylesheets])
+AC_ARG_ENABLE([docs],
+    [AS_HELP_STRING([--enable-docs], [Enable building documentation])],
+    [ENABLE_DOCS="${enableval}"], [ENABLE_DOCS='yes'])
+AM_CONDITIONAL(ENABLE_DOCS, [test x$ENABLE_DOCS = xyes])
+if test "x$ENABLE_DOCS" = "xyes"; then
+    JH_CHECK_XML_CATALOG([http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl], [DocBook XSL Stylesheets])
+fi
```
yann-morin-1998: If we were to try and be pedantic, this should be written as:

```
AS_IF([test "x$ENABLE_DOCS" = "xyes"],
      [JH_CHECK_XML_CATALOG([http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl],
                            [DocBook XSL Stylesheets])])
```

erig0: Or use AM_COND_IF([ENABLE_DOCS], [... ])

erig0:
I'm working on cleaning this up manually. I'll should push it later today.

erig0:
I just pushed the changes. I had to fix it up because a few things were broken. Namely "make distclean". Thanks for the work!

configure.ac
yann-morin-1998: > AM_COND_IF()

Oh, I did not know about that one, thanks! :-)

aduskett:
Thank you so much! I really appreciate your help!

PR #616, opened 2020-04-25
Added service definitions for AFP and WS-Discovery.

mymatenige:
In a mixed Linux/Windows/Mac OS X environment traffic from AFP and WS-Discovery is often seen. It would be good to add service definitions for these to firewalld.

### DIFF
--- a/config/services/afp.xml
+++ b/config/services/afp.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Apple Filing Protocol</short>
+  <description>The Apple Filing Protocol (AFP), formerly AppleTalk Filing Protocol, is a proprietary network protocol, and part of the Apple File Service (AFS), that offers file services for macOS and the classic Mac OS.</description>
+  <port protocol="tcp" port="548"/>
+</service>
--- a/config/services/ws-discovery.xml
+++ b/config/services/ws-discovery.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Web Services Dynamic Discovery</short>
+  <description>Web Services Dynamic Discovery (WS-Discovery) is a technical specification that defines a multicast discovery protocol to locate services on a local network.</description>
+  <port protocol="tcp" port="3702"/>
+  <port protocol="udp" port="3702"/>
+</service>


marcosfrm:
WS-Discovery also needs

```xml
<destination ipv4="239.255.255.250" ipv6="ff02::c"/>
```

marcosfrm:
@mymatenige can you update the PR, also prefixing each commit with "feat(service): add ..."?

mymatenige:
If I do, will it be considered and progressed?  Please don't read that as me being difficult, I'm asking a serious question as it's been sat here for over a year, I don't want to spend effort on something and have it sat for another 15 months ...

erig0:
> If I do, will it be considered and progressed?

Yes.

> Please don't read that as me being difficult, I'm asking a serious question as it's been sat here for over a year, I don't want to spend effort on something and have it sat for another 15 months ...

Understood. I'm overloaded and things fall through the cracks. Polite pings are helpful.

erig0:
> WS-Discovery also needs
> 
> ```
> <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
> ```
> 
> Edit:
> http://docs.oasis-open.org/ws-dd/discovery/1.1/os/wsdd-discovery-1.1-spec-os.html

Yes. Please add this as well.

mymatenige:
Thanks.  I'll revisit within a week and update it as appropriate.

marcosfrm:
> WS-Discovery also needs
> 
> ```
> <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
> ```
> 
> Edit:
> http://docs.oasis-open.org/ws-dd/discovery/1.1/os/wsdd-discovery-1.1-spec-os.html

ooops. The destination IP needs to apply only to the UDP port, otherwise it will not work. Is that possible?

`<port protocol="udp" port="3702" destination ipv4="239.255.255.250" ipv6="ff02::c"/>` or so...

erig0:
> > WS-Discovery also needs
> > ```
> > <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Edit:
> > http://docs.oasis-open.org/ws-dd/discovery/1.1/os/wsdd-discovery-1.1-spec-os.html
> 
> ooops. The destination IP needs to apply only to the UDP port, otherwise it will not work. Is that possible?
> 
> `<port protocol="udp" port="3702" destination ipv4="239.255.255.250" ipv6="ff02::c"/>` or so...
> 
> Edit:
> Same situation with LLMNR since [497c1e8](https://github.com/firewalld/firewalld/commit/497c1e88d3e271673917e8c919c5fe2dc137c9ac). I did not notice it because Windows implementation is UDP-only.

No. This is correct. The elements can be separate. It will match UDP port + destination in the same iptables/nftables rule.

marcosfrm:
Hence we cannot set the source address for just one port/protocol in a single service with multiple ports/protocols, right?

Perhaps something like this could be implemented:

```xml
  <!-- without source address -->
  <port protocol="tcp" port="3702"/>

  <!-- with source address -->
  <port protocol="udp" port="3702">
    <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
  </port>
```

erig0:
> Hence we cannot set the destination address for just one port/protocol in a single service with multiple ports/protocols, right?
> 
> Perhaps something like this could be implemented:
> 
> ```
>   
>   <port protocol="tcp" port="3702"/>
> 
>   
>   <port protocol="udp" port="3702">
>     <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
>   </port>
> ```

I get what you're saying now. Another option is to add `ws-discovery-tcp` and `ws-discovery-udp`. Then make `ws-discovery` include both. I think prefer this approach to expanding the service definition syntax.

marcosfrm:
IMHO having three services for each one bring confusion. But at least just WS-Discovery and LLMNR need this as far as I can see.

erig0:
> IMHO having three services for each one bring confusion. But at least just WS-Discovery and LLMNR need this as far as I can see.

Right, but I think it's fair to split some services into TCP/UDP. Some services may operate on both TCP/UDP, but networks often only use one, e.g. UDP. I would guess LLMNR falls into this category. AFAICS, for LLMNR TCP only makes sense if you have a "DNS server", but then that's functionally the same as a normal DNS server.

marcosfrm:
Yes, LLMNR in Windows, the main implementation, only use UDP. WS-Discovery on the other hand needs the TCP port because it is where it serves SOAP content via HTTP.

Would expanding the service  syntax too complex? Who knows what crazy protocol may appear in the future...

marcosfrm:
@mymatenige LLMNR is now split into TCP and UDP, take a look https://github.com/firewalld/firewalld/pull/833

erig0:
Closing as duplicate of successor PR #838.

PR #619, opened 2020-04-29
fix(direct): rule in a zone chain

erig0:


erig0:
On Wed, Apr 29, 2020 at 07:15:16AM -0700, Tomáš Doležal wrote:
> test steps looks nearly complete, I'd like to request just one more
> thing: load the rules from saved configuration. This could be done via
> --runtime-to-permanent follwed by --reload/--complete-reload.

Normally I would add --permanent coverage, but it's not relevant for
this bug. The coverage here is for generating the runtime iptables
chains based off the requested direct rule chain name, e.g.
PRE_public_allow.

todoleza:
If you wish to keep this test strictly related to runtime and the bug report, I am for it.
Additionally I have to ask whether ipv6 coverage was omitted on purpose or should be added as well.

Please consider that the coverage could be improved with the permanent check to catch other bugs related to working with direct rules in future when the test is reusable with a subtle addition like this not so ideal example:
```
+FWD_CHECK([-q --runtime-to-permanent])
+FWD_RESTART
+NS_CHECK([iptables-save -t raw | grep -c '10\.10\.10\.0/24' | grep 6], 0, [ignore], [ignore])
+NS_CHECK([iptables-save -t mangle | grep -c '10\.10\.10\.0/24' | grep 6], 0, [ignore], [ignore])
+NS_CHECK([iptables-save -t nat | grep -c '10\.10\.10\.0/24' | grep 12], 0, [ignore], [ignore])
+NS_CHECK([iptables-save -t filter | grep -c '10\.10\.10\.0/24' | grep 18], 0, [ignore], [ignore])
```

I found no other upstream test case covering direct rules in permanent config, only invalid direct rules and very specific cases in [src/tests/cli/firewall-cmd.at#L956-L1088](https://github.com/firewalld/firewalld/blob/c0ad3a0b3340a27c34b33128f756f64acc3a771b/src/tests/cli/firewall-cmd.at#L956-L1088).

erig0:
On Wed, Apr 29, 2020 at 09:42:56AM -0700, Tomáš Doležal wrote:
> If you wish to keep this test strictly related to runtime and the bug report, I am for it.
> Additionally I have to ask whether ipv6 coverage was omitted on purpose or should be added as well.

Not intentional, but the code under test is common to both.

> Please consider that the coverage could be improved with the permanent
> check to catch other bugs related to working with direct rules

This may be a good opportunity to go through the process of submitting
test improvements/enhancements.

> +NS_CHECK([iptables-save -t raw | grep -c '10\.10\.10\.0/24' | grep 6], 0, [ignore], [ignore])

There are existing macros to list rule, e.g. `IPTABLES_LIST_RULES`.
These handle things like namespace, output normalization, locking, etc.

> I found no other upstream test case covering direct rules in permanent
> config

There is a bit more coverage in regression tests. But I agree. There
could be more.

```
$ ag -l "FWD_CHECK.*--permanent.*--direct" src/tests/    
src/tests/cli/firewall-cmd.at
src/tests/regression/rhbz1723610.at
src/tests/regression/rhbz1498923.at
```

PR #647, opened 2020-06-10
Add collectd service

vrindle:
I added the missing service collectd.


vrindle:
Updated commit message. Should be good to go.

PR #649, opened 2020-06-10
Update syslog.xml for tcp

sghosh151:
Added non-TLS TCP support for port 514


marcosfrm:
Straightforward change. Please change commit title to "improvement(service): ...".

erig0:
syslog on `514/tcp` is not defined by IANA. Nor does it appear to be in wide spread use. [RFC 6587[(https://datatracker.ietf.org/doc/html/rfc6587) says 514 is commonly used, but it's actually allocated for something else.

Is using `514/tcp` common enough to be the defacto standard? i.e. do user _expect_ `514/tcp` to be allowed when enabling the the `syslog` service?

PR #652, opened 2020-06-15
doc(README): add libxslt for doc generation

vrindle:
Updated the commit message and should be good to go


erig0:
@vrindle, thanks!

PR #653, opened 2020-06-22
fix(dbus): handle all zone settings in setZoneSettings2 method

danc86:
The new setZoneSettings2 D-Bus method was not handling a few zone
attributes which are plain attributes and do not have add/remove
methods: version, short, description, and target.

The problem was most evident with the client setZoneSettings method,
which was feeding back the complete zone settings, which was failing
like this:

      File "src/firewall/server/decorators.py", line 67, in dbus_handle_exceptions
        return func(*args, **kwargs)
      File "src/firewall/server/firewalld.py", line 917, in setZoneSettings2
        self.fw.zone.set_config_with_settings_dict(zone, settings, sender)
      File "src/firewall/core/fw_zone.py", line 398, in set_config_with_settings_dict
        raise FirewallError(errors.INVALID_SETTING, "Unhandled setting type {} key {}".format(type(settings[key]), key))
    firewall.errors.FirewallError: INVALID_SETTING: Unhandled setting type <class 'dbus.String'> key description

This patch makes setZoneSettings2 handle the attributes that it was
previously ignoring, and adds test coverage for setZoneSettings in the
client API as well as a direct test case for the setZoneSettings2 D-Bus
method.

Fixes: 67d057508f81 ("improvement(dbus): new dict based APIs for zones")

### DIFF
--- a/src/firewall/client.py
+++ b/src/firewall/client.py
@@ -114,6 +114,8 @@ class FirewallClientZoneSettings(object):
     def getSettingsDict(self):
         settings = {}
         for key,value in zip(self.settings_name, self.settings):
+            if key == 'UNUSED':
+                continue
             settings[key] = value
         return settings
     @handle_exceptions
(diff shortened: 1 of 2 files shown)


src/firewall/core/fw_zone.py
```diff
@@ -363,6 +363,13 @@ def add_rule_wrapper(zone, rule_str, timeout=0, sender=None):
         def remove_rule_wrapper(zone, rule_str):
             self.remove_rule(zone, Rich_Rule(rule_str=rule_str))
 
+        # These settings have no add/remove functions, they are plain attributes
```
erig0, line 366: These are only allowed to be changed in the permanent configuration. As such we shouldn't handle them here. Instead in the client class we should remove them in `setZoneSettings()` after getting the settings with `getSettingsDbusDict()`.

This will also simplify some of your other changes. i.e. the changes is ` src/tests/dbus/zone_runtime_functional.at` aren't relevant anymore.

https://github.com/firewalld/firewalld/blob/af3f7cd074f737c584a42cf1028f18e6fa597204/doc/xml/firewalld.dbus.xml#L1320-L1333

danc86: Ahh that makes sense. Will update the PR.

danc86: Do you think it's okay to just have a hardcoded list of attributes to be discarded inside the `setZoneSettings` client method?

Does `UNUSED` belong in that list, or should it still be filtered out in `getSettingsDbusDict`?

erig0: > Do you think it's okay to just have a hardcoded list of attributes to be discarded inside the `setZoneSettings` client method?

Maybe introduce getRuntimeSettingsDict() and getRuntimeSettingsDbusDict(). It has the benefit that the knowledge is localized in the FirewallClientZoneSettings class.

> Does `UNUSED` belong in that list, or should it still be filtered out in `getSettingsDbusDict`?

It should be filtered out as you've already done.

danc86:
Amended to add the new client methods as discussed.

erig0:
Looks great @danc86! Thanks!

PR #657, opened 2020-06-24
freeipa-trust: update dynamic DCE RPC ports

abbra:
Samba did change DCE RPC dynamic port range to 49152-65535 with version 4.7.

Signed-off-by: Alexander Bokovoy <abokovoy@redhat.com>

### DIFF
--- a/config/services/freeipa-trust.xml
+++ b/config/services/freeipa-trust.xml
@@ -9,6 +9,6 @@
   <port protocol="udp" port="389"/>
   <port protocol="tcp" port="445"/>
   <port protocol="udp" port="445"/>
-  <port protocol="tcp" port="1024-1300"/>
+  <port protocol="tcp" port="49152-65535"/><!-- Dynamic RPC Ports -->
   <port protocol="tcp" port="3268"/>
 </service>


abbra:
This issue needs to be fixed in RHEL and Fedora (and other distros too)

abbra:
Fedora bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1850419
RHEL 8 bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1850418

config/services/freeipa-trust.xml
```diff
@@ -9,6 +9,6 @@
   <port protocol="udp" port="389"/>
   <port protocol="tcp" port="445"/>
   <port protocol="udp" port="445"/>
-  <port protocol="tcp" port="1024-1300"/>
+  <port protocol="tcp" port="49152-65535"/><!-- Dynamic RPC Ports -->
```
erig0, line 12: Would it be better to add a `samba-rpc` service definition and then include, see 1fc208bf93171920196ed771b6a4c709ce6607aa, that from there and `samba-dc` ? `freeipa-trust` has a lot of overlap with `samba-dc`.

abbra: FreeIPA and Samba AD are different enough, it is better to keep the port definitions for them separated.

For this single item I do not see a need to create a separate definition -- unless you have a way to define a virtual service that could be overridden by an administrator and default to something otherwise. These port ranges depend on what is defined on AD side -- if AD administrators set a different range for their DCs, Linux administrators would need to change this and Samba configuration in FreeIPA.

erig0: ACK. Good enough for me.

abbra:
Thank you!

PR #658, opened 2020-06-25
core: rich: Catch ValueError on non-numeric priority values

n0-1:
Be a bit more user-friendly by printing:

| Error: INVALID_RULE: invalid 'priority' attribute value 'ab'.

instead of Python's default, which is:

| Error: invalid literal for int() with base 10: 'ab'


src/tests/regression/rhbz1689429.at
```diff
@@ -0,0 +1,9 @@
+FWD_START_TEST([rich rule invalid priority])
+AT_KEYWORDS(rich rhbz1689429)
+
+FWD_CHECK([--add-rich-rule='rule priority=foo accept'], 122, [],
```
erig0, line 4: Can you add the `--permanent` flag here? That'll magically get us `firewall-offline-cmd` coverage as well. Then you can change the `FWD_RELOAD` to check for the runtime error.
i.e.
```
FWD_RELOAD([122], [ignore], [ignore], [251])
```

erig0: As you noted on IRC.. the regression tests aren't ran for `firewall-offline-cmd`. Sorry for misleading. We should still add `--permanent` flag though.

erig0:
merged. Thanks Phil!

PR #663, opened 2020-06-29
fix(cli): add ipset type hash:mac is incompatible with the family par…

vrindle:
…ameter

Fixes: rhbz1541077


erig0:
Great work Vrinda! Thanks.

PR #665, opened 2020-07-07
improvement(service): IPsec: Update description and add TCP port 4500

letoams:
IKE and IPsec over TCP is defined in RFC 8229. It specifically mentions
no ports to allow administrators to configure any port to prevent being
blocked by networks.

However, most IKE/IPsec blocking seems to come from unwanted accidental
UDP blocks, so any TCP would usually ensures IPsec can still work on
such networks. The default is therefor to pick the same TCP port as IKE
and IPsec over UDP uses, port 4500.


erig0:
Thanks. Merged.

PR #678, opened 2020-08-11
feat(rich): add tcp-mss-clamp support

vrindle:
feat(rich): add tcp-mss-clamp support


src/firewall/core/io/zone.py
```diff
@@ -654,6 +655,18 @@ def startElement(self, name, attrs):
...
+        elif name == "tcp-mss-clamp":
+            value = None
+            if "value" in attrs:
+                value = attrs["value"]
+                if value not in ["pmtu"] or not value.isdigit()
+                    log.warning('Invalid tcp-mss-clamp: Value "%s" invalid',
+                                attrs["value"])
+                    self._rule_error = True
+                    return
+             self._rule = rich.Rich_Tcp_Mss_clamp(value=value)
```
erig0, line 668: Should be setting `self._rule.element` here. You also need to add the `Rich_Tcp_Mss_Clamp` class before any of the XML parsing will work.

erig0, line 668: This whole bit needs to be guarded by `if self._rule` as tcp-mss-clamp is only valid in rich rules. It's not a zone level feature.

src/firewall/core/io/zone.py
```diff
@@ -835,6 +848,13 @@ def zone_writer(zone, path=None):
         handler.ignorableWhitespace("  ")
         handler.simpleElement("masquerade", { })
         handler.ignorableWhitespace("\n")
+ 
+    # tcp-mss-clamp
+    if zone.tcp_mss_clamp:
+        attrs = { }
+        handler.ignorableWhitespace("  ")
+        handler.simpleElement("tcp-mss-clamp:", {"value": value })
+        handler.ignorableWhitespace("\n")
```
erig0, line 857: Similar comment as above. This only applies to rich rules so it needs to be moved down inside the `# rules` section.

src/firewall/core/io/zone.py
```diff
@@ -654,6 +655,19 @@ def startElement(self, name, attrs):
             if "priority" in attrs:
                 priority = int(attrs["priority"])
             self._rule = rich.Rich_Rule(family=family, priority=priority)
+        
+        elif name == "tcp-mss-clamp":
+            if self._rule:
+                value = None
+                if "value" in attrs:
+                    value = attrs["value"]
+                    if value != "pmtu":
```
erig0, line 664: We should also accept integer values, e.g. 1500, 1450, etc.

src/firewall/core/rich.py
```diff
@@ -417,6 +426,12 @@ def _import_from_string(self, rule_str):
                     in_elements.pop() # protocol
                 else:
                     raise FirewallError(errors.INVALID_RULE, "invalid 'protocol' element")
+            elif in_element == 'tcp-mss-clamp':
+                if attr_name == 'value':
+                    self.element = Rich_Tcp_Mss_Clamp(attr_value)
+                    in_elements.pop()   #tcp-mss-clamp
+                else:
```
erig0, line 433: I think `value` is optional, right? This will throw an error for the following:
```
firewall-cmd --add-rich-rule='rule family=ipv4 tcp-mss-clamp'
```

You should add test coverage for this.

src/firewall/core/rich.py
```diff
@@ -655,16 +670,20 @@ def check(self):
...
                 raise FirewallError(errors.INVALID_PROTOCOL, self.element.protocol)
-
-        # other element and not empty?
-        elif self.element is not None:
-            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" % 
-                                type(self.element))
+        
+        # tcp-mss-clamp
+        elif type(self.element) == Rich_Tcp_Mss_Clamp:
+            pass
```
erig0, line 681: Should check if `value` is: pmtu, [>= 536](https://en.wikipedia.org/wiki/Maximum_segment_size)

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,24 @@
+dnl ###############
+dnl ## XML tests ##
+dnl ###############
+
+FWD_START_TEST([tcp-mss-clamp])
+AT_KEYWORDS(tcp-mss-clamp)
+
+AT_CHECK([mkdir -p ./zones])
+
+AT_DATA([./zones/foobar.xml], [m4_strip([dnl
```
erig0, line 10: The `m4_strip()` is unnecessary.

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,24 @@
...
+<zone>
+  <service name="ssh"/>
+  <service name="dhcpv6-client"/>
+  <rule>
+    <tcp-mss-clamp value="ba"/>
+    <accept/>
+  </rule>
+</zone>
+])
+FWD_CHECK([--check-config],0)
```
erig0, line 21: Add `--permanent` flag so `firewall-offline-cmd` gets run as well.

src/firewall/core/io/policy.py
```diff
@@ -77,6 +77,19 @@ def common_startElement(obj, name, attrs):
...
+        if obj._rule:
+            if obj._rule.element:            
+                log.warning("Invalid rule: More than one element in rule '%s', ignoring.",
+                            str(obj._rule))
+                obj._rule_error = True
+                return True
+            obj._rule.element = rich.Rich_Tcp_Mss_Clamp(attrs["value"])
+        else:
+            log.warning("Tcp-mss-clamp value '%s' already set, ignoring.",
+                        attrs["value"])
```
erig0, line 91: This is non-sense. The entire `else` block in invalid because we should never encounter the `tcp-mss-clamp` element outside of rich rules. I think you can simply remove the entire `else` block.. or better yet throw an error with a message "tcp-mss-clamp is only valid inside of rich elements."

src/firewall/core/rich.py
```diff
@@ -655,16 +670,29 @@ def check(self):
...
                 raise FirewallError(errors.INVALID_PROTOCOL, self.element.protocol)
-
-        # other element and not empty?
-        elif self.element is not None:
-            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" % 
-                                type(self.element))
+        
+        # tcp-mss-clamp
+        elif type(self.element) == Rich_Tcp_Mss_Clamp:
+            if str(self.element.value):
```
erig0, line 681: The XML attribute value should always be a string. Using `str()` in this area is unnecessary.

src/firewall/core/rich.py
```diff
@@ -655,16 +670,29 @@ def check(self):
...
+                if ((str(self.element.value)).isdigit()):
+                    if int(str(self.element.value)) < 536:
+                        raise FirewallError(errors.INVALID_RULE, self.element.value)
+                    else:
+                        pass
```
erig0, line 686: `else` block is redundant.

src/firewall/core/rich.py
```diff
@@ -655,16 +670,29 @@ def check(self):
...
+                elif str(self.element.value) != "pmtu":
+                    raise FirewallError(errors.INVALID_RULE, self.element.value)
+                else:
+                    pass
```
erig0, line 690: `else` block is redundant.

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,24 @@
...
+  <rule>
+    <tcp-mss-clamp value="536"/>
+    <accept/>
+  </rule>
+</zone>
+])
+FWD_CHECK([--check-config],0, ignore)
+FWD_RELOAD
+FWD_END_TEST()
+
```
erig0, line 24: Please also add a check for if `value` attribute is not specified.

erig0 (commented):
Some clean up comments, but otherwise it looks good. Time to move on the CLI.

src/firewall/core/ipXtables.py
```diff
@@ -1080,14 +1081,16 @@ def _rich_rule_audit(self, policy, rich_rule, enable, table, rule_fragment):
...
         add_del = { True: "-A", False: "-D" }[enable]
-
+        
         _policy = self._fw.policy.policy_base_chain_name(policy, table, POLICY_CHAIN_PREFIX)
 
         chain_suffix = self._rich_rule_chain_suffix(rich_rule)
         chain = "%s_%s" % (_policy, chain_suffix)
-        if type(rich_rule.action) == Rich_Accept:
+        
+        if "TCPMSS" in rule_fragment:
```
vrindle, line 1091: I am not sure if this is the best way to solve the problem. Without this conditional 2 -j flags are addded which causes an error to occur. If you think there is a better way please let me know @erig0.

src/firewall/core/fw_policy.py
```diff
@@ -833,6 +834,10 @@ def check_protocol(self, protocol):
         if not checkProtocol(protocol):
             raise FirewallError(errors.INVALID_PROTOCOL, protocol)
 
+    def check_tcp_mss_clamp(self, tcp_mss_clamp_value):
+        if not checkTcpMssClamp(tcp_mss_clamp_value):
+            raise FirewallError(errors.INVALID_RULE, tcp_mss_clamp_value)
```
erig0, line 839: This error string could be better. Maybe something like: "tcp-mss-clamp value must be greater than or equal to 536, or the special value 'pmtu'. Invalid value '%s'" % (tcp_mss_clamp_value)

src/firewall/core/ipXtables.py
erig0: Since TCP/MSS is itself an action you should not be calling this function from `build_policy_tcp_mss_clamp_rules()`.

src/firewall/core/rich.py
```diff
@@ -655,16 +670,25 @@ def check(self):
...
-                                type(self.element))
+        
+        # tcp-mss-clamp
+        elif type(self.element) == Rich_Tcp_Mss_Clamp:
+            if self.element.value:
+                if (self.element.value).isdigit():
+                    if int(self.element.value) < 536:
+                        raise FirewallError(errors.INVALID_RULE, self.element.value)
+                elif self.element.value != "pmtu":
+                    raise FirewallError(errors.INVALID_RULE, self.element.value)
```
erig0, line 686: This seems redundant with the checks in `checkTcpMssClamp()`. Why not just call that function?

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,81 @@
...
+    <tcp-mss-clamp value="thisdoesnotexist"/>
+    <accept/>
+  </rule>
+</zone>
+])
+m4_ifdef([TESTING_FIREWALL_OFFLINE_CMD], [FWD_CHECK([--check-config], 0, ignore, [dnl
+WARNING: INVALID_RULE: thisdoesnotexist: rule tcp-mss-clamp value="thisdoesnotexist" accept
+WARNING: INVALID_RULE: thisdoesnotexist: rule tcp-mss-clamp value="thisdoesnotexist" accept
+])], [FWD_CHECK([--check-config], 0, ignore, [dnl
+])])
```
erig0, line 51: Since the "online" tests cause the warning to be printed to the log file you can check the expected warning get logged with the macro `FWD_GREP_LOG`. This is only needed since the error code is wrong (0).

erig0 (changes requested):
There are a handful of unnecessary white space changes that I'd like you to remove. Otherwise this is looking good.

erig0 (changes requested):
Otherwise looks good. Please make corrections, add docs, fix commit messages, and push again for another review.

src/firewall/core/io/policy.py
```diff
@@ -77,6 +77,16 @@ def common_startElement(obj, name, attrs):
...
+    
+    elif name == "tcp-mss-clamp":
+        if obj._rule:
+            if obj._rule.element:            
+                log.warning("Invalid rule: More than one element in rule '%s', ignoring.",
+                            str(obj._rule))
+                obj._rule_error = True
+                return True
+            obj._rule.element = rich.Rich_Tcp_Mss_Clamp(attrs["value"])
+
```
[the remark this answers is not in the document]
vrindle: Would it be best to leave a FirewallError rather than a log.warning? Because looking at similar code, with a similar errors it says something like log.warning("Invalid rule: icmp-block '%s' outside of rule",attrs["name"]) rather than raising FirewallError.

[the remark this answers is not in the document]
erig0: Use `log.warning()`... that way the XML parsing can continue. Seems like the only place we raise exceptions in this area is inside the check functions.

erig0 (changes requested):
The whitespace fixups in commit `chore(rich): fix whitespace for tcp-mss-clamp ` should be squashed into the commit that introduced the changes.

doc/xml/firewalld.zone.xml
```diff
@@ -181,6 +181,22 @@
...
+        <varlistentry>
+          <term>value="<replaceable>value</replaceable>"</term>
+          <listitem>
+            <para>
+              Value has to 'pmtu' or be greater than or equal to 536.
+            </para>
+          </listitem>
+        </varlistentry>
+      </variablelist>
+    </refsect2>
```
erig0, line 199: This should be moved to `policy_zone_descriptions.xml`.

erig0 (changes requested):

src/firewall/core/rich.py
```diff
@@ -661,10 +674,16 @@ def check(self):
...
 
-        # other element and not empty?
-        elif self.element is not None:
-            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" % 
-                                type(self.element))
+        # tcp-mss-clamp
+        elif type(self.element) == Rich_Tcp_Mss_Clamp:
+            if self.element.value:
+                if not functions.checkTcpMssClamp(self.element.value):
+                    raise FirewallError(errors.INVALID_RULE, self.element.value)
```
erig0, line 681: We also need to check that no action is specified. `tcp-mss-clamp` is an implicit action. This is similar to `masquerade`. This will also cause a change in your expected test output for nftables rule listings.
See here:

https://github.com/firewalld/firewalld/pull/678/files#diff-12066b4f26399116ab8d7b0c4ebc6198R629

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,102 @@
...
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=0 accept'],122,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=536 accept'],0,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=pmtu accept'],0,ignore, ignore)
+
+dnl start runtime tests for tcp/mss clamp
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=pmtu accept'],0,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=0 accept'],122,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp accept'],122,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=536 accept'],0,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=thisdoesnotexist accept'],122,ignore,ignore)
```
erig0, line 81: `tcp-mss-clamp` is mutually exclusive with an explicit action. So you'll have to remove the `accept` in all of these.

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,102 @@
...
+IPTABLES_LIST_RULES([filter], [FWDO_public_allow], 0, [dnl
+    TCPMSS tcp -- 0.0.0.0/0 0.0.0.0/0 tcp flags:0x06/0x02 TCPMSS clamp to PMTU
+    TCPMSS tcp -- 0.0.0.0/0 0.0.0.0/0 tcp flags:0x06/0x02 TCPMSS set 536
+    ])
+
+IP6TABLES_LIST_RULES([filter], [FWDO_public_allow], 0, [dnl
+    TCPMSS tcp ::/0 ::/0 tcp flags:0x06/0x02 TCPMSS clamp to PMTU
+    TCPMSS tcp ::/0 ::/0 tcp flags:0x06/0x02 TCPMSS set 536
+    ])
+
```
erig0, line 101: Please add negative tests to make sure `tcp-mss-clamp` with an aciton, e.g. `accept` is rejected.

erig0 (changes requested):

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,102 @@
...
+  </rule>
+</zone>
+])
+m4_ifdef([TESTING_FIREWALL_OFFLINE_CMD], [FWD_CHECK([--check-config], 2, ignore,ignore)], 
+[FWD_CHECK([--check-config],254, ignore, ignore)])
+
+dnl start cli tests (permanent configuration)
+AT_DATA(rm ./zones/tcp.xml)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=thisdoesnotexist accept'],122,ignore,ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp accept'],122,ignore, ignore)
```
erig0, line 71: A bare `tcp-mss-clamp` should be valid.. i.e `value` is optional.

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,102 @@
...
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=0 accept'],122,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=536 accept'],0,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=pmtu accept'],0,ignore, ignore)
+
+dnl start runtime tests for tcp/mss clamp
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=pmtu accept'],0,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=0 accept'],122,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp accept'],122,ignore, ignore)
```
erig0, line 79: After removing `accept` this should be a valid rule.

erig0 (changes requested):

erig0 (approved):

src/firewall/core/rich.py
```diff
@@ -661,10 +678,18 @@ def check(self):
             if self.element.protocol not in [ "tcp", "udp", "sctp", "dccp" ]:
                 raise FirewallError(errors.INVALID_PROTOCOL, self.element.protocol)
 
-        # other element and not empty?
-        elif self.element is not None:
-            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" % 
-                                type(self.element))
```
erig0, line 667: Still some whitespace errors here. That's why the diff is showing this section as deleted and then re-added. Looks trailing spaces or tabs.

src/firewall/core/rich.py
```diff
@@ -417,6 +424,15 @@ def _import_from_string(self, rule_str):
                     in_elements.pop() # protocol
                 else:
                     raise FirewallError(errors.INVALID_RULE, "invalid 'protocol' element")
+            elif in_element == 'tcp-mss-clamp':
+                if attr_name == 'value':
+                    self.element = Rich_Tcp_Mss_Clamp(attr_value)
+                    in_elements.pop()
```
erig0, line 430: It's unnecessary to create the `Rich_Tcp_Mss_Clamp` instance here. Check other examples.. they set `attrs` which will be used later in the `else` block.

src/firewall/core/rich.py
```diff
@@ -661,10 +678,18 @@ def check(self):
...
                 raise FirewallError(errors.INVALID_PROTOCOL, self.element.protocol)
 
-        # other element and not empty?
-        elif self.element is not None:
-            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" % 
-                                type(self.element))
+        # tcp-mss-clamp
+        elif type(self.element) == Rich_Tcp_Mss_Clamp:
+            if self.action is not None:
+                raise FirewallError(errors.INVALID_RULE, "tcp-mss-clamp and action")
```
erig0, line 684: Please improve this error message.

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,98 @@
...
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=thisdoesnotexist'],122,ignore,ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp'],0,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=0'],122,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=536'],0,ignore, ignore)
+FWD_CHECK([--permanent --add-rich-rule='rule tcp-mss-clamp value=pmtu'],0,ignore, ignore)
+
+dnl start runtime tests for tcp/mss clamp
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=pmtu'],0,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp value=0'],122,ignore, ignore)
+FWD_CHECK([--add-rich-rule='rule tcp-mss-clamp'],0,ignore, ignore)
```
erig0, line 74: Can you verify doing a `--delete-rich-rule` or `--query-rich-rule` still works if `value` is omitted? It's worth adding coverage for these.

src/tests/features/rich_tcp_mss_clamp.at
```diff
@@ -0,0 +1,98 @@
...
+  <service name="dhcpv6-client"/>
+  <rule>
+    <tcp-mss-clamp/>
+  </rule>
+</zone>
+])
+FWD_CHECK([--check-config], 0, ignore,ignore)
+
+dnl start cli tests (permanent configuration)
+AT_DATA(rm ./zones/tcp.xml)
```
erig0, line 64: This should be `AT_CHECK`, not `AT_DATA`.

erig0 (changes requested):
This _very_ close. Some suggested cleanups and improvements. I suspected the next version will be mergeable.

src/firewall/core/rich.py
erig0: Still unnecessary white space changes here.

erig0 (changes requested):

src/firewall/core/rich.py
```diff
@@ -661,9 +677,17 @@ def check(self):
...
+            if self.action is not None:
+                raise FirewallError(errors.INVALID_RULE, "tcp-mss-clamp and %s are mutually exclusive" % self.action)
+            if self.element.value:
+                if not functions.checkTcpMssClamp(self.element.value):
+                    raise FirewallError(errors.INVALID_RULE, self.element.value)
+
         # other element and not empty?
         elif self.element is not None:
-            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" % 
+            raise FirewallError(errors.INVALID_RULE, "Unknown element %s" %
```
erig0, line 690: Almost. :)

erig0 (changes requested):

erig0 (approved):

src/tests/features/rich_tcp_mss_clamp.at
vrindle: Firewalld doesn't print the warning to the log file unless the debug flag is enabled. This means that without the --debug flag FWD_GREP_LOG fails. Do you have any suggestions on how to get around this? I tested out a number of things and the only way I know how to resolve this issue is reverting back to the code snippet above.

erig0: Hrm. I thought WARNING logs were always printed. Even without `--debug`

erig0: The problem is actually because you're not calling `--check-config` before `FWD_GREP_LOG`.

PR #684, opened 2020-08-26
fix(zone): cache rule_str for rich rules

erig0:
@todoleza, please take a look at the test case. I normally avoid anything time dependent in tests, but I didn't know what else to do for this.


todoleza:
My tests for the speedup shown no easily comparable results so I think that the 45s set as a timeout in the test are okay.
The limit might prove small for higher parallelization but there is still a space.
```
# of rich rules: 1000     5000     10000
this PR        0.748s   5.086s   16.331s
master         2.331s  45.336s  186.161s
```

todoleza (approved):

PR #692, opened 2020-09-17
fix(config): Fix RPM macros to test if firewall-cmd is executable

Conan-Kudo:
While unlikely, we cannot exactly predict how a running system has its
permissions set, but we can guard against it to prevent weird failures.

This should have no appreciable impact, but ensures that we do not try
to run `firewall-cmd` if it is not actually executable.


marcosfrm:
Straightforward change.

erig0 (approved):

Issue #691, opened 2020-09-17
test rhbz1871298 timeouts on slow architectures


LocutusOfBorg:
Hello, looks like in Ubuntu/arm64 we are getting a test failure on rhbz1871298.

I patched to increase the timeout from 45 seconds to 450 (not a good number, I know), but the test passed.

I don't know which value might be a good number, but can you please increase it a little bit?

http://launchpadlibrarian.net/497760591/firewalld_0.9.0-1_0.9.0-1ubuntu1.diff.gz

https://autopkgtest.ubuntu.com/packages/f/firewalld/groovy/arm64

cc @erig0

LocutusOfBorg:
(it would be really nice, as test feature, to print the seconds passed for each test, so we can quickly understand how different architectures performs

erig0:
Fixed via 1401c61620a69 (#969).

PR #698, opened 2020-10-18
Update rhbz1871298 to fix a timeout on slow arm architectures (See: #691)

LocutusOfBorg:
Some architectures, such as arm64, cannot finish the test in time, relaxing a little bit the timeout helps them in figuring out the test result

Closes: #691 
Look e.g. https://autopkgtest.ubuntu.com/packages/f/firewalld/groovy/arm64
https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-groovy/groovy/arm64/f/firewalld/20201005_090931_e08a1@/log.gz

### DIFF
--- a/src/tests/regression/rhbz1871298.at
+++ b/src/tests/regression/rhbz1871298.at
@@ -13,6 +13,6 @@ NS_CHECK([echo "</zone>" >> ./zones/foobar.xml])
 if test "x${FIREWALLD_DEFAULT_CONFIG}" != x ; then
     FIREWALL_OFFLINE_CMD_ARGS+=" --default-config ${FIREWALLD_DEFAULT_CONFIG}"
 fi
-NS_CHECK([timeout 45 firewall-offline-cmd --system-config ./ $FIREWALL_OFFLINE_CMD_ARGS --check-config], 0, [ignore])
+NS_CHECK([timeout 75 firewall-offline-cmd --system-config ./ $FIREWALL_OFFLINE_CMD_ARGS --check-config], 0, [ignore])
 
 FWD_END_TEST


erig0:
@todoleza , are you okay bumping this from 45 to 90? I originally hesitated committing this test. Timeout based tests like this are very error prone as @LocutusOfBorg has found. Maybe we should simply scrap this upstream test?

derekdai:
Since firewalld talk to nftables through libnftables1 in JSON, the parsing and serializing speed does matter on slow machine.

Although average rules in firewall won't that much, but ipset will be. For example, I have 20k+ entries in my ipset, load by firewalld. If backend is nftables, it takes 2m+ to `--reload` on my PI4. Reloading time of iptables backend is just OK, no timeout.

The slowness is caused by two issues, both relate to JSON.

First issue is in libnftables1 which look up command by seqnum linearly. The issue and patch are here
https://bugzilla.netfilter.org/show_bug.cgi?id=1479

The second issue is caused by slow JSON library which firewalld using. Will it be possible migrate from the Python official on to [orjson](https://github.com/ijl/orjson) or some other faster one?

erig0:
> Hi @erig0, since firewalld talk to nftables through libnftables1 in JSON, in my experience the parsing and serializing speed does matter on slow machine.
> 
> Although average rules in firewall won't that much, but ipset will be. For example, I have 20k+ entries in my ipset, load by firewalld. If backend is nftables, it takes 2m+ to `--reload` on my PI4. Reloading time of iptables backend is just OK, no timeout.

This particular issue was fixed by this nftables change `c96c7da272e3 ("JSON: Improve performance of json_events_cb()")`

erig0:
Furthermore, this particular report is about the XML parsing. It has nothing to do with the libnftables JSON backend.

derekdai:
> Furthermore, this particular report is about the XML parsing. It has nothing to do with the libnftables JSON backend.

Sorry for off topic. But I think both XML and JSON processing speed need to be addressed.

The `c96c7da272e3` reduces call to `seqnum_to_json()`, but [this patch](https://bugzilla.netfilter.org/show_bug.cgi?id=1479) improve efficiency of `seqnum_to_json()` - which reduces ipset loading time on my PI 4 from 2m+ down to 10s+.

erig0:
On Tue, Nov 10, 2020 at 08:36:51PM -0800, Derek 呆 wrote:
> > Furthermore, this particular report is about the XML parsing. It has nothing to do with the libnftables JSON backend.
> 
> Sorry for off topic. But I think both XML and JSON processing speed need to be addressed.
> 
> The `c96c7da272e3` reduces call to `seqnum_to_json()`, but [this
> patch](https://bugzilla.netfilter.org/show_bug.cgi?id=1479) improve
> efficiency of `seqnum_to_json()` - which reduces ipset loading time on
> my PI 4 from 2m+ down to 10s+.

I pinged one of the nftables maintainers to take a look at this patch.
Thanks!

LocutusOfBorg:
ping?

LocutusOfBorg:
@erig0 mentioned a NM fix somewhere that might be missing in Ubuntu.

erig0:
> 60s timeout looks enough now to make the test pass.

Maybe on your hardware. I hesitated to even add this test because it's sensitive to the hardware/environment. I think we should simply remove the test. It's cause more pain than benefit.

erig0:
> @erig0 mentioned a NM fix somewhere that might be missing in Ubuntu.

This was in regards to the networkmanager-ovs plugin. One of the integration tests relies upon that, but the test failure reported here does not. I'm hiding both comments in this thread to avoid confusion.

LocutusOfBorg:
FYI 60 seconds is not enough on ubuntu arm64 builders.

erig0:
> FYI 60 seconds is not enough on ubuntu arm64 builders.

Yea.. let's remove this test. There will always be super slow architectures or shared VMs that will have issues with this performance based test.

LocutusOfBorg:
thanks!

LocutusOfBorg:
it would be nice to print how long did it take to complete :)

erig0:
> it would be nice to print how long did it take to complete :)

Then used `time` instead of `timeout`.

erig0:
Sorry. I think I dropped the ball here. This timeout was bumped in commit 1401c61620a69 (#969).

As such, closing this PR.

Issue #705, opened 2020-11-05
Policy HOST -> "SOME ZONE" is wrong


VadimSvyatsky:
If the outbound policy uses a service or port based on tcp, connection is not established.
The likely reason is that there is no rule in the chain "OUTPUT" for states "RELATED,ESTABLISHED"

erig0:
Can you share your configuration. Specifically any policies you've created.

VadimSvyatsky:
<?xml version="1.0" encoding="utf-8"?>
<policy target="REJECT">
  <ingress-zone name="HOST"/>
  <egress-zone name="external"/>
  <rule family="ipv4" priority="1">
    <destination address="195.177.240.0/23"/>
    <service name="ntp"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="2">
    <destination address="195.177.240.0/23"/>
    <service name="dns"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="3">
    <destination address="195.177.240.0/23"/>
    <service name="smtp"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="4">
    <service name="http"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="5">
    <service name="https"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="6">
    <log prefix="debug: "/>
  </rule>
</policy>

VadimSvyatsky:
`<?xml version="1.0" encoding="utf-8"?>
<policy target="REJECT">
  <ingress-zone name="HOST"/>
  <egress-zone name="external"/>
  <rule family="ipv4" priority="1">
    <destination address="195.177.240.0/23"/>
    <service name="ntp"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="2">
    <destination address="195.177.240.0/23"/>
    <service name="dns"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="3">
    <destination address="195.177.240.0/23"/>
    <service name="smtp"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="4">
    <service name="http"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="5">
    <service name="https"/>
    <accept/>
  </rule>
  <rule family="ipv4" priority="6">
    <log prefix="debug: "/>
  </rule>
</policy>
`


In this configuration all udp-based services is working and all tcp-based services - is not working

erig0:
[...]

PR #709, opened 2020-11-09
fix(policy): generic accept of established connections in OUTPUT

erig0:
Now that policies allow filtering on the OUTPUT we must also generically
accept established connections in OUTPUT. Otherwise a policy that blocks
all OUTPUT will cause even selectively allowed traffic to drop. This is
because "allows" match on "ct state new", which only matches on the
_first_ packet.

Fixes: #705

### DIFF
--- a/src/firewall/core/ipXtables.py
+++ b/src/firewall/core/ipXtables.py
@@ -757,6 +757,7 @@ class ip4tables(object):
         default_rules["filter"] += [
             "-N OUTPUT_direct",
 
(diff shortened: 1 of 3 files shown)


VadimSvyatsky:
Hello
I have applied pr #709
and then i compiled the firewalld.
Everything works now.
This issue can be clossed.

VadimSvyatsky:
I have compiled the code from the branch "master".
And.............
Error: PARSE_ERROR: 'INT-EXT.xml': Missing attribute address for destination

PR #714, opened 2020-11-13
fix(rich): non-printable characters removed from rich rules

vrindle:
Fixes: rhbz 1596304


src/firewall/functions.py
```diff
@@ -311,6 +311,13 @@ def checkIPnMask(ip):
                 return False
     return True
 
+NOPRINT_TRANS_TABLE = {
+    i: None for i in range(0, sys.maxunicode + 1) if not chr(i).isprintable()
+}
+
+def stripCharacters(rule_str, table):
```
erig0, line 318: Let's create a second function called `stripNonPrintable`. Then we don't need to expose NOPRINT_TRANS_TABLE.

erig0 (changes requested):
No other comments other than that it's disappointing this isn't easier to do in Python. :/

src/tests/regression/rhbz1596304.at
```diff
@@ -0,0 +1,24 @@
+FWD_START_TEST([rich rules strip non-printable characters])
+	AT_KEYWORDS(rich rhbz1596304)
```
erig0, line 2: looks like a wild tab here.

src/firewall/functions.py
```diff
@@ -311,6 +311,17 @@ def checkIPnMask(ip):
                 return False
     return True
 
+def stripNonPrintableCharacters():
+    NOPRINT_TRANS_TABLE = {
+        i: None for i in range(0, sys.maxunicode + 1) if not chr(i).isprintable()
+    }
```
erig0, line 317: The table should still be initialized at the file level. Otherwise we regenerate it every time this function is called. I expect this function to call `rule_str.translate()` with `NOPRINT_TRANS_TABLE` as an argument.

src/firewall/core/rich.py
```diff
@@ -328,6 +328,8 @@ def _import_from_string(self, rule_str):
         if not rule_str:
             raise FirewallError(errors.INVALID_RULE, 'empty rule')
 
+        rule_str = functions.stripCharacters(rule_str)
```
erig0, line 331: Should call `stripNonPrintableCharacters()` here as that's actually what the function is actually doing.

src/firewall/functions.py
```diff
@@ -311,6 +311,17 @@ def checkIPnMask(ip):
...
+    return NOPRINT_TRANS_TABLE
+
+def stripCharacters(rule_str):
+    table = stripNonPrintableCharacters()
+
+    return rule_str.translate(table)
```
erig0, line 323: No need for this function after changing `stripNonPrintableCharacters()` as per other comments.

erig0 (changes requested):

erig0:
LGTM.. Add "Fixes: #480" to the commit message and I'll merge.

erig0 (approved):

erig0:
This fix caused a drastic increase in memory usage. The building of `NOPRINT_TRANS_TABLE` is the cause. I will attempt to find a solution, but may simply revert this change. Unfortunately it made it into all the recent releases. :(

vrindle:
Oh that's a shame :( Hopefully you can find a solution for this fix soon :)

erig0:
@vrindle , it was fixed in commit 015704b44f81d535a868fe28368f977cefd28638.

Issue #703, opened 2020-10-28
Trouble with --add-source=<MAC address>


Jonrysh:
An attempt to add a MAC style source to a zone fails.
```
# firewall-cmd --zone=home --add-source=34:7e:5c:3a:4c:32
Error: __init__() takes from 2 to 3 positional arguments but 4 were given
```
This looks like a bug.  If so, where should it be reported?

erig0:
> This looks like a bug. If so, where should it be reported?

You just did. Thanks!

Jonrysh:
Does anyone know a work around?

erig0:
On Wed, Oct 28, 2020 at 07:55:56AM -0700, Jonrysh wrote:
> Does anyone know a work around?

I don't think there is a workaround.

Jonrysh:
Thanks

Jonrysh:
I'm having a similar problem with "--add-source=IP-address .  Is it possible that my python installation is misconfigured?  How can I test it?  Log of the failure follows.
```
# firewall-cmd --zone=home --add-source=192.168.1.69
Error: COMMAND_FAILED: 'python-nftables' failed: internal:0:0-0: Error: Could not process rule: No such file or directory

internal:0:0-0: Error: Could not process rule: No such file or directory

internal:0:0-0: Error: Could not process rule: No such file or directory

internal:0:0-0: Error: Could not process rule: No such file or directory

internal:0:0-0: Error: Could not process rule: No such file or directory

internal:0:0-0: Error: Could not process rule: No such file or directory

internal:0:0-0: Error: Could not process rule: No such file or directory


JSON blob:
[...]

PR #715, opened 2020-11-18
fix: add source with mac address

vrindle:
Fixes: #703


erig0:
Please add the following Fixes tags to your commit message:
```
Fixes: df4aefcbe7b7 ("improvement(ipXtables): add utility function match sources")
Fixes: 1582c5dd736a ("feat: nftables: convert to libnftables JSON interface")
```

src/firewall/core/ipXtables.py
```diff
@@ -906,7 +906,7 @@ def _rule_addr_fragment(self, opt, address, invert=False):
         elif check_mac(address):
             # outgoing can not be set
             if opt == "-d":
-                raise FirewallError(INVALID_ADDR, "Can't match a destination MAC.")
+                return [""]
```
erig0, line 909: Just an empty list. No need for the empty string.

erig0 (changes requested):

src/firewall/core/ipXtables.py
vrindle: Returning just an empty list causes the test to fail.

erig0:
On Fri, Nov 20, 2020 at 09:50:16AM -0800, Vrindle wrote:
> @vrindle commented on this pull request.
> 
> 
> 
> > @@ -906,7 +906,7 @@ def _rule_addr_fragment(self, opt, address, invert=False):
>          elif check_mac(address):
>              # outgoing can not be set
>              if opt == "-d":
> -                raise FirewallError(INVALID_ADDR, "Can't match a destination MAC.")
> +                return [""]
> 
> Returning just an empty list causes the test to fail. 

Can you share more details about the test failure?

vrindle:
When returning only an empty list the error is:
`+Error: COMMAND_FAILED: '/usr/sbin/ip6tables-restore -w -n' failed: ip6tables-restore: line 83 failed`

vrindle:
@erig0 just double checking, [I've confirmed tests fail with an empty list](https://github.com/firewalld/firewalld/pull/715#issuecomment-732338229) Is this PR good to go or is anything you would like me to change?

src/firewall/core/nftables.py
```diff
@@ -1217,7 +1217,7 @@ def _rule_addr_fragment(self, addr_field, address, invert=False):
         else:
             if check_mac(address):
                 if addr_field == "daddr":
-                    raise FirewallError(INVALID_RULE, "%s._rule_addr_fragment()", (self.__class__))
+                    return ""
```
erig0, line 1220: I think this should return `{}` instead of `""`. That's what other functions don that are support to return a fragment.

erig0 (changes requested):
> When returning only an empty list the error is:
> `+Error: COMMAND_FAILED: '/usr/sbin/ip6tables-restore -w -n' failed: ip6tables-restore: line 83 failed`

Please find out why. I still think returning the truly empty list should work. Use the `-v -e` flags when running the testsuite.

erig0 (approved):

erig0:
Hi Vrinda, I hope you don't mind that I finished this up and merged it. Thanks!

vrindle:
@erig0 no problem :D

PR #724, opened 2020-12-01
feat(service): add galera service

vrindle:
Fixes: rhbz1696260

### DIFF
--- a/config/services/galera.xml
+++ b/config/services/galera.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Galera</short>
+  <description>MariaDB-Galera Database Server</description>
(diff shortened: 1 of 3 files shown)


config/services/galera.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Galera</short>
+  <description>MariaDB-Galera Database Server</description>
+  <port protocol="tcp" port="3306"/>
+  <port protocol="tcp" port="4567"/>
```
erig0, line 6: Just noting that as per https://galeracluster.com/library/documentation/firewall-settings.html, this port can *optionally* be opened for multicast replication. I think it's okay to keep it closed until a user requests the UDP port.

erig0 (commented):
LGTM. Thanks V.

PR #754, opened 2021-02-03
ci: add nftables-master to test matrix

erig0:
The testsuite often catches regressions in nftables. Lets always run
against nftables master to catch regression earlier.

### DIFF
--- a/src/tests/functions.at
+++ b/src/tests/functions.at
@@ -392,6 +392,7 @@ m4_define([NFT_LIST_RULES_NORMALIZE], [dnl
         -e '/type.*hook.*priority.*policy.*/d'dnl
         dnl tranform ct state { established,related } to ct state established,related
         -e '/ct \(state\|status\)/{s/\(ct \(state\|status\)\) {/\1/g; s/ }//; s/\(@<:@a-z@:>@*\), /\1,/g;}' dnl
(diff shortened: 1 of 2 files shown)


erig0:
@todoleza, tagging you only as FYI.

src/tests/functions.at
```diff
@@ -392,6 +392,7 @@ m4_define([NFT_LIST_RULES_NORMALIZE], [dnl
         -e '/type.*hook.*priority.*policy.*/d'dnl
         dnl tranform ct state { established,related } to ct state established,related
         -e '/ct \(state\|status\)/{s/\(ct \(state\|status\)\) {/\1/g; s/ }//; s/\(@<:@a-z@:>@*\), /\1,/g;}' dnl
+        -e 's/reject with icmp[[x6]]\? type port-unreachable/reject/' dnl
```
n0-1, line 395: ICMPv6 syntax is: 'reject with icmpv6 ...'. So the match part should read: 'reject with icmp\(x\|v6\)\? type port-unreachable'.

erig0: Thanks for catching that. I'll push a follow up.

erig0: I directly pushed commit 3a3b4676ccb7b40cf304b773456dec2662783425.

PR #755, opened 2021-02-08
Added OMG DDS rules.

pkubanek:


config/services/dds.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>OMG DDS</short>
+  <description>Open Management Group (OMG) Data Distribution Service (DDS) is protocol supporting various applications. It is usally used in control systems. The provided set is for domain ID 0 and maximum (120) number of applications. Please see https://community.rti.com/content/forum-topic/statically-configure-firewall-let-omg-dds-traffic-through for details.</description>
+  <port protocol="udp" port="7400-7649"/>
+  <destination ipv4="239.255.0.1"/>
```
erig0, line 6: This will scope the ports to the destination as well. i.e. this only accepts packets if the destination is `239.255.0.1` _and_ the destination port is in the range `7400-7649`. According to the OMG DDS docs only ports `7400-7401` listen for multicast. Other ports are unicast.

To accomplish this I think you need three serivces:
  1. `dds-multicast` which uses `destination` of `239.255.0.1` and ports `7400-7401`
  2. `dds-unicast` which uses ports `7402-7649`
  3. `dds` which has an `include` for both `dds-multicast` and `dds-unicast`
 

Lastly, you need to update files `config/Makefile.am` and `po/POTFILES.in`.

erig0 (changes requested):

config/services/dds.xml
pkubanek: Thanks for suggestions. I didn't know about include, as I was using old firewalld version available on CentOS 7. Working on fixes.

pkubanek: Hopefully resolved, rebased & force pushed the changes.

PR #768, opened 2021-03-06
Add ident service

Poorchop:


TorontoMedia:
Could you add this service file to `config/Makefile.am` and `po/POTFILES.in` and amend this commit with the following message commit type semantic `feat(service): Add ident service`?

erig0 (approved):
I updated the necessary build files. Will merge after CI.

PR #771, opened 2021-03-18
improvement(service):  Samba: Use service includes

ykgmfq:


marcosfrm:
`samba-client.xml` can include `netbios-ns.xml` .

`<helper name="netbios-ns"/>` can then be dropped from `samba.xml`, `samba-client.xml` and `samba-dc.xml`, no? (as far as I know it only works with 137/udp)

erig0:
> `samba-client.xml` can include `netbios-ns.xml` .
> 
> `<helper name="netbios-ns"/>` can then be dropped from `samba.xml`, `samba-client.xml` and `samba-dc.xml`, no? (as far as I know it only works with 137/udp)

It should go further than that. There is a also a `netbios-ns` service.
e.g. `samba-dc` include `samba` which includes `samba-client` which includes `netbios-ns`.

erig0 (approved):
LGTM. Thanks!

erig0:
Note: CI failure is due to an upstream nftables change in the rule output. I'll address it separately and not block this PR.

PR #780, opened 2021-04-10
feat(service): Add WireGuard service definition

gsauthof:
Of course, one can set up wireguard to listen on any other port.
However, 51820 is a good default that is used in many places.

### DIFF
--- a/config/services/wireguard.xml
+++ b/config/services/wireguard.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>WireGuard</short>
+  <description>WireGuard is the simple, fast and modern VPN. The port needs to be open if a peer has this host explicitly configured as endpoint.</description>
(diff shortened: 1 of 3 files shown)


config/services/wireguard.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>WireGuard</short>
+  <description>WireGuard is the simple, fast and modern VPN.</description>
```
erig0, line 4: AFAIK, this service is only need on the VPN server side. Clients won't need it. If that's accurate then please add a sentence or two to the description stating as such.

erig0 (changes requested):
Additionally files `config/Makefile.am` and `po/POTFILES.in` need to be updated.

erig0:
Commit message needs fixed, otherwise looks good. I'll merge after that's addressed.

erig0:
Thanks. Merged.

PR #785, opened 2021-04-15
ci(github): cleanup no longer needed dependencies

erig0:
six hasn't been needed since dropping python2 support in 74dd57ed3df2
("chore(README): drop python2 support") and following commits.

decorator hasn't been needed since it was dropped in e7fbb41ae113
("chore: drop dependency on python-decorator").


erig0:
Dropping PR. Still need decorator due to slip.

PR #793, opened 2021-05-05
drop dependency python-slip

erig0:
This drops the dependency on `python-slip`. Policy Kit authorization is now done via a custom decorator.

### DIFF
--- a/.github/workflows/source-checks.yml
+++ b/.github/workflows/source-checks.yml
@@ -9,7 +9,7 @@ jobs:
         iptables libdbus-1-dev libgirepository1.0-dev libglib2.0-dev \
         libxml2-utils nftables pkg-config python3-nftables xsltproc
       pip-dependencies: |
-        decorator dbus-python PyGObject flake8
+        dbus-python PyGObject flake8
 
     runs-on: ubuntu-20.04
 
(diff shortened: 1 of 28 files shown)


todoleza:
BUG: dbus communication broken after polkit has been either *stopped* or *restarted*. Then an ongoing error on any communication is present until firewalld is restarted as well.
``` 
# systemctl restart polkit
# firewall-cmd --list-all
ERROR:dbus.proxies:Introspect error on :1.44456:/org/fedoraproject/FirewallD1: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable
Error: The name is not activatable

erig0:
> BUG: dbus communication broken after polkit has been either _stopped_ or _restarted_. Then an ongoing error on any communication is present until firewalld is restarted as well.
> 
> ```
> # systemctl restart polkit
> # firewall-cmd --list-all
> ERROR:dbus.proxies:Introspect error on :1.44456:/org/fedoraproject/FirewallD1: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable
> Error: The name is not activatable
> ```

What happens if you also restart dbus-daemon?

erig0:
@tcanabrava, @lbiaggi, et al. plasma firewall, pinging you as a heads up about this change. The polkit auth was completely rewritten. In theory you should not notice anything, but in case you see any auth issues against firewalld v1.0.0, then suspect this change.

Thanks.
Eric.

PR #795, opened 2021-05-08
ipset remove_entry should remove the entry from its copy of the ipset list, or it won't let you re-add an entry

hackerbob:
firewall-cmd --permanent --new-ipset=test --type=hash:ip --option=timeout=0
firewall-cmd --reload
firewall-cmd --ipset=test --add-entry=1.2.3.4
success
firewall-cmd --ipset=test --remove-entry=1.2.3.4
success
firewall-cmd --ipset=test --add-entry=1.2.3.4
Warning: ALREADY_ENABLED: '1.2.3.4' already is in 'test'
success
firewall-cmd --ipset=test --remove-entry=1.2.3.4
Error: COMMAND_FAILED: 'python-nftables' failed: internal:0:0-0: Error: Could not process rule: No such file or directory

JSON blob:
{"nftables": [{"metainfo": {"json_schema_version": 1}}, {"delete": {"element": {"family": "inet", "table": "firewalld", "name": "test", "elem": ["1.2.3.4"]}}}]}


erig0:
Thanks for filing the issue and this PR. I updated your PR with proper commit message and added test case. Once CI passes I'll merge it.

erig0 (approved):

PR #813, opened 2021-06-30
improvement(testsuite): install to $(pkgdatadir)/testsuite

erig0:
Install the testsuite. This will make it easier for end users to verify
they have a sane environment for firewalld. Users often build custom
kernels which may be missing features firewalld relies upon.

--->8---

New RPM subpackage: firewalld-test

New files:
```
[root@vm-bos-rhel8-1 testsuite]# find /usr/share/firewalld/testsuite/
/usr/share/firewalld/testsuite/
/usr/share/firewalld/testsuite/README
/usr/share/firewalld/testsuite/testsuite
/usr/share/firewalld/testsuite/integration
/usr/share/firewalld/testsuite/integration/testsuite
/usr/share/firewalld/testsuite/python
/usr/share/firewalld/testsuite/python/firewalld_config.py
/usr/share/firewalld/testsuite/python/firewalld_direct.py
/usr/share/firewalld/testsuite/python/firewalld_rich.py
/usr/share/firewalld/testsuite/python/firewalld_test.py

[root@vm-bos-rhel8-1 testsuite]# du -sh /usr/share/firewalld/testsuite/
12M     /usr/share/firewalld/testsuite/
```

Run the base testsuite (equivalent of `make check`):
```
[root@vm-bos-rhel8-1 /tmp]# /usr/share/firewalld/testsuite/testsuite
## ----------------------------- ##
## firewalld 1.0.999 test suite. ##
## ----------------------------- ##

firewall-offline-cmd

  1: basic options                                   ok
  2: get/list options                                ok
[..]
```

Run the integration testsuite (equivalent of `make check-integration`):
```
[root@vm-bos-rhel8-1 /tmp]# /usr/share/firewalld/testsuite/integration/testsuite
## ----------------------------- ##
## firewalld 1.0.999 test suite. ##
## ----------------------------- ##

NetworkManager (nftables)

  1: NM overrides interface on reload                ok
  2: reload don't consider non IP capable interfaces ok
[..]
```

### DIFF
--- a/firewalld.spec
+++ b/firewalld.spec
@@ -49,6 +49,12 @@ Summary: Firewalld directory layout and rpm macros
 This package provides directories and rpm macros which
 are required by other packages that add firewalld configuration files.
 
+%package -n firewalld-test
+Summary: Firewalld testsuite
+
+%description -n firewalld-test
+This package provides the firewalld testsuite.
+
 %package -n firewall-applet
 Summary: Firewall panel applet
 Requires: %{name} = %{version}-%{release}
@@ -205,6 +211,18 @@ fi
 %dir %{_prefix}/lib/firewalld/services
 %dir %{_prefix}/lib/firewalld/zones
 %{_rpmconfigdir}/macros.d/macros.firewalld
+
+%files -n firewalld-test
+%dir %{_datadir}/firewalld/testsuite
+%{_datadir}/firewalld/testsuite/README
+%{_datadir}/firewalld/testsuite/testsuite
+%dir %{_datadir}/firewalld/testsuite/integration
+%{_datadir}/firewalld/testsuite/integration/testsuite
+%dir %{_datadir}/firewalld/testsuite/python
+%{_datadir}/firewalld/testsuite/python/firewalld_config.py
+%{_datadir}/firewalld/testsuite/python/firewalld_direct.py
+%{_datadir}/firewalld/testsuite/python/firewalld_rich.py
+%{_datadir}/firewalld/testsuite/python/firewalld_test.py
 
 %files -n firewall-applet
 %attr(0755,root,root) %dir %{_sysconfdir}/firewall
--- a/src/tests/Makefile.am
+++ b/src/tests/Makefile.am
@@ -20,6 +20,16 @@ EXTRA_DIST = \
 
 DISTCLEANFILES = atconfig
 
+testsuitedir = $(pkgdatadir)/testsuite
+dist_testsuite_SCRIPTS = $(TESTSUITE)
+dist_testsuite_DATA = README
+
+testsuite_integrationdir = $(pkgdatadir)/testsuite/integration
+dist_testsuite_integration_SCRIPTS = $(TESTSUITE_INTEGRATION)
+
+testsuite_pythondir = $(pkgdatadir)/testsuite/python
+dist_testsuite_python_DATA = python/*.py
+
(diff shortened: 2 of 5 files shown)


erig0:
@stepnem, if you're happy then approve the PR and it will automerge.

stepnem:
> NOTE: I can't find a way to make autotest NOT generate the following
> directories when run.

Do you mean `testsuite.dir` and `testsuite.log` (the latter is a
regular file)? If so, why? Do you not want the testsuite to produce
logs?

One glitch I noticed is that the installed test suite preserves the
source tree assumptions, e.g. in case of failures, the report still
references `src/tests` and `sed` tries to access `../../config.log`:

```
## ------------- ##
## Test results. ##
## ------------- ##

ERROR: All 38 tests were run,
32 failed unexpectedly.
sed: can't read ../../config.log: No such file or directory
## -------------------------- ##
## testsuite.log was created. ##
## -------------------------- ##

Please send `src/tests/testsuite.log' and all information you think might help:

   To: <https://github.com/firewalld/firewalld>
   Subject: [firewalld 1.0.999] testsuite: 90 91 92 93 113 114 117 121 123 126 128 131 132 133 164 167 193 194 195 196 216 217 219 222 224 228 230 233 234 235 266 269 failed

You may investigate any problem if you feel able to do so, in which
case the test suite provides a good starting point.  Its output may
be found below `src/tests/testsuite.dir'.
```

I think this all also needs some documentation and spec file adjustments?

erig0:
> > NOTE: I can't find a way to make autotest NOT generate the following
> > directories when run.
> 
> Do you mean `testsuite.dir` and `testsuite.log` (the latter is a
> regular file)? If so, why? Do you not want the testsuite to produce
> logs?

Yes. The bummer is they are created in `/usr/share/firewalld` instead of somewhere like `/tmp`.

> One glitch I noticed is that the installed test suite preserves the
> source tree assumptions, e.g. in case of failures, the report still
> references `src/tests` and `sed` tries to access `../../config.log`:

The only way to solve these is to make a "standalone" testsuite that doesn't rely on `atconfig` or `atlocal`. That also means dropping the 4 python tests. The major benefit is the testsuite could be run from any location, e.g. `/tmp`.

> ```
> ## ------------- ##
> ## Test results. ##
> ## ------------- ##
> 
> ERROR: All 38 tests were run,
> 32 failed unexpectedly.
> sed: can't read ../../config.log: No such file or directory
> ## -------------------------- ##
> ## testsuite.log was created. ##
> ## -------------------------- ##
> 
> Please send `src/tests/testsuite.log' and all information you think might help:
> 
>    To: <https://github.com/firewalld/firewalld>
>    Subject: [firewalld 1.0.999] testsuite: 90 91 92 93 113 114 117 121 123 126 128 131 132 133 164 167 193 194 195 196 216 217 219 222 224 228 230 233 234 235 266 269 failed
> 
> You may investigate any problem if you feel able to do so, in which
> case the test suite provides a good starting point.  Its output may
> be found below `src/tests/testsuite.dir'.
> ```
> 
> I think this all also needs some documentation and spec file adjustments?

Right. I was thinking about dropping a README next to the testsuite script. That sound good?

erig0:
@mbiebl, heads up in case you want to package a `firewalld-test` subpackage for this.

stepnem:
@erig0, thank you for addressing the issues, it appears to work very well now (I briefly tested both the `make install` and the RPM use cases).

I think it would be very useful if the installed README also included the tip for listing all keywords, similar to the project README, but that can be done on top of this, so I'm not going to block the merge any longer.

stepnem (approved):

PR #815, opened 2021-07-01
Add policy completion for zsh

cathaysia:


erig0:
Thanks. LGTM. Enabled automerge.

erig0 (approved):

PR #821, opened 2021-07-16
[WIP] docs(zones,zone): describe intra zone forwarding

yrro:
Describe the Intra Zone Forwarding setting and its effect.

### DIFF
--- a/doc/xml/firewalld.zone.xml
+++ b/doc/xml/firewalld.zone.xml
@@ -61,7 +61,7 @@
 
     <para>
       A firewalld zone configuration file contains the information for a zone.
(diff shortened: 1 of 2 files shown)


yrro:
While modifying `firewalld.zones.xml` it occurred to me that a description of the other properties of a zone, such as the target and ICMP block inversion setting would be nice to have. Tell me if you want me to send separate PRs to add these, or whether I should expand the scope of this PR and add additional commits.

erig0:
> While modifying `firewalld.zones.xml` it occurred to me that a description of the other properties of a zone, such as the target and ICMP block inversion setting would be nice to have. Tell me if you want me to send separate PRs to add these, or whether I should expand the scope of this PR and add additional commits.

Separate PRs please! Thanks for improving the docs!

erig0 (approved):

PR #822, opened 2021-07-20
Switch PyQt5 to PySide2

cathaysia:
I had notice that some distrubitions has no packing PyQt5 (such as Fedora), but has PySide2 instead, so firewalld has no ability to show a applet for KDE, switch PyQt to PySide2 may can improve this stitution.

This code has test passed. All code need be change just replace `from PyQt5 import QtGui, QtCore, QtWidgets` to `from PySide2 import QtGui, QtCore, QtWidgets`


cathaysia:
Sorry, this is a error PR, I'd recreate it sooner

PR #823, opened 2021-07-20
feat: Switch PyQt5 to PySide2

cathaysia:
I had notice that some distrubitions has no packing PyQt5 (such as Fedora), but has PySide2 instead, so firewalld has no ability to show a applet for KDE, switch PyQt to PySide2 may can improve this stitution.

This code has test passed. All code need be change just replace `from PyQt5 import QtGui, QtCore, QtWidgets` to `from PySide2 import QtGui, QtCore, QtWidgets`

### DIFF
--- a/src/firewall-applet.in
+++ b/src/firewall-applet.in
@@ -21,7 +21,7 @@
 #
 
 import sys
-from PyQt5 import QtGui, QtCore, QtWidgets
+from PySide2 import QtGui, QtCore, QtWidgets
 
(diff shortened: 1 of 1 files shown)


cathaysia:
For Fedora 34, all dependences about this component should just be `python3-dbus` and `python3-pyside2`. No pyqt5 was needed.

cathaysia:
Sorry for this mistake, this is a useless pr.

PR #824, opened 2021-07-20
docs(zones): describe ICMP block inversion

yrro:
Add a description of ICMP block inversion to firewalld.zones(5)


erig0 (approved):
Thanks Sam!

PR #825, opened 2021-07-20
feat(firewalld): drop linux capabilities

erig0:


erig0:
@mbiebl. FYI, in case you want to add this optional dependency in Debian.

mbiebl:
Wouldn't it be easier to just use systemd's builtin mechanisms for dropping capabilities?
This would mean no need for custom code and no 3rd party dependency (on python-capng)

erig0:
> Wouldn't it be easier to just use systemd's builtin mechanisms for dropping capabilities?

Yes. I think dropping in the daemon is just as easy.

Two reasons to drop in the daemon vs systemd service definition:
1. testsuite support and coverage
2. support other init systems

> This would mean no need for custom code and no 3rd party dependency (on python-capng)

The python dependency is optional. There is no harm in downstream changes to drop capabilities in systemd and avoid the packaging dependency on python-capng.

mbiebl:
Well, changing it downstream has the downside that I need to keep the list of caps up-to-date in the .service file.

Can you give me any pointers which parts of the testsuite cover/require dropping of caps?

I also noticed, that event after installing python3-cap-ng (0.7.9) on Debian, I don't get 
```
            log.info(log.INFO1, "Dropped Linux capabilities to NET_ADMIN, NET_RAW, SYS_MODULE.")
```
in my journal log. What am I doing wrong?
```
# grep INFO1 /var/log/firewalld
<empty>
```

mbiebl:
btw, would it be possible to log this to the journal (when systemd is used)? I'm so used to the journal by now, that I find it cumbersome having to check external log files.

erig0:
> Can you give me any pointers which parts of the testsuite cover/require dropping of caps?

There are no tests specific to dropping capabilities. I meant that the entire testsuite runs with reduced capabilities. Which verifies that running with reduced capabilities actually works.

> I also noticed, that event after installing python3-cap-ng (0.7.9) on Debian, I don't get
> 
> ```
>             log.info(log.INFO1, "Dropped Linux capabilities to NET_ADMIN, NET_RAW, SYS_MODULE.")
> ```
> 
> in my journal log. What am I doing wrong?

It's possible SELinux is blocking it. That was the case in Fedora.

https://bugzilla.redhat.com/show_bug.cgi?id=1985494

erig0:
> btw, would it be possible to log this to the journal (when systemd is used)? I'm so used to the journal by now, that I find it cumbersome having to check external log files.

I don't know how you'd do that. Would firewalld have to detect it's running with reduced capabilities?

PR #830, opened 2021-07-28
services: ssdp includes

marcosfrm:
About...
https://github.com/firewalld/firewalld/blob/c11510e065fdbaeb8ff20f32e2628bdd70d30dd7/config/services/upnp-client.xml

can a service only include other?

### DIFF
--- a/config/services/ssdp.xml
+++ b/config/services/ssdp.xml
@@ -3,5 +3,5 @@
   <short>Simple Service Discovery Protocol (SSDP)</short>
   <description>The Simple Service Discovery Protocol (SSDP) is a network protocol based on the Internet protocol suite for advertisement and discovery of network services and presence information.</description>
   <port protocol="udp" port="1900"/>
-  <destination ipv4="239.255.255.250" ipv6="FF02::C"/>
+  <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
 </service>
--- a/config/services/minidlna.xml
+++ b/config/services/minidlna.xml
@@ -3,5 +3,5 @@
   <short>MiniDLNA</short>
   <description>MiniDLNA is a simple media server software with the aim to be fully compliant with DLNA/UPNP-AV clients. Enable this service if you run minidlna service.</description>
   <port protocol="tcp" port="8200"/>
-  <port protocol="udp" port="1900"/>
+  <include service="ssdp"/>
 </service>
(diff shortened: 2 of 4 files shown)


config/services/ssdp.xml
```diff
@@ -3,5 +3,5 @@
   <short>Simple Service Discovery Protocol (SSDP)</short>
   <description>The Simple Service Discovery Protocol (SSDP) is a network protocol based on the Internet protocol suite for advertisement and discovery of network services and presence information.</description>
   <port protocol="udp" port="1900"/>
-  <destination ipv4="239.255.255.250" ipv6="FF02::C"/>
+  <destination ipv4="239.255.255.250" ipv6="ff02::c"/>
```
erig0, line 6: Was this change intentional? Seems unrelated. Please remove.

erig0 (changes requested):
> can a service only include other?

I'm not sure what you're asking. Services may include other services.

marcosfrm:
> > can a service only include other?
> 
> I'm not sure what you're asking. Services may include other services.

I mean "UPnP Client" from a firewall POV is "UPnP discovery protocol", i.e. SSDP... hence can `upnp-client.xml` _only_ include ssdp?

marcosfrm:
https://github.com/firewalld/firewalld/blob/c11510e065fdbaeb8ff20f32e2628bdd70d30dd7/config/services/upnp-client.xml#L5

btw this is wrong, because the source port may be random for inbound traffic (M-SEARCH, NOTIFY).

erig0:
> > > can a service only include other?
> > 
> > 
> > I'm not sure what you're asking. Services may include other services.
> 
> I mean "UPnP Client" from a firewall POV is "UPnP discovery protocol", i.e. SSDP... hence can `upnp-client.xml` _only_ include ssdp?

Sure. That effectively makes it an alias.

erig0:
> https://github.com/firewalld/firewalld/blob/c11510e065fdbaeb8ff20f32e2628bdd70d30dd7/config/services/upnp-client.xml#L5
> 
> btw this is wrong, because the source port may be random for inbound traffic (M-SEARCH, NOTIFY).

Correct. It was wrong. `upnp-client` probably never worked. Having it include `ssdp` seems like the correct thing to do.

erig0 (approved):
Thanks!

PR #833, opened 2021-08-04
improvement(service): LLMNR: split into llmnr-tcp and llmnr-udp

marcosfrm:
TCP implementation cannot work with destination addresses.


erig0 (changes requested):
You need to add the new service: `llmnr-tcp`, `llmnr-udp` to `config/Makefiles.am` and `po/POTFILES.in`. Otherwise LGTM.

erig0 (approved):
ACK. Thanks.

PR #880, opened 2021-11-05
test(functions): network-manager - manage test dev

dbungert:
On some configurations, dummy devices may not be managed by
network-manager, even with specification of the config data.
Add dummy to the devices to be managed, along with other devices types
I've found in the test suites.

With this rhbz1928860.at is able to pass on Ubuntu.
From https://bugs.launchpad.net/ubuntu/+source/firewalld/+bug/1945596


erig0:
Thanks. This change was merged via #938.

PR #888, opened 2021-11-17
feat(build): distribute an OCI container image

erig0:
See commit message and README.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+FROM registry.access.redhat.com/ubi8/ubi
+
+LABEL description="Firewalld daemon and command line interface"
+MAINTAINER Firewalld Maintainers <firewalld-users@lists.fedorahosted.org>
+
+# firewalld build/runtime dependencies
+#
+RUN dnf -y install automake autoconf make intltool \
+                   docbook-style-xsl python3-nftables \
+                   python3-gobject-base libxslt glib2-devel
+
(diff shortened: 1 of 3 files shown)


erig0:
@mheon, if you have time can you take a quick look? Is this a sensible way to do this?

mheon:
I wouldn't consider it necessary to dump the image as a tar file - I'd be content to stop at a built image and print the name and SHA of the built image, then let the user proceed from there. Main reasoning that that passing around archives isn't a common way of handling images - IMO, users pushing to a registry they control would be a more typical next step.

Beyond that, it looks fine.

erig0:
> I wouldn't consider it necessary to dump the image as a tar file - I'd be content to stop at a built image and print the name and SHA of the built image, then let the user proceed from there.

I was intending to post the built images alongside the released source tarball, i.e. `make dist`.

> Main reasoning that that passing around archives isn't a common way of handling images - IMO, users pushing to a registry they control would be a more typical next step.

So as part of the firewalld release process, tagging and pushing an image to quay.io or other would be preferable?

mheon:
Ah, sorry for misunderstanding - thought this was mainly intended for development.

Tagging + pushing to a place like `quay.io` is probably a better workflow, yes. Distributing a tarball of the image alongside it (e.g. in the Github release) is not a bad idea, but I'd view that as a backup/archival solution in case the registry has issues or purges older images after a large number of images are hosted.

Issue #797, opened 2021-05-10
policy: zone drifting possible if using two policies and ingress zones are source based


erig0:
**What happened**:
Policies match the source interface and or the source address. The matching occurs before zone dispatch. As such it's possible fro policyA to match the source address of ingress zoneA and policyB to match the source address/interface of ingress zoneB. This is very similar to zone drifting.

**What you expected to happen**:
Only the policies related to the ingress zone should be considered.

**Solution**:
This can be solved by moving the policy dispatch to _inside_ the zone's chain. It also means the policy dispatch won't have to match the ingress interface. The positive value policies should occur just before the catch-all reject/deny/accept in the zone.

**Unsolved problem**:
The egress/outgoing zones may still drift if the above strategy is used. Need to find a way to avoid that.

PR #892, opened 2021-12-01
fix(policies): Splitting interfaces with wildcards

ekhandigi:
When creating a policy firewalld does not handle wildcard interfaces
(i.e. tun+). These interfaces need to be dealt with differently due to
the fact nftables cannot parse them in a 'set' (i.e. it will not
register a wildcard if given {"tun*", "net1"}).

This have been overcome through identifying the interfaces with
wildcards and splitting them out.

Before:
ifname { "eth0", "eth1" "tun+" } oifname { "eth1", "eth2" } jump filter_FWD_policy_test

After:
ifname "tun*" oifname { "eth1", "eth2" } jump filter_FWD_policy_test
ifname { "eth0", "eth1" } oifname { "eth1", "eth2" } jump filter_FWD_policy_test


erig0:
Hi, thanks for the fix and the associated test case!

I have not reviewed this yet, but I'm also making changes in this area to address #797. Assuming review is okay then I'll take your change before mine for #797. It will be good to have your test case. :)

AddisonG:
Looks good to me. Waiting on this fix.

erig0:
Hi, sorry for the delay. I got busy and this fell off my radar. Anyways, I rebased it and simplified the code a bit. Just waiting on CI before I merge.

Thanks!

erig0 (approved):

PR #895, opened 2021-12-07
Add HTTP/3 service

jorti:
Expand the https service definition to include HTTP/3 (443/udp)


erig0:
I'd rather see this as a new service called `http3`. There are many people still running http-1.1 servers so opening UDP does not make sense.

jorti:
> I'd rather see this as a new service called `http3`.

Makes sense. I've updated the PR to create a new http3 service.

erig0 (approved):
lgtm. Thanks!
Waiting on CI before hitting merge.

erig0 (approved):

PR #915, opened 2022-02-16
docs(policy): add missing "nflog" documentation for rule

darxriggs:
Fix minor syntax error for `log` element too.

source of information: `firewalld.richlanguage.xml`


erig0 (approved):
Thanks for spending time on the docs! It really helps the project.

PR #922, opened 2022-03-04
fix(build): oci: use centos:stream8

erig0:


thkukuk:
If you run your own dbus daemon inside the container, how should other tools like e.g. podman and docker communicate with it?

erig0:
> If you run your own dbus daemon inside the container, how should other tools like e.g. podman and docker communicate with it?

They don't. This is the "less bad" option.

If you don't spawn a dbus-daemon inside the container then you have to provide a dbus policy and maybe also a polkit policy. Once you do that you might as well install in from you distribution instead.

So we either have a self contained container, or we force the user to provide a dbus policy to use the container. The former is o much better option.

PR #942, opened 2022-04-08
docs(firewall-cmd): fix typo in set-target option

Cycatz:
The value type should be target rather than zone


erig0 (approved):
lgtm. Thanks!

Issue #414, opened 2018-10-26
systemd unit file: possible dependency loop


mohag:
I'm not sure if this is a bug in firewalld. The default `firewalld.service` file which gets used in the Debian/Ubuntu package end up with this dependency loops involving `firewalld.service`. (I have my theories, but I'm not sure why that is the case)

On stopping:
```
Oct 26 09:43:51 hostname systemd[1]: firewalld.service: Found ordering cycle on sysinit.target/stop
Oct 26 09:43:51 hostname systemd[1]: firewalld.service: Found dependency on cloud-init.service/stop
Oct 26 09:43:51 hostname systemd[1]: firewalld.service: Found dependency on networking.service/stop
Oct 26 09:43:51 hostname systemd[1]: firewalld.service: Found dependency on network-pre.target/stop
Oct 26 09:43:51 hostname systemd[1]: firewalld.service: Found dependency on firewalld.service/stop
Oct 26 09:43:51 hostname systemd[1]: firewalld.service: Job sysinit.target/stop deleted to break ordering cycle starting with firewalld.service/stop
```

On starting:
```
Oct 26 09:44:34 hostname kernel: [   14.321552] systemd[1]: sysinit.target: Found ordering cycle on cloud-init.service/start
Oct 26 09:44:34 hostname kernel: [   14.327144] systemd[1]: sysinit.target: Found dependency on networking.service/start
Oct 26 09:44:34 hostname kernel: [   14.332163] systemd[1]: sysinit.target: Found dependency on network-pre.target/start
Oct 26 09:44:34 hostname kernel: [   14.336991] systemd[1]: sysinit.target: Found dependency on firewalld.service/stop
Oct 26 09:44:34 hostname kernel: [   14.341466] systemd[1]: sysinit.target: Found dependency on basic.target/start
Oct 26 09:44:34 hostname kernel: [   14.345469] systemd[1]: sysinit.target: Found dependency on sockets.target/start
```

Fedora (and likely as a result RHEL and CentOS) have their [own unit file](https://src.fedoraproject.org/rpms/firewalld/blob/56e929d1f4e6841dd961343d0053abd68bd911a2/f/firewalld.service), which does not exhibit this problem.

Replacing the unit file with this version (A mix of the default on and the Fedora one), seems to resolve the problem: (Mainly non-forking and keeping some of the dependencies)
```
[Unit]
Description=Firewall dynamic change handling daemon
After=syslog.target
After=dbus.service
After=polkit.service
[...]

PR #946, opened 2022-04-18
fix(systemd): order Before network.target

erig0:
This avoids a dependency loop with cloud-init and other early boot
services.

The alternative is to also become an early boot service by using
`DefaultDependencies=no`, but that would mean changes inside of firewalld
to handle dbus not being available.

Fixes: #414

### DIFF
--- a/config/firewalld.service.in
+++ b/config/firewalld.service.in
@@ -1,7 +1,6 @@
 [Unit]
 Description=firewalld - dynamic firewall daemon
-Before=network-pre.target
-Wants=network-pre.target
+Before=network.target
 After=dbus.service
 After=polkit.service
 Conflicts=iptables.service ip6tables.service ebtables.service ipset.service nftables.service


mbiebl:
The commit message is misleading.

Ordering before network.target does *not* fix the dependency loop.

Dropping Wants/Before network-pre.target does.

I'm actually not sure what the Before=network.target is supposed to achieve?

mbiebl:
If you want to setup the firewall rules *before* the network is configured, you need to use network-pre.target.

`Before=network.target` does *not* achieve that, so you might just as well drop it.

erig0:
> The commit message is misleading.
> 
> Ordering before network.target does _not_ fix the dependency loop.

Right. But ordering it later does.

> Dropping Wants/Before network-pre.target does.

Agree.

> I'm actually not sure what the Before=network.target is supposed to achieve?

Doesn't it still order firewalld before things that want `After=network-online.target` ?

PR #948, opened 2022-04-18
feat(service): add netdata service

jcpunk:
Add a service definition for netdata.


erig0:
This looks like it's only exposing the dashboard. AFAICS, there is management and other access. Would it be better to call this `netdata-dashboard` ?

jcpunk:
That works for me, I've updated the patch.

erig0 (approved):
lgtm. Thanks for the updates.

Issue #940, opened 2022-04-08
duplicate double quote in log prefix 


MorphBonehunter:
**What happened**:
Today i realized that some log prefixes in nftables generated rules have duplicate double quotes.

```
nft list ruleset |grep "log prefix"
                log prefix ""filter_IN_public_DROP: ""
                log prefix ""filter_FWD_public_DROP: ""
```

**What you expected to happen**:
```
nft list ruleset |grep "log prefix"
                log prefix "filter_IN_public_DROP: "
                log prefix "filter_FWD_public_DROP: "
```

**How to reproduce it (as minimally and precisely as possible)**:
Enable `LogDenied=all` in firewalld.conf and reload firewalld.

**Anything else we need to know?**:
I guess the Problem is in [nftables.py#L913](https://github.com/firewalld/firewalld/blob/11f997f04fd62c1852f6c381241c089d59ab72bd/src/firewall/core/nftables.py#L913) because of the escaped double quotes.
Another occurrence is in [nftables.py#L1540](https://github.com/firewalld/firewalld/blob/11f997f04fd62c1852f6c381241c089d59ab72bd/src/firewall/core/nftables.py#L1540).

I removed the escaped quotes and it works like expected but i doesn't know if this may have another impact.

**Environment**:
- Firewalld Version (if Fedora based `dnf info firewalld` or commit hash if developing from git `git log -n1 --format=format:"%H"`):
    - 1.0.3
- Firewalld Backend (`cat /etc/firewalld/firewalld.conf | grep FirewallBackend`): 
    - `FirewallBackend=nftables`
- OS (e.g: `cat /etc/os-release`):
    - `NAME="Arch Linux"`

PR #952, opened 2022-04-22
fix: Remove duplicate quotation from log prefixes

TorontoMedia:
Closes: #940


erig0:
Hi, I recall there being issues when `IndividualCalls=yes`. Can you add test coverage for that? iptables for example, will pass the strings via CLI instead of using a temporary file and `iptables-save`. So the quoting can be tricky.

TorontoMedia:
> Hi, I recall there being issues when `IndividualCalls=yes`. Can you add test coverage for that? iptables for example, will pass the strings via CLI instead of using a temporary file and `iptables-save`. So the quoting can be tricky.
> 
> Edit: I just realized this is fixing a particular instance of over quotation. Maybe it's okay after all. Perhaps the quotes are left over from when firewalld used to call the `nft` CLI.

I also believe that it's just leftover from the previous implementation but I've found a few more issues with duplicate quotation and have added the test for `IndividualCalls`. 

At some point we should consolidate the testing of all logging features.

erig0:
> At some point we should consolidate the testing of all logging features.

I've been adding new coverage in `src/tests/features`. But for fix for specific bugs I'm still dropping them in `src/tests/regressions`.

You could also tag the test case `AT_KEYWORDs(log)`. Then `make check TESTSUITEFLAGS="-k log"` will run it.

erig0 (approved):
This looks good. Thanks! Merging.

PR #954, opened 2022-04-28
Reorganize Kubernete's services

krouma:
I have a few improvements to default service definitions:

- Service `kube-api` should be named `kubelet`, having that and `kube-apiserver` is quite confusing
- Although `kubelet-worker` isn't that bad, it is usually referred to as `kube-worker` or Kubernetes worker
- `kubelet` is not part of control plane, although I get that control plane nodes are nearly always workers, too
- `kubelet` also has read-only port 10255
- There are new secure variants for `kube-controller-manager` and `kube-scheduler`. The non-secure variants are even disabled by default since Kubernetes 1.22

With these changes, configuring firewall for running Kubernetes nodes is simplified and (from my point of view) more clear to follow, since you only have to enable `kube-control-plane-secure` and `kube-worker` for control plane nodes and only the latter for worker nodes.


config/Makefile.am
```diff
@@ -191,12 +191,17 @@ CONFIG_FILES = \
 	services/kpasswd.xml \
 	services/kprop.xml \
 	services/kshell.xml \
-	services/kube-api.xml \
```
erig0, line 194: Dropping a service definition is a breaking change.

Consider making this an alias to `kubelet` via the `include` feature.

config/Makefile.am
```diff
@@ -191,12 +191,17 @@ CONFIG_FILES = \
...
+	services/kubelet.xml \
+	services/kubelet-readonly.xml \
 	services/kube-apiserver.xml \
 	services/kube-control-plane.xml \
+	services/kube-control-plane-secure.xml \
 	services/kube-controller-manager.xml \
+	services/kube-controller-manager-secure.xml \
+	services/kube-nodeport-services.xml \
 	services/kube-scheduler.xml \
-	services/kubelet-worker.xml \
```
erig0, line 199: Same as above. This is a breaking change.

I think you need to alias to `kube-worker`.

erig0:
@dcbw, if you have time can you review these changes?

config/services/kubelet-readonly.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Kubernetes Kubelet read</short>
+  <description>The kubelet API is used to communicate between kube-scheduler and the node.</description>
+  <port protocol="tcp" port="10255"/>
```
erig0, line 5: Does it really makes sense to separate `kubelet` and `kubelet-readonly`? Would a user ever want to keep the readonly port closed? If no, then please combine this with `kubelet`.

config/services/kube-control-plane.xml
```diff
@@ -7,5 +7,4 @@
   <include service="kube-apiserver" />
   <include service="kube-controller-manager" />
   <include service="kube-scheduler" />
-  <include service="kube-api" />
```
erig0, line 10: Dropping this may surprise users (kubelet may no longer be exposed). I think it's okay since the service was intended to be for the _control plane_. The presence of this here should be considered a bug.

erig0 (changes requested):
Thanks for cleaning these up. I have some suggestions. I'd like someone better versed in k8s to take a look as well.

config/services/kubelet-readonly.xml
krouma: I was thinking more of enabling only the readonly port in a certian zone but I don't really know how much this is used.

erig0: Okay. We'll leave it as-is since it was the existing behavior. In the future we could `include` the read-only service from the `kubelet` service.

config/Makefile.am
krouma: Done

config/Makefile.am
krouma: Done, it is now with the rest of kubelet services few lines above.

erig0 (approved):
This looks great. Thanks @krouma !

erig0:
> This looks great. Thanks @krouma ! I'll merge after CI.

My bad for the delay. I got distracted why waiting on CI. ;)

PR #969, opened 2022-06-13
test: rich rule parsing bottleneck increase timeout

afreof:
On some low performance devices (ARM Cortex A35) this test needs much
longer.


erig0:
Thanks! Merging.

If this continues to be a problem on low end devices then I think we'll just completely remove the timeout.

PR #978, opened 2022-06-25
add protocol test in policy.at

kircherlike:
add protocol test in policy.at


erig0:
> add protocol test in policy.at

There are already protocol tests in `src/tests/features/protocols.at`. These seem redundant.

erig0:
Closing because there is already test coverage for this area.

PR #993, opened 2022-07-17
Add Nebula service

Fale:
[Nebula](https://github.com/slackhq/nebula) is a Noise-based VPN that uses port 4242/UDP by default. The port needs to be open on all discoverable hosts, which are called `lighthouse` in Nebula.


erig0:
Hi, this looks good to merge.

I just want to confirm that you want firewalld to ship these service definitions. Some projects prefer to ship the definitions as part of their package. That may make it easier to manage version differences, avoid package conflicts, etc. See https://github.com/firewalld/firewalld/pull/1002 (and https://github.com/firewalld/firewalld/pull/552) for one such case.

I'm more than happy to take these though. Just let me know.

Fale:
I think we can proceed with the inclusion within Firewalld.
The upstream project is not shipping the firewalld configuration nor seems interested in doing so.
Also, due to the nature of nebula, it would not make much sense to ship firewalld configuration.

erig0 (approved):

PR #997, opened 2022-07-19
improvement(github): translate templates to yaml

TorontoMedia:
Convert issue templates from markdown to yaml. This will require adopting two labels `bug` and `triage`.


erig0:
I created the labels. Do you feel we need to automatically tag new issues with "bug"? Maybe we can just use "triage" and only tag as "bug" if someone has confirmed it is: a) still an issue and b) actually a bug.

TorontoMedia:
> I created the labels. Do you feel we need to automatically tag new issues with "bug"? Maybe we can just use "triage" and someone manually tags as "bug" if it is: a) still an issue and b) actually a bug. Rationale is that a lot of reports fall into those two categories.

I agree. To make the templates more ambiguous, I removed the titles and added a config file that allows blank templates.

TorontoMedia:
I also made `What happened` in `bug_report.yml` and `What would you like to be added` in `enhancement.yml` a requirement.

erig0 (approved):
Thanks @TorontoMedia!

erig0:
> 



> @erig0 Thank you for merging my pull request. In config.yml, I believe we should capitalize the mailing list names to improve the appearance of the template list. Are you open to another pull request that does this?

Sure.

Issue #1000, opened 2022-07-22
Firewalld fails on systems with non-modular kernels


alexkingnz:
**What happened**:
On a VPS with a monolithic kernel, after loading firewalld I see the following traceback in the journal:
```
Jul 21 03:00:46 test.domu.build.rimuhosting.com firewalld[26579]: Traceback (most recent call last):
                                                                    File "/usr/lib/python3.6/site-packages/firewall/server/decorators.py", line 67, in dbus_handle_exceptions
                                                                      return func(*args, **kwargs)
                                                                    File "/usr/lib/python3.6/site-packages/firewall/server/firewalld.py", line 335, in completeReload
                                                                      self.fw.reload(True)
                                                                    File "/usr/lib/python3.6/site-packages/firewall/core/fw.py", line 1092, in reload
                                                                      raise start_exception
                                                                    File "/usr/lib/python3.6/site-packages/firewall/core/fw.py", line 1010, in reload
                                                                      self._start(reload=True, complete_reload=stop)
                                                                    File "/usr/lib/python3.6/site-packages/firewall/core/fw.py", line 444, in _start
                                                                      self.modules_backend.unload_firewall_modules()
                                                                    File "/usr/lib/python3.6/site-packages/firewall/core/modules.py", line 99, in unload_firewall_modules
                                                                      for module in self.get_firewall_modules():
                                                                    File "/usr/lib/python3.6/site-packages/firewall/core/modules.py", line 79, in get_firewall_modules
                                                                      (mods2, deps) = self.loaded_modules()
                                                                    File "/usr/lib/python3.6/site-packages/firewall/core/modules.py", line 43, in loaded_modules
                                                                      with open("/proc/modules", "r") as f:
                                                                  FileNotFoundError: [Errno 2] No such file or directory: '/proc/modules'
```
`firewall-cmd --state` shows `failed`
No traffic is able to be received on port 80, while I believe configuration is in place to allow that, if only firewalld were functional.

**What you expected to happen**:
No traceback, A state other than "failed" in response to `firewall-cmd --state`  A functional firewall.

**How to reproduce it (as minimally and precisely as possible)**:
Reboot with a non-modular kernel.  Or test on a hosting provider who use that by default.

**Environment**:
```
# dnf info firewalld
Last metadata expiration check: 3:18:53 ago on Thu 21 Jul 2022 22:18:50 UTC.
Installed Packages
Name         : firewalld
Version      : 0.9.3
Release      : 13.el8
[...]

PR #1001, opened 2022-07-22
Fix for firewalld running on systems with non-modular kernels

alexkingnz:
Don't error if /proc/modules is missing.
This is a fix for #1000 
Not extensively tested, but works well for me.

I'm not aware of why a kernel module would need to be unloaded.   This can't happen on non-modular kernels.  Without being aware of the reason why firewalld attempts to unload kernel modules, I can't comment on what effects the non-unloading of them will have.

### DIFF
--- a/src/firewall/core/modules.py
+++ b/src/firewall/core/modules.py
@@ -40,17 +40,20 @@ class modules(object):
         """ get all loaded kernel modules and their dependencies """
         mods = [ ]
         deps = { }
-        with open("/proc/modules", "r") as f:
-            for line in f:
-                if not line:
-                    break
-                line = line.strip()
-                splits = line.split()
-                mods.append(splits[0])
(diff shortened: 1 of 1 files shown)


erig0:
> I'm not aware of why a kernel module would need to be unloaded. This can't happen on non-modular kernels. Without being aware of the reason why firewalld attempts to unload kernel modules, I can't comment on what effects the non-unloading of them will have.

Probably historical reason that are lost to me. New firewalld versions do _not_ unload modules by default.

See `CleanupModulesOnExit` in `firewalld.conf`:
https://github.com/firewalld/firewalld/blob/9b27e51e921428d72738c166ad51fb0e47b9a497/config/firewalld.conf#L14-L19

erig0 (approved):
Thanks!

PR #1010, opened 2022-08-28
fix(bash): fix ipset commands autocompletion

freedge:
Fix typo that prevent autocompletion of --ipset commands.


erig0 (approved):
lgtm. Thanks!

PR #1011, opened 2022-08-28
fix(firewalld): prevent exception checking ipsets

freedge:
in the case of --remove-entries-from-file,
check_for_overlapping_entries can be called with no entry in input,
which fails with an exception.

Fixes: rhbz2121985


erig0 (approved):
Thanks. I pushed a new test case on top of your commit. I'll merge after CI completes.

PR #1012, opened 2022-08-28
feat(service): add Ceph Prometheus exporter

freedge:
Add Prometheus module listening on default port 9283.
(https://docs.ceph.com/en/quincy/mgr/prometheus/)


erig0 (approved):
lgtm. Thanks.

PR #1031, opened 2022-10-05
Llmnr client

yrro:


config/services/llmnr.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <service>
   <short>LLMNR</short>
-  <description>Link-Local Multicast Name Resolution (LLMNR) allows both IPv4 and IPv6 hosts to perform name resolution for hosts on the same local link.</description>
+  <description>Link-Local Multicast Name Resolution (LLMNR) allows both IPv4 and IPv6 hosts to perform name resolution for hosts on the same local link. This service matches incoming queries; it will allow this host to be resolved by other hosts.</description>
   <include service="llmnr-tcp"/>
   <include service="llmnr-udp"/>
```
erig0, line 6: What do you think about allowing `llmnr-client` here? I think a user enabling `llmnr` will also want to hear the replies.

Note that `ws-discovery` does NOT include `ws-discovery-client`, but maybe it should. I wonder how @gbsf and @marcosfrm feel about this.

marcosfrm: We should avoid `source-port` as much as possible. rfc4795 says implementations are free to discard unsolicited responses...

erig0: @marcosfrm, fair enough. Let's leave out the `-client` variants then.

gbsf: The thing is, while the `-client` services are a kludge, they are basically required to use on desktop machines. Until we can get conntrackd working, we have to rely on an implementation detail that Linux applications in general use the same request port number as the response, instead of a random one. The multicast-to-unicast dance leaves no other option, otherwise we end up discarding _solicited_ responses too…
If they are not enabled by default alongside the main service, I feel that at least the description of those should mention that `-client` is currently needed to receive replies.

erig0: Let me ask it this way.

Would a user _expect_ that enabling the `llmnr` service will allow them to make queries?

marcosfrm: I stand against any rule with `source-port` being enabled indirectly.

config/services/llmnr-client.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>LLMNR Client</short>
+  <description>LLMNR allows both IPv4 and IPv6 hosts to perform name resolution for hosts on the same local link. This service matches incoming responses.</description>
+  <description>Link-Local Multicast Name Resolution (LLMNR) allows both IPv4 and IPv6 hosts to perform name resolution for hosts on the same local link. This service matches incoming responses; it will allow this host to be resolved by other hosts. This service matches by source-port; if added to an untrusted zone, an attacker within that zone could bypass the firewall. Only add this service to reasonably-safe zones.</description>
```
erig0, line 5: This is a duplicate description.

erig0 (changes requested):
One change needed, otherwise lgtm.

erig0 (approved):
I cleaned up this commit and split it into two. Will merge after CI completes.

yrro:
Ah thanks :)

PR #1040, opened 2022-11-03
fix:  correct conditional statement and raise exceptions

TorontoMedia:
Rather than just initiating exceptions, raise them instead, and fix the conditional statement issue that was discovered after raising an exception.


erig0 (approved):
lgtm. Thanks for the code clean up.

Issue #1057, opened 2022-11-29
Remove application specific service files (at least from main package)


tebrevik:
### What would you like to be added

I would like application specific service defintions to be removed. At least from the main package.

### Why is this needed

Right now there is an open jellyfin issue due to #861:   https://github.com/jellyfin/jellyfin/issues/8587

I get that application specific service files has been implemented for a reason and that there might be good arguments for doing it this way. But in my opinion having application specific service files in the main firewalld package is a mistake. This create strong bindings to the applications that I don't think should be there. It should be (and has been) the application that should distribute this service file. What if the application want to add another port or something, should it then need to delay that release until firewalld service files has been updated? And why do the firewalld project want to take on the responsibilty for opening ports that suddenly might not be used by the next version of the application?

If the firewalld project feels that these service files belong then I suggest that they at least should be separated out into individual package files, like firewalld-jellyfin.rpm. Another option would be to gather the application services in firewalld-services.rpm.

erig0:
If Jellyfin has been shipping a service definition for some time, then we can probably safely revert #861. @cmuellner, any thoughts?

There are advantages and disadvantages to including the service definition in firewalld itself. Some packages prefer to ship them by themselves, others prefer to include them in firewalld.

tebrevik:
I'm just a user of both applications, so worst case you both remove your service defintions...

But my highly personal opinion is that it's a risk for a firewall application to carry service defintions that likely will be outdated or mismatched at some point due to changes in the original application.

cmuellner:
My understanding is, that Podman installation is the preferred installation method on Fedora.
In this case no host configuration for firewalld will be installed for Jellyfin.
And the remote management console (cockpit) benefits from named firewalld configs (that was the use case that I intended to improve with the patch).

So as long as Jellyfin documents installation methods for Fedora that don't install firewalld configs, there will be an inconsistent user experience if we revert the support from firewalld.

Anyway, I don't mind if the change gets reverted.

erig0:
> But my highly personal opinion is that it's a risk for a firewall application to carry service defintions that likely will be outdated or mismatched at some point due to changes in the original application.

As you say, this is your opinion. Others prefer to ship it in via firewalld, e.g. #1009.

If the jellyfin maintainers want to see the service definition reverted in firewalld, then I'll revert. I just need an official word.

cmuellner:
[...]

PR #1058, opened 2022-11-29
Revert "feat(service): Add jellyfin service"

cmuellner:
This reverts commit 8c168d9eda5fd035d91422393ffea9b6ebcd2728.

As reported in https://github.com/jellyfin/jellyfin/issues/8587, this causes an issue with RPMs shipped by Jellyfin. This has been reported to firewalld in https://github.com/firewalld/firewalld/issues/1057, where @joshuaboniface stated that the Jellyfin project would like to remove the Jellyfin config from firewalld, which is what this change does.

CC: @erig0

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -192,7 +192,6 @@ CONFIG_FILES = \
 	services/irc.xml \
 	services/iscsi-target.xml \
 	services/isns.xml \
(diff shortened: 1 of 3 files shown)


erig0 (approved):
Thanks!

PR #1065, opened 2022-12-09
fix(tcp-mss-clamp): duplicate rule fragments

TorontoMedia:
Remove duplicate rich rule fragments and assign default value to chain suffix variable.

Also directly return concatenated rule fragments.


erig0 (approved):
Thanks! lgtm.

TorontoMedia:
> Thanks! lgtm.

Thanks for merging. I rearranged the tcp flags to the beginning and forgot to stage the change 😞 I'll fix that in a future commit.

Issue #878, opened 2021-11-05
Cannot reload with firewall-cmd --reload


agross:
**What happened**:

```sh
$ firewall-cmd --reload
Argument 1 does not allow None as a value
```

```
Nov 05 17:10:52 router firewalld[178828]: Traceback (most recent call last):
                                            File "/usr/lib/python3.10/site-packages/firewall/server/decorators.py", line 67, in _impl
                                              return func(*args, **kwargs)
                                            File "/usr/lib/python3.10/site-packages/firewall/server/firewalld.py", line 323, in reload
                                              self.fw.reload()
                                            File "/usr/lib/python3.10/site-packages/firewall/core/fw.py", line 1082, in reload
                                              for interface in nm_get_interfaces_in_zone(zone):
                                            File "/usr/lib/python3.10/site-packages/firewall/core/fw_nm.py", line 184, in nm_get_interfaces_in_zone
                                              if zone == nm_get_zone_of_connection(conn):
                                            File "/usr/lib/python3.10/site-packages/firewall/core/fw_nm.py", line 76, in nm_get_zone_of_connection
                                              con = nm_get_client().get_connection_by_uuid(connection)
                                          TypeError: Argument 1 does not allow None as a value
```

**What you expected to happen**:

Reloading works.

**How to reproduce it (as minimally and precisely as possible)**:

Not sure.

**Anything else we need to know?**:

**Environment**:
- Firewalld Version (if Fedora based `dnf info firewalld` or commit hash if developing from git `git log -n1 --format=format:"%H"`): Name         : firewalld
Version      : 1.0.1
Release      : 2.fc35
- Firewalld Backend (`cat /etc/firewalld/firewalld.conf | grep FirewallBackend`): FirewallBackend=nftables
- OS (e.g: `cat /etc/os-release`): Fedora Linux 35 (Server Edition)
- Others:

stejoo:
I have the exact same issue and my journal shows the same error message.

My first thought was this was due to network connection profiles in NetworkManager not having their `connection.zone` property set. No something I usually set, but for this issue I did all of them: that did not resolve the issue.

Looking further my next guess is `nm_get_client().get_connection_by_uuid(connection)` needs a `connection` to be filled in. I have an interface that is not managed by NetworkManager, but is known to NetworkManager. It's a WireGuard connection to be precise and it's not managed by NetworkManager but by `wg-quick`.

[...]

PR #1070, opened 2023-01-03
fix(fw_nm): skip interfaces without a connection profile

erig0:
Some interfaces may not have an associated connection profile. This is the case for down point-to-point interfaces.

Fixes: #878

### DIFF
--- a/src/firewall/core/fw_nm.py
+++ b/src/firewall/core/fw_nm.py
@@ -193,7 +193,7 @@ def nm_get_connection_of_interface(interface):
     """
     check_nm_imported()
 
-    device = nm_get_client().get_device_by_iface(interface)
+    device = nm_get_client().get_device_by_ip_iface(interface)
     if device is None:
         return None
 


src/tests/integration/gh878.at
```diff
@@ -0,0 +1,21 @@
...
+AT_KEYWORDS(gh878)
+
+NS_CHECK([ip link add dummy0 type dummy])
+
+START_NETWORKMANAGER()
+
+NS_CHECK([ip link show], 0, [ignore])
+echo NS_CMD([nmcli connection delete dummy0]) >> ./cleanup
+NS_CHECK([nmcli connection add type dummy ifname dummy0 con-name dummy0 ipv6.method ignore], 0, [ignore])
+NS_CHECK([nmcli connection up dummy0], 0, [ignore])
```
erig0, line 11: [CI fails](https://github.com/firewalld/firewalld/actions/runs/4001478444/jobs/6867744159#step:15:1259) here trying to get an IP address. Need to disable ipv4.

PR #1092, opened 2023-03-07
fix(fw_nm): use IP interface names for connection lookup

amachronic:
The glue layer between FirewallD and NetworkManager uses what NM calls "IP interface" names to identify interfaces. However, these names can't be passed to NM.Client.get_device_by_iface() since that checks the "interface name", not the IP interface name. Since NM does not provide a method to look up a device by its IP interface, we need to do that manually.

Fixes: #878


erig0 (approved):
lgtm. Thanks! Will merge after CI.

PR #1093, opened 2023-03-10
Correct Foreman related services

ekohl:
This a draft, since I'm not sure what should be the end state. Recently we investigated using the firewalld services and recommending it to users (https://github.com/theforeman/foreman-documentation/pull/1943), but came to the conclusion that we shouldn't because they're wrong.

Foreman itself only runs on HTTP and HTTPS. It is the Foreman Proxy which can talk to additional services such as DNS, DHCP, TFTP and Puppet. However, all those services are optional. If anything, they should be included in the foreman-proxy service.

Architecturally speaking we have:

* Foreman running on HTTP + HTTPS
* Foreman Proxy running on port 8443 for our regular deployments, on port 9090 if it's a Katello installation (something we hope to unify some day)
* On a Foreman Proxy service it is optional to deploy services such as DNS, DHCP, TFTP, Puppetserver and recently MQTT
* On a Katello installation it's possible to deploy QPID (5646 + 5647 mentioned in the Satellite services), but deprecated
* On a Foreman Proxy with content (RH Satellite terminology: Capsule) we run Pulp on HTTP + HTTPS

Now I struggle with opening optional ports by default: it leads to a less secure system.

Satellite also wants to drop the version number, so we can move to Satellite 7 at some point. This has been proposed multiple times, but so far there have been too many issues holding us back.

With this PR I hope to start the discussion on what should be the profile.


erig0:
Hrm, the question is "will removing these ports break users?". The foreman team is best suited to answer that. Assuming the answer is "no" I suggest the following:

- add a `foreman-proxy` service, includes ports 8443 and 9090
- add a `foreman-proxy-optional` service, includes dns, dhcp, etc.

IF you want to retain the existing behavior of the `foreman` service, then you could use the `include` element in the XML to include services `foreman-proxy` and `foreman-proxy-optional`.

As for versioning the services.. you could always keep the version number, e.g. `satellite-6`, but `include` it from an unversioned service `satellite`. Then you could always change the include to point to say `satellite-7`.

ekohl:
> Hrm, the question is "will removing these ports break users?". The foreman team is best suited to answer that.

It depends on the user's setup.

> add a `foreman-proxy` service, includes ports 8443 and 9090

There already is a `foreman-proxy` service, but it only has port 8443 now.

The user can also choose to enable plain text HTTP support, which runs on port 8000. Is there a precedent in firewalld for those "it depends" options?

> add a `foreman-proxy-optional` service, includes dns, dhcp, etc.

So the tricky bit is that each component is opt-in by itself. So it's perfectly valid to run a Foreman Proxy with only DNS. It really depends on the setup. Even enabling the Foreman Proxy DNS module does not imply a DNS server runs on the same machine: it can talk to a remote DNS server using nsupdate or REST API, depending on the provider. More concrete: if you enable the route53 DNS provider then it talks to AWS Route53 and Amazon hosts the actual DNS zones.

As such I'm leaning to recommending users in the manual for specific scenarios to open the service explicitly.

erig0:
> Is there a precedent in firewalld for those "it depends" options?

I think a fair example is services that have TCP and UDP support. They can be split into `foobar-tcp` and `foobar-udp`. Then the `foobar` service includes both. This is because TCP is the norm, but sometimes UDP is also used.

> As such I'm leaning to recommending users in the manual for specific scenarios to open the service explicitly.

Alternatively, you could suggest that users _modify_ the `foreman` service instead of opening a port, e.g. `--zone <zone> --add-port 9090/tcp`.

```
# firewall-cmd --permanent --service foreman --add-port 9090/tcp`
```

Modifying the service has the downside that users won't get updates to the stock service shipped by firewalld. Once the service is modified it gets copied to `/etc/firewalld/services` and overrides the one in `/usr/lib/firewalld/services`.

ekohl:
I've updated this to incorporate the feedback. See the commit message for all the details.

I don't know about backporting to older versions. We're hoping to soon start introduce support for EL9 and this could be a candidate. Then we'll leave EL8 as it is.

config/services/foreman.xml
```diff
@@ -2,11 +2,6 @@
 <service>
   <short>foreman</short>
   <description>Foreman is a complete lifecycle management tool for physical and virtual servers.</description>
-  <include service="dns"/>
   <include service="http"/>
   <include service="https"/>
-  <include service="dhcp"/>
-  <include service="tftp"/>
-  <port protocol="udp" port="68"/>
-  <port protocol="tcp" port="8140"/>
```
erig0, line 11: These two ports are dropped here, but not added to any other service. They're also not mentioned in the commit. Is that expected?

erig0 (commented):
A previous user of the `foreman` service may now need to enable `foreman-proxy` and `foreman-proxy-optional`. Usually we don't "break" users in this way.

Is this acceptable for foreman users? Foreman team is the best to answer that, IMO.

config/services/foreman.xml
ekohl: I have no clue why 68 was mentioned there. It is DHCP client, but that never made sense.
8140 is puppetserver (in firewalld still called puppetmaster).

ekohl:
> Is this acceptable for foreman users? Foreman team is the best to answer that, IMO.

Hard to say. In the documentation we have never pointed to firewalld. It was added in https://github.com/firewalld/firewalld/pull/645 by you.

The thing that got this whole ball rolling was that we were looking at using these definitions and realized they're not really correct.

I still struggle to think about the whole migration, because as long as we support EL8 (which I doubt will be updated) our instructions won't really be any easier.

erig0:
> > Is this acceptable for foreman users? Foreman team is the best to answer that, IMO.
> 
> Hard to say. In the documentation we have never pointed to firewalld. It was added in #645 by you.

lol. That's embarrassing.

AFAICS, I added it trying to consolidate services as part of https://bugzilla.redhat.com/show_bug.cgi?id=1839781.

config/services/foreman.xml
erig0: ACK. looks like 68 came from the original `RH-satellite-6` service.

erig0: Do you think 68 should be added back to `RH-satellite-6`?

jcpunk:
I think `68` came in along with the dhcp definitions.

erig0:
> The thing that got this whole ball rolling was that we were looking at using these definitions and realized they're not really correct.

In that case, lets fix them. I will merge.

erig0 (approved):

ekohl:
I'll discuss this in our team again and try to come up with a plan

erig0:
> I'll discuss this in our team again and try to come up with a plan

ACK. I'll wait on your response then.

ehelms:
This is mostly my not understanding how it works. What does a user do if the policy includes a port that is no longer used by Satellite / Foreman, such as when we drop 8443?

ehelms:
> I'd argue that shipping it ourselves would be more flexible and allows us to keep it correcy for multiple versions

I lean that way as well, given the set of ports can vary not only by Foreman / Satellite version, but also based upon the enabled feature set. If we had a policy in firewalld I think it would need to default the minimal setup based on the ports that we anticipate to be consistent.

erig0:
> ourselves would be more flexible and allows us to keep it correcy for multiple versions

That's cool. So what do we want to do with existing foreman/satellite/etc services that are included with firewalld?

erig0:
> > ourselves would be more flexible and allows us to keep it correcy for multiple versions
> 
> That's cool. So what do we want to do with existing foreman/satellite/etc services that are included with firewalld?

bump

ehelms:
I believe where we have landed is to remove Satellite/Foreman from being included in firewalld. @erig0 is that a change we should open a PR for?

erig0:
> I believe where we have landed is to remove Satellite/Foreman from being included in firewalld. @erig0 is that a change we should open a PR for?

Yeah. It's technically a breaking change.

As such, decide how to propagate that change to distros, e.g. RHEL, is the difficult part.

ehelms:
I think that's OK if we can't propagate this backwards -- it's had ups and downs where it was accurate and where it was not. I'm more keen on fixing this with a forward looking approach.

erig0:
Then open a PR dropping them. I will merge it.

PR #1098, opened 2023-03-16
fix(policy): disallow zone drifting

erig0:
This change addresses #797. Secondly, it unifies the the policy and zone dispatch into a single dispatch chain per netfilter hook, e.g. INPUT. There are also some various clean ups I made along the way.

These changes are meant to be compatible with previous firewalld releases. However, it's a very significant change in the rule layout. As such, I'm debating bumping the version to v2.0.0.

I did make a behavior change to `--set-default-zone`. It now implicitly triggers a `--reload`. This simplifies a lot of the code to handle the default zone. In theory users should rarely notice this because while firewalld is reloading its maintains existing connections. Additionally, changing the default zone is a rare operation.

The major commits are:
- fix(policy): disallow zone drifting
  - actual fix for #797, I do NOT intend to backport this to other releases.
- chore(zone): now dispatched like regular policies
  - unifies policy and zone dispatch

I have tested:
- RHEL-9: podman, libvirt NAT, libvirt routed
- Ubuntu 20.04: docker, libvirt NAT, libvirt routed

It would be great if it could be tested by others.
@TorontoMedia, @jcpunk, @mbiebl, @todoleza


jcpunk:
My quick attempts at `make test-rpm` seems somewhat unhappy.  I'm thinking the `firewalld.spec` in the repo may be out of date...

Can you pop a test package up in copr for Fedora and RHEL9?

erig0:
> My quick attempts at `make test-rpm` seems somewhat unhappy. I'm thinking the `firewalld.spec` in the repo may be out of date...

I don't think I changed anything that would break the rpm build. I tested by building the RPM. Make sure you install all the build dependencies first..

```
# dnf build-dep ./firewalld.spec
# make rpm
```

If its still fails, then ping me on IRC/matrix and I can help.

jcpunk:
My local testing thus far looks pretty good.  My most fancy system loaded the rules in a different order, so I suspect a `2.0` release is a good idea.  I'm not seeing any actual changes in the traffic that is allowed vs denied in the real world, but the rule order change could hit someone else.

erig0:
> My most fancy system loaded the rules in a different order, so I suspect a 2.0 release is a good idea.

Can you elaborate?

erig0:
I'll be on PTO for a week. I'll merge this PR the first week of April unless issues are identified.

PR #1117, opened 2023-05-05
feat(service): define statsrv from RFC 996

jcpunk:
This service comes in handy at odd times...

### DIFF
--- a/config/services/statsrv.xml
+++ b/config/services/statsrv.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Statistics Server</short>
+  <description> Statistics Server protocol provides a way to give statistics about a system to the outside world. Defined in RFC-996</description>
+  <port protocol="tcp" port="133"/>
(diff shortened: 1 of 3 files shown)


config/services/statsrv.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Statistics Server</short>
+  <description> Statistics Server protocol provides a way to give statistics about a system to the outside world. Defined in RFC-996</description>
+  <port protocol="tcp" port="133"/>
+  <port protocol="udp" port="133"/>
+<service>
```
erig0, line 7: CI doesn't like this because the `service` tag is not closed.

Should be `</service>`, not `<service>`.

erig0 (changes requested):

config/services/statsrv.xml
jcpunk: In theory that is fixed now.

erig0 (approved):

PR #1118, opened 2023-05-05
feat(service) Add syscomlan

jcpunk:
Port 1065 is registered to syscomlan at IANA.  I'll confess I don't totally understand what it is for, but one of our local applications is using this port so I figured I'd share.

### DIFF
--- a/config/services/syscomlan.xml
+++ b/config/services/syscomlan.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>syscomlan</short>
+  <description>Local system communication</description>
+  <port protocol="tcp" port="1065"/>
+  <port protocol="udp" port="1065"/>
+</service>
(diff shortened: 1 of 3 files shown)


config/services/syscomlan.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>syscomlan</short>
+  <description>Local system communication</description>
+  <port protocol="tcp" port="1065"/>
+  <port protocol="udp" port="1065"/>
+<service>
```
erig0, line 7: Should be `</service>`.

erig0 (changes requested):

config/services/syscomlan.xml
jcpunk: In theory that is fixed now.

erig0 (approved):

PR #1131, opened 2023-05-26
Add mqtt service to Foreman

maximiliankolb:
There are two types for Foreman Clients: push-based (SSH from Foreman server/Smart Proxy to your managed hosts) and pull-based (from your managed host to Foreman server/Smart Proxies).
Pull-based Foreman clients use yggdrasil which use mqtt for communication.

current Foreman/Katello documentation: [Configuring a Host to Use the Pull Client](https://docs.theforeman.org/nightly/Managing_Hosts/index-katello.html#Configuring_a_Host_to_Use_the_Pull_Client_managing-hosts)

### DIFF
--- a/config/services/foreman-proxy.xml
+++ b/config/services/foreman-proxy.xml
@@ -3,5 +3,6 @@
   <short>foreman-proxy</short>
   <description>The Smart Proxy is a project which provides a restful API to various sub-systems.</description>
   <include service="foreman"/>
+  <include service="mqtt"/>
   <port protocol="tcp" port="8443"/>
 </service>


erig0:
Hrm. Does it make sense to open `mqtt` for both modes then? Do we need a separate service?

maximiliankolb:
I had a look at "foreman.xml" and "RH-Satellite-6.xml": both services open ports that technically might not be necessary in _every_ scenario. For example, you can run Foreman without using its DHCP/DNS/TFTP capabilities if you only use cloud-init.

What do you think @adamruzicka ?

adamruzicka:
I wouldn't mind including it even when it is not needed. However, shouldn't it go into foreman-proxy and RH-Satellite-6-Capsule instead?

maximiliankolb:
> I wouldn't mind including it even when it is not needed. However, shouldn't it go into foreman-proxy and RH-Satellite-6-Capsule instead?

I thought it was Foreman because you can register hosts directly to Foreman server?

maximiliankolb:
> Where you register hosts doesn't really have anything to do with it. ssh/mqtt can be configured separately for each proxy/capsule

My throught process: There's a "foreman" service in firewalld which is only activated on Foreman server; and there's "foreman-proxy" which runs on all Smart Proxies; that is Foreman server and all Smart Proxy servers.

the documentation PR also enables this port for Foreman server: https://github.com/theforeman/foreman-documentation/pull/2206/files#diff-df05e5097d91fd3438fdab587618e488bcf443b8d68ecab0a64c1120142ff7f2R26

Does both "foreman" and "foreman-proxy" run on Foreman servers and only "foreman-proxy" on Smart Proxy servers?

adamruzicka:
> Does both "foreman" and "foreman-proxy" run on Foreman servers and only "foreman-proxy" on Smart Proxy servers?

Yes. In Satellite lands the documentation sometimes distinguishes between internal (foreman-proxy on foreman server) and external (foreman-proxy on smart proxy servers) capsules.

erig0:
IMO, its up to the foreman folks to decide what is best for their users. I will happily merge whatever you land upon.

maximiliankolb:
Relates to https://github.com/theforeman/foreman-documentation/pull/2673

erig0:
@maximiliankolb , What would you like to do with this PR?

maximiliankolb:
Thanks for the ping @erig0 

I have moved the service to "foreman-proxy" as suggested & adjusted my commit message. cc @adamruzicka

ekohl:
As for the foreman vs foreman-proxy: at one point I looked at it and opened https://github.com/firewalld/firewalld/pull/1093. I don't have time to take that further, but perhaps @maximiliankolb wants to take a look.

ekohl:
I've taken the time to rewrite https://github.com/firewalld/firewalld/pull/1093 and I think that's a better approach. Merging this would create a situation where part of the optional components are listed for foreman, but another part in foreman-proxy. I think my PR is consistent.

maximiliankolb:
I am closing my PR because I'm off next week and want to avoid confusion. I'll have a look at Ewouds PR soon and report back.

PR #1150, opened 2023-06-16
fix(applet): ActivationReason.Trigger

buxit:
QtWidgets.QSystemTrayIcon.Trigger was moved to QtWidgets.QSystemTrayIcon.ActivationReason.Trigger in qt6


erig0:
Hi, is this an enum? I made a bunch of changes to support Qt6 in 410420ea9f1ff8e1c96218f6e91077cd386836e3. Perhaps I missed this one.

PR #1154, opened 2023-07-04
fix(icmp): fix check_icmpv6_name() to use correct IPv6 names

thom311:
This was only called from IPSet.check_entry() to validate the input. Also, some of the IPv4 names are the same as for IPv6. Overall, the impact of this was probably low.

Fixes: 11567b74317e ('New firewall.core.icmp providing names and types for icmp and icmpv6 values')


erig0 (approved):
Thanks!

PR #1155, opened 2023-07-04
fix(tests): avoid deprecated assertRaisesRegexp for assertRaisesRegex

thom311:
Avoid warnings:
```
    src/tests/python/firewalld_test.py::TestFirewallD::test_zone_add_get_query_removeForwardPort
      /data/src/firewalld/src/tests/python/firewalld_test.py:226: DeprecationWarning: Please use assertRaisesRegex instead.
        self.assertRaisesRegexp(Exception, 'ALREADY_ENABLED', self.fw_zone.addForwardPort, zone, port, protocol, toport, toaddr, 0)
```
unittest.TestCase.assertRaisesRegexp is deprecated for assertRaisesRegex since Python 3.2. It exists since Python 3.1. Even CentOS Linux 8 brings Python 3.6.

- https://docs.python.org/3.3/library/unittest.html#unittest.TestCase.assertRaisesRegex
- https://bugs.python.org/issue10273

### DIFF
--- a/src/firewall/functions.py
+++ b/src/firewall/functions.py
@@ -34,7 +34,6 @@ import socket
 import os
 import os.path
 import shlex
-import pipes
 import string
 import tempfile
 from firewall.core.logger import log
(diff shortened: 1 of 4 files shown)


erig0 (approved):
lgtm. Thanks for the clean ups.

PR #1157, opened 2023-07-06
[th/fw-in-container] add script for starting a podman container for testing/development

thom311:
adds two scripts:

- `contrib/fedora/configure-for-system.sh`: configures firewalld in a way similar as the Fedora spec file. The result is, if you type `make install` it will overwrite the files from the package management system. You wouldn't want to do that on your workstation, as it messes with your installation. It's useful in a VM or a container, to quickly install firewalld from sources, and where we don't care about potential fall-out.

- `contrib/fedora/configure-for-system.sh` starts a rootless, priviledged, system-container of Fedora in podman. It bind mounts the current git working directory. The idea is, you just type 
```
(cd /firewalld && m && m install && systemctl daemon-reload && systemctl restart firewalld.service && PAGER= systemctl status firewalld.service)
```
(which you find in the bash history) to run firewalld from sources.

As this is a rootless container, not *everything* works. But most if it works. Currently 3 tests fail. Probably that should be fixed.

The benefit is to have an easy way for testing firewalld changes, without messing up your system or without need to setup a VM. Of course, setting up a VM can also be fast and convenient (e.g. if you use vagrant and/or ansible). Doing it from scratch without an existing script is not fast. This script does that, except using a container instead of a VM.

Also, the source directory from the host is shared with the container. So you can conveniently edit files from both sides.


erig0:
> The benefit is to have an easy way for testing firewalld changes, without messing up your system or without need to setup a VM.

FWIW, `make check` target is already non destructive. All tests are run inside network namespaces. It does require root however.

erig0 (commented):
> This script configures firewalld the same way, as during our RPM package build.

Why not just do `make rpm` then?

thom311:
> Why not just do `make rpm` then?

Indeed, that would work too, as the package build for firewalld is quite fast (16 seconds on my machine). But it's not as fast as `(cd /firewalld && make install -j 10 && systemctl daemon-reload && systemctl restart firewalld.service)` (1 second).

thom311:
> FWIW, make check target is already non destructive. All tests are run inside network namespaces. It does require root however.

It's more for manual testing during development. Some `firewall-cmd` commands and reading the log.

I also would prefer to avoid messing with my system as root. To be fair, I think for most tests, `make check` could be adjusted not to require root -- except for tests that require `modprobe` or the like.

erig0:
> I also would prefer to avoid messing with my system as root. To be fair, I think for most tests, `make check` could be adjusted not to require root -- except for tests that require `modprobe` or the like.

In theory, when the test starts we could detect if `uid == 0`, if not, then we try to create a user namespace. This can all be done inside the existing macros, e.g. `FWD_START_TEST`.

thom311:
> I would much rather pursue running make check as an unprivileged user than merge this PR. I will put it on my TODO list for a couple weeks from now.

Running tests is only one aspect of the fw-in-container script. It's really useful for ad-hoc, manual debugging in that it allows to spin up a new, rootless test environment. There is little overlap with improving tests. The only overlap is that it also should be possible to run the tests inside `fw-in-container` (and it mostly is, a small number of tests currently fail)

erig0:
I don't like this PR. I get that it's useful to _you_ for testing. However it includes a lot of cruft that was copied from NetworkManager. The upstart/build cost of the container is similar to starting and provisioning a VM. The main difference is that the VM's hypervisor reserves memory; avoiding that is nice. The `make check` testsuite does require root, but tests run inside network namespaces and thus don't interfere with the host machine. The main benefit of this PR seems to be that you can run it as an unprivileged user.

At any rate, it's under `contrib`. We won't install, ship, or support it. I suppose that makes it harmless to have in the repository.

---

The testsuite failures are caused by the netlink messages being too large (probably limited due to user namespace). It's probably not a bad idea for us to limit how much we are willing to send to libnftables at a single time. Which means we will trade atomic transactions for lower memory usage. There are other scenarios in which firewalld's rule application is not atomic, e.g. large ipsets.

It's debatable if the batching should be implemented in libnftables instead of doing it in firewalld.

The above can be addressed independently of this PR or deferred until it's a real problem.

thom311:
Thanks Eric!

PR #1159, opened 2023-07-07
docs: fix reference to lockdown-whitelist.xml in SYNOPSIS section

mbiebl:
Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1040547


erig0 (approved):
Thanks!

PR #1161, opened 2023-07-10
improvement(doc): list required Python version in "README.md"

thom311:
It's not clear which is the minimally supported/required Python version. Document it. Pick 3.6, because that version is in Centos7 and contains a lot of API that we may want to rely on.

Optimally, we would have CI tests that check the minimum version. I think currently we don't test Python 3.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ These are the runtime dependencies:
 
 ```
   linux >= 5.3
+  python >= 3.7
   python3-dbus
   python3-gobject
   python3-nftables >= 0.9.4


erig0:
firewalld on CentOS-7 uses python2. I don't feel comfortable documenting python-3.6 unless it's asserted by CI. Currently we only test python 3.7 in CI.

thom311:
> firewalld on CentOS-7 uses python2. I don't feel comfortable documenting python-3.6 unless it's asserted by CI. Currently [we only test python 3.7 in CI](https://github.com/firewalld/firewalld/blob/75634eb18b734b58b6fc34b42cf17d4306a5170d/.github/workflows/testsuite.yml#L26).

`yum install python3` works on CentOS 7, and you get Python 3.6. That the default `/usr/bin/python` is Python 2 (and unusable to firewalld) seems not relevant. The point is, that Python 3.6 is *so* old, that it's even available on CentOS 7 (a system where recent firewalld doesn't even run). The larger point is, that this dependency really should be everywhere, where a user reasonably is attempting to run firewalld 2.0+.

Currently no minimal Python version is documented. That means for example, when I want to use `shlex.quote()` (which is documented since to exist since v3.3), I don't know whether that can be used in the firewalld code base. It may be infeasible to document every dependency exactly, but at least the minimal Python version should be documented.

Is your point that the documented version should be Python 3.7? I don't mind that...

erig0 (approved):

Issue #699, opened 2020-10-19
firewalld sets with timeouts on nftables back end not getting the timeouts.


nikconwell:
Hi, as per note in [the firewalld-users list](https://lists.fedorahosted.org/archives/list/firewalld-users@lists.fedorahosted.org/thread/2ZY73SZWGAAIL6PYCO7UHJ5ZMGH6PMLT/), I'm opening this about firewalld timed sets not being created with a timeout with the nftables back-end.

In CentOS 8 with firewalld-0.8.0-4.el8.noarch:

```
# firewall-cmd --permanent --new-ipset set3 --type=hash:ip --family=inet --option=timeout=5
# firewall-cmd --reload
```
however listing the set shows it doesn't have a timeout:

```
# nft list table inet firewalld|grep -A4 set
set set3 {
type ipv4_addr
flags interval
}
```

and in fact when I add something to the running set it doesn't disappear after 5
seconds:

```
# firewall-cmd --ipset=set3 --add-entry=1.2.3.4
success

# nft list table inet firewalld|grep -A4 set
set set3 {
type ipv4_addr
flags interval
elements = { 1.2.3.4 }
}
```

I see in the docs that firewall-cmd won't list members of timeout sets, but nft list should show them since that is where they are controlled.  (And actually, if you want me to open up an issue, shouldn't the firewall-cmd be able to list the members by calling nft list and parsing out the elements at that particular time?)

firewalld rocks!  Thanks for creating this service.
-nik

PR #1167, opened 2023-07-12
fix(ipset): fix configuring "timeout","maxelem" values for ipsets with nftables

thom311:
With
```bash
    firewall-cmd --permanent --delete-ipset=testipset || :
    firewall-cmd --permanent --new-ipset=testipset --type=hash:ip --option=maxelem=65536 --option=family=inet --option=hashsize=4096 --option=timeout=14400
    firewall-cmd --reload
```
firewalld would send the JSON request
```json
    {
      "add": {
        "set": {
          "family": "inet",
          "table": "firewalld",
          "name": "testipset",
          "type": "ipv4_addr",
          "flags": [
            "interval"
          ],
          "timeout": "14400",
          "size": "65536"
        }
      }
    },
```
But the "timeout","size" keys are NUMBER types in libnftables-json. They are silently ignored otherwise ([1]). The fix is to pass them as numbers.

Try also:
```bash
    nft delete table inet testtable &>/dev/null || :
    nft add table inet testtable
    echo '
        {
          "nftables": [
            {
              "metainfo": {
                "json_schema_version": 1
              }
            },
            {
              "add": {
                "set": {
                  "family": "inet",
                  "table": "testtable",
                  "name": "testipset",
                  "type": "ipv4_addr",
                  "flags": [
                    "interval"
                  ],
                  "timeout": "14400",
                  "size": "65536"
                }
              }
            },
            {
              "flush": {
                "set": {
                  "family": "inet",
                  "table": "testtable",
                  "name": "testipset"
                }
              }
            }
          ]
        }
    ' | nft -j -f -
    nft list ruleset | grep -C6 testipset
```
[1] https://git.netfilter.org/nftables/tree/src/parser_json.c?id=1335ade24e55199069b8ae79e34746a59ae48c01#n1440

https://github.com/firewalld/firewalld/issues/699
https://bugzilla.redhat.com/show_bug.cgi?id=2055330

Fixes: 1582c5dd736a ('feat: nftables: convert to libnftables JSON interface')

### DIFF
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -1618,9 +1618,9 @@ class nftables:
 
         if options:
             if "timeout" in options:
-                set_dict["timeout"] = options["timeout"]
+                set_dict["timeout"] = int(options["timeout"])
             if "maxelem" in options:
-                set_dict["size"] = options["maxelem"]
+                set_dict["size"] = int(options["maxelem"])
 
         return [{"add": {"set": set_dict}}]
 
--- a/src/tests/regression/rhbz1506742.at
+++ b/src/tests/regression/rhbz1506742.at
@@ -1,5 +1,5 @@
 FWD_START_TEST([ipset with timeout])
-AT_KEYWORDS(ipset rhbz1506742)
+AT_KEYWORDS(ipset rhbz1506742 rhbz2055330 gh699 gh908)
 
 CHECK_IPSET
 
@@ -9,7 +9,9 @@ NFT_LIST_SET([foobar], 0, [dnl
     table inet firewalld {
         set foobar {
             type ipv4_addr
-            flags interval
+            size 1000000
+            flags interval,timeout
+            timeout 10m
         }
     }
 ])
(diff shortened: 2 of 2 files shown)


src/firewall/core/nftables.py
```diff
@@ -1618,10 +1620,10 @@ def build_set_create_rules(self, name, type, options=None):
...
 
         if options:
-            if "timeout" in options:
-                set_dict["timeout"] = options["timeout"]
-            if "maxelem" in options:
-                set_dict["size"] = options["maxelem"]
+            firewall.functions.dict_add_number(
+                set_dict, "timeout", options.get("timeout")
+            )
+            firewall.functions.dict_add_number(set_dict, "size", options.get("maxelem"))
```
erig0, line 1626: Could always do `int()`. `dict_add_number()` isn't adding any value here. The return value is not being checked. It silently fails if the value is None. I'd rather let it fail loudly. The type/values should be checked long before we get to the code path.

```
            if "timeout" in options:
                set_dict["timeout"] = int(options["timeout"])
            if "maxelem" in options:
                set_dict["size"] = int(options["maxelem"])
```

erig0 (changes requested):

src/firewall/core/nftables.py
thom311: changed.

erig0 (changes requested):
Changes lgtm. Can you put the bugs in a Fixes tag so github automatically closes the bug?
e.g.
`Fixes: #699`

thom311:
> Changes lgtm. Can you put the bugs in a Fixes tag so github automatically closes the bug? e.g. Fixes: #699

Done!! (?)

erig0 (approved):

PR #1170, opened 2023-07-13
add `.flake8` configuration file to ignore errors

thom311:
add a `.flake8` config file. This allows to run plain `flake8` in our source tree, and it does the right thing.

Also, drop a few errors from the ignore list, which don't produce any failures (anymore).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -81,10 +81,8 @@ EXTRA_DIST = \
 
 CLEANFILES = *~ *\# .\#* *.py?
 
-FLAKE8_IGNORE = E126,E127,E128,E201,E202,E203,E211,E221,E222,E225,E226,E231,E241,E251,E261,E262,E265,E266,E301,E302,E303,E305,E306,E402,E501,E502,W503,W504,E722,E741
 check-local:
-	find . -name '*.py' -or -name '*.py.in' |xargs flake8 --ignore="$(FLAKE8_IGNORE)"
-	flake8 --ignore="$(FLAKE8_IGNORE)" $(dist_bin_SCRIPTS_in) $(dist_sbin_SCRIPTS_in)
+	flake8
 	@echo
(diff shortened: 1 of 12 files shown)


thom311:
Odd, in github CI I see flake8 warnings for `E126`, `E226`, `E241`, `W503`, `W504`. I am not able to reproduce those warnings.

I even tried setting up a python-3.7 virtual environment and install the same flake8 version via pip. Dunno...

Eventually, I think we should reduce the number of ignores. So having warnings that only trigger in github CI is odd, as it's unclear why they are needed there, but not otherwise (for me).

src/Makefile.am
```diff
@@ -80,7 +80,7 @@ EXTRA_DIST = \
 
 CLEANFILES = *~ *\# .\#* *.py?
 
-FLAKE8_IGNORE = E126,E127,E128,E201,E202,E203,E211,E221,E222,E225,E226,E231,E241,E251,E261,E262,E265,E266,E301,E302,E303,E305,E306,E402,E501,E502,W503,W504,E722,E741
+FLAKE8_IGNORE = E126,E127,E128,E201,E202,E203,E211,E221,E222,E225,E226,E231,E241,E251,E261,E262,E265,E266,E301,E302,E303,E305,E306,E402,E501,E502,E722,E741,W503,W504
```
erig0, line 83: Can we remove this since you added a `.flake8` ? i.e. `flake8 --config $(top_src_dir)/.flake8`

erig0 (changes requested):
Otherwise, lgtm.

src/Makefile.am
thom311: yes, I don't think it requires `--config`, it should fine the file automatically.

Also, then the `.flake8` file should be dist-ed, so that the test works with the tarball too...

changed.

erig0 (dismissed):
I fixed occurrences of E721 that trigger on flake8 6.1.0 (pyflakes 3.1.0). Will merge after CI.

erig0 (approved):

PR #1173, opened 2023-07-19
[th/black] reformat all Python files with python-black

thom311:
Quote commit message:

Automated code formatting is great. Use python-black for that ([1]).
Black us not configurable much (uncompromising) and follow PEP8.
It thus generated quite standard, pleasing format.

But the point isn't whether the formatting is always the most
subjectively pleasing. The point is that the style is generated and
enforceable by an automated tool. It takes away places for nit picking,
while resulting in a more consistent code base.

As this reformats the entire code base, it introduces a large diff and
interferes with tools like git-blame. That will be somewhat alleviated
by adding a "blame.ignoreRevsFile" for git next.

The change is done automatically, by running python-black version
black-22.8.0-2.fc38.noarch:

    $ black .

The result is the same as

    $ ( git grep -l '#!.*\(PYTHON\|python\)' ; git ls-files '*.py' ) \
        | sort -u \
        | xargs -d '\n' black

Note that the exact result depends on the version of black. Currently we
use black==22.12.0 from pip. Yes, this means when we bump the version
in the future, some reformatting will happen.

[1] https://github.com/psf/black


thom311:
@erig0 hi, what do you think about this?

.git-blame-ignore-revs
erig0, line 1: As you said in our discussion, this will have to be dropped and added later. Github will change the commits SHAs after the merge.

.github/workflows/source-checks.yml
```diff
@@ -43,6 +43,11 @@ jobs:
           ./configure
           make -j $(nproc)
 
+      - name: check formatting with python-black
+        run: |
+          black --version
+          black --check .
```
erig0, line 49: This means PRs/CI will fail if formatting is not exact. I'm okay with that, but it does mean contributors/developers will have to use this EXACT version of black.

I do worry that we'll be chasing availability of black versions. I think it effectively forces everyone to use what's available in pip. It also imposes a bit of maintenance burden to update black occasionally.

erig0 (changes requested):
Drop the last commit, and I think we're good to merge.

I think this will force contributors to do something like this before submitting PRs:
```
git rebase -i upstream/main -x "black ."
```

Lastly, I bet this make a lot of the flake8 checks pass. Maybe we need to re-evaluate that list of exceptions.

.git-blame-ignore-revs
thom311: I didn't know `git rebase -i upstream/main -x "black ."` exists. This is amazing.

I already expected to need a `git-filter-branch`/`git-filter-repo` command (which is entirely non-trivial to me).

thom311: > Lastly, I bet this make a lot of the flake8 checks pass.

It probably makes some flake8 options pass, that are about the style anyway. But since the style will be generated, those warnings are not very relevant. Also, some flake8 options are known to disagree with black (e.g. those about where to put the line break).

Nonetheless, reviewing the flake warnings is of course a good idea!

erig0 (approved):
Just waiting on one last out-of-band ACK before I merge this.

PR #1174, opened 2023-07-19
[th/no-all-module-variable] improvement(style): drop __all__ module variables

thom311:
Their main practical use is to control what gets imported by `from
MODULE import *`.

But we never use such an import, and it seems bad style anyway.

Otherwise, `__all__` could act as a documentation of what is supposed to
be "public" API. For example, we could write a tool that checks that
every function that is considered public has a docstring. That doesn't
seem very useful.

Sure, our users install the firewalld Python code, so all of our code
provides some sort of API. It would therefore be good to properly
document it and make clear which part of it is supposed to be public.
But in practice, we probably have few external users. Also, I don't
think that we are sufficiently careful to not break API for the exported
symbols, so the API we mark with __all__ isn't really stable API for
external users.

Overall, it seems a pointless effort to maintain the `__all__` variable.
Often we forget to expose API, or we expose API that isn't actually used
by anybody outside the module that provides it. And nobody notices,
because it has no practical effect for us.

Just drop it.

### DIFF
--- a/src/firewall/command.py
+++ b/src/firewall/command.py
@@ -21,8 +21,6 @@
 
 """FirewallCommand class for command line client simplification"""
 
-__all__ = [ "FirewallCommand" ]
-
 import sys
 
 from firewall import errors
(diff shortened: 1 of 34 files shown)


erig0 (approved):

erig0:
Thanks. I've always hated these.

PR #1177, opened 2023-07-26
Revert "firewall/client: handle_exceptions: Use loop in decorator"

thom311:
From the commit message of e77a70ca4dad ('firewall/client: handle_exceptions: Use loop in decorator') it's not clear to me, what the purpose of NotAuthorizedLoop is.

It seems wrong to me, and very ugly. It's already ugly to wrap most function in a @handle_exceptions decorator. This retry loop seems extra ugly.

Revert it, to drop some code.

If a change is non-obvious, then the reasoning or background information must be recorded (e.g. in the commit message or a code comment). Otherwise, changes are bound to be reverted or broken.

This reverts commit e77a70ca4dade324b12b7a4bfc1f479fb1b5ce9e.

### DIFF
--- a/src/firewall-config.in
+++ b/src/firewall-config.in
@@ -1249,7 +1249,6 @@ class FirewallConfig(object):
         self.fw = client.FirewallClient(wait=1)
         self.__use_exception_handler = True
         self.fw.setExceptionHandler(self._exception_handler)
-        self.fw.setNotAuthorizedLoop(True)
 
         self.fw.connect("panic-mode-enabled", self.panic_mode_enabled_cb)
         self.fw.connect("panic-mode-disabled", self.panic_mode_disabled_cb)
--- a/src/firewall/client.py
+++ b/src/firewall/client.py
@@ -37,45 +37,39 @@ import dbus
 import traceback
 
 exception_handler = None
-not_authorized_loop = False
 
 def handle_exceptions(func):
     """Decorator to handle exceptions
     """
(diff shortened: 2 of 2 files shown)


erig0:
I would guess this is for a dbus/polkit prompt. Maybe if the user miss-types a password this will allow them to get re-prompted. Can you check that from firewall-config?

thom311:
you can test it by dropping a file with `/etc/polkit-1/rules.d/40-firewalld.rules` with
```js
polkit.addRule(function(action, subject) {
    if (!action.id.match("org.fedoraproject.FirewallD1.")) {
        return polkit.Result.NOT_HANDLED;
    }
    #return polkit.Result.NO;
    return polkit.Result.AUTH_SELF;
});
```
and start firewall-config.

With `polkit.Result.NO` the code on `master` does give the desired behavior. Meaning, firewall-config tries to authenticate, fails, shows a "OK/Quit" dialog, and OK/Quit behaves as expected. With the revert, OK no longer works and we are left with the other "QUIT" dialog.

Anyway. So the code does something, which may be better than the alternative. What happens is that firewall-config ends up calling `_exception_handler()` in `src/firewall-config.in`, which shows the dialog with `_error()`. OK does nothing, and relies on the loop to retry.

I think the way of slapping a dialog on top of this, blocking the mainloop underneath, relying on "do-nothing" to just retry the request, it all very ugly. But I am not going to change that now, so I close the PR.

PR #1178, opened 2023-07-27
[th/dbus-signals] cleanup handling of D-Bus signals in client.py

thom311:
Currently looking how D-Bus is handled in client.py.

Trying to improve it somewhat.


thom311:
@erig0 , could you please take a look at this?

erig0 (approved):

Issue #1062, opened 2022-11-30
List of firewall-cmd query commands is incomplete (in 0.9.3)


sjvudp:
### What would you like to be added

There is no query to query the existence of a specific zone, or the existence of a specific service that would help to avoid needless creation of a zone or service if such exists already.
Probably there are more than just the two cases mentioned.
(It's possible to query whether a specific service is activated in a zone, but not whether the service exists at all)

### Why is this needed

Trying to write some idempotent firewall configuration script that should not try to create or activate items that do exist already.
For example don't create a service if such service exists already, and don't add a port to a service if such service with such port exists already.
Otherwise the script would create a lot of error messages.

erig0:
So add `--query-zone` and `--query-service`. Not the latter already exists but is used to query if the service is enabled in a zone or policy, e.g. `--zone public --query-service ssh`. So `--query-service` would become overload (which is fine).

PR #1186, opened 2023-08-08
[th/firewall-cmd-show-overview] show more information with plain `firewall-cmd` output

thom311:
Let `firewall-cmd` (without command arguments) show an overview of the current configuration.

The fields that are shown are subject to change and improvement.

Depending on what exactly is requested, this could fix https://github.com/firewalld/firewalld/issues/1062

Example:
```$ firewall-cmd
ActiveZones:
  public (default)
ActivePolicies:
  allow-host-ipv6
    ingress-zones: ANY
    egress-zones:  HOST

usage: 'firewall-cmd --help' for usage information or see firewall-cmd(1) man page
```
```
$ firewall-cmd -v
State: running
Version: 2.0.999
ActiveZones:
  public (default)
Zones:
  FedoraServer
  FedoraWorkstation
  block
  dmz
  drop
  external
  home
  internal
  nm-shared
  trusted
  work
ActivePolicies:
  allow-host-ipv6
    ingress-zones: ANY
    egress-zones:  HOST

usage: 'firewall-cmd --help' for usage information or see firewall-cmd(1) man page
```

### DIFF
--- a/src/firewall/command.py
+++ b/src/firewall/command.py
@@ -8,7 +8,9 @@
 """FirewallCommand class for command line client simplification"""
 
 import sys
+import dbus
 
+import firewall.config
 from firewall import errors
 from firewall.errors import FirewallError
 from dbus.exceptions import DBusException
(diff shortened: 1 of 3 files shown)


src/firewall/command.py
```diff
@@ -625,3 +627,210 @@ def get_ipset_entries_from_file(self, filename):
...
+            exit_code = errors.NOT_AUTHORIZED
+        else:
+            state = "NOT_RUNNING"
+            s = "not running"
+            exit_code = errors.NOT_RUNNING
+        if verbose or state != 'RUNNING':
+            msg += f"State: {s}\n"
+
+        try:
+            v = fw.real_get_property("version", dbus.String)
```
erig0, line 652: If `get_state()` above returned `NOT_RUNNING` then I don't follow why we would attempt any other DBUS call. IMO we should only implement the `real_get_property()` function for the state. If firewalld is not running, then we're not going to have any useful information to print.

src/firewall/client.py
```diff
@@ -3211,9 +3217,14 @@ def checkPermanentConfig(self):
         self._get_fw().checkPermanentConfig()
 
     @handle_exceptions
-    def get_property(self, prop):
-        return dbus_to_python(self._get_fw_properties().Get(
-            config.dbus.DBUS_INTERFACE, prop))
+    def get_property(self, prop, expected_type=None):
+        return self.real_get_property(prop, expected_type)
+
+    def real_get_property(self, prop, expected_type=None):
```
erig0, line 3223: s/real/raw/

src/firewall/client.py
```diff
@@ -2700,6 +2700,9 @@ def addIPSet(self, name, settings):
 
     @handle_exceptions
     def getZoneNames(self):
+        return self.real_getZoneNames()
+
+    def real_getZoneNames(self):
```
erig0, line 2705: I don't see the value is implementing the `real_*()` functions outside of `real_get_property()`. Basically we need to know if firewalld is running. If it's not, then we dump "firewalld is not running" and we're done. If it is running, we can call the normal `get*()` functions.

erig0 (changes requested):
Most of this looks good. I appreciate the clean ups.

src/firewall/client.py
thom311: renamed.

src/firewall/command.py
thom311: The code after NOT_RUNNING anyway handles D-Bus exceptions for every "raw_()" request (because there could be other errors, or firewalld quit just in that moment). As it's already handled, adding more special case seems unnecessary.

In general, since we make a series of D-Bus requests, the configuration in the server can change at any time (in between those requests). So with the current approach, it's anyway not possible to get a 100% self-consistent output. E.g. you might see "State: running" followed by "Version: unknown".

Not to touch anything after NOT_RUNNING, doesn't solve this problem in general, so it's not clear what that the special handling solves.

(btw, libnm's NMClient builds a client-side, fully-consistent cache. Then you can show the consistent information from one point in time. But that is hard to do).

src/firewall/client.py
thom311: If think `@handle_exceptions` is overused and the handler just killing the application is unsuitable in certain cases.

If firewalld is not running, then the output currently would be:
```
State: not running
Version: unknown (client: 2.0.999)

usage: 'firewall-cmd --help' for usage information or see firewall-cmd(1) man page
```
Instead of "Error: org.freedesktop.DBus.Error.NameHasNoOwner: firewalld is not running" from the handler.


You might think, just check for "NOT_RUNNING", afterwards all accesses (like `self.getZoneNames()`) are safe. But they are not. Any subsequent call can still fail with a D-Bus exception, which would result the `@handle_exceptions` handler to kill the process.

Instead, by handling errors ourselves and using raw/real methods, the error behavior is determined by `show_state_and_exit()`.

src/firewall/client.py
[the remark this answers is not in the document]
thom311: That's possible. But that seems non-obvious, obfuscating behavior.

With those `_get_*()` methods, it's clear where a user calls `_get_fw_ipset()`.

erig0, line 1: I like this idea, but instead of adding things like `_get_fw_ipset()` I'd rather override `__getattribute__()`. I think it's more likely that new code written may do `self.fw` and not realize there are these helper functions. It also avoids changing all these call sites.

```python
def __getattribute__(self, attr):
    value = super().__getattribute__(attr)
    if attr in ["fw", "fw_ipset" ... ] and value is None:
        raise dbus.exceptions.DBusException(
            "org.freedesktop.DBus.Error.NameHasNoOwner: firewalld is not running"
        )
    return value
```

src/firewall/client.py
```diff
@@ -3034,6 +3034,9 @@ def addIPSet(self, name, settings):
 
     @handle_exceptions
     def getZoneNames(self):
+        return self.raw_getZoneNames()
+
+    def raw_getZoneNames(self):
```
erig0, line 3039: These are only used for `show_state_and_exit()`. If one fails I would expect them all to fail. So in `show_state_and_exit()` couldn't we have a single failure scenario that prints "unknown" for all things. i.e. fetch all the days in `try`, and if you get an exception set all the printed info to "unknown".

erig0 (changes requested):

PR #1187, opened 2023-08-08
[th/fix-policy-tcp-mss-clamp] fix docs and validation error with <tcp-mss-clamp/> in policy

thom311:
See https://lists.fedorahosted.org/archives/list/firewalld-users@lists.fedorahosted.org/thread/7J423T2P5R3Y6ASNCN4HDPVHZUVHSYGD/


erig0 (approved):
lgtm. Thanks!

PR #1189, opened 2023-08-09
[th/test-in-container-fixes] fixes for running tests in fw-in-container

thom311:


thom311:
all previously failing tests are now investigated and skipped (due to the failures in rootless containers).

... except `nftables flowtable` test, which I don't understand why it fails. Still TODO.

src/tests/features/zone_priority.at
```diff
@@ -1,6 +1,22 @@
 FWD_START_TEST([zone - priority])
 AT_KEYWORDS(zone)
 
+m4_if(nftables, FIREWALL_BACKEND, [
```
erig0, line 4: Please lift this out to a macro in `src/tests/functions.at`. Call it something like `SKIP_IF_FW_IN_CONTAINER`.

Please keep the modifications to the tests as minimal as possible.

src/tests/features/nftables_flowtable.at
```diff
@@ -6,6 +6,10 @@ AT_SKIP_IF([! NS_CMD([command -v iperf3 >/dev/null 2>&1])])
 AT_SKIP_IF([! NS_CMD([command -v sysctl >/dev/null 2>&1])])
 AT_SKIP_IF([! NS_CMD([command -v jq >/dev/null 2>&1])])
 
+dnl I don't understand why, tbut this test fails in fw-in-container.
```
erig0, line 9: Same as other comments. Just call `SKIP_IF_FW_IN_CONTAINER()` at the start of the test.

erig0 (changes requested):

src/tests/features/zone_priority.at
erig0: I'd really like to minimize the changes to the tests. I want to see only `SKIP_IF_FW_CONTAINER` at the start of the test. If you want the comments then they should be in the macro.

I'm not concerned about skipping `iptables`. It's not the default backend and is still covered by other testing/ci.

thom311: `SKIP_IF_FW_CONTAINER` already  has a general comment, saying that it's skipped because the test is somehow not suitable in fw-in-container environment. But the exact reason why, depends on the test. And that can only be a comment in the test.

The comment explains the test-specific reason why the test is skipped. That cannot move to functions.at.

It takes time to investigate why a test fails, to identify that it fails for good reasons, and that it should be skipped. Those reasons are non-obvious, and if they are not recorded in a comment, it's hard to find out afterwards. Well, I'd have to dig through the git history that added the SKIP, like I did for commit "test(pr323): add comment about why CHECK_MODULE_PROTO_GRE is skipped". But that costs time, repeatedly. It's not an improvement to omit a useful comment.

We should not skip a perfectly passing iptables test, only because it would be too cumbersome to add an `IF(nftables)`. If an iptables test fails, the first thing I will do is reproduce it in fw-in-container. It's not helpful, if the test would be skipped in the debugging environment (although it is supposed to pass).

erig0: Please lift the specific reasons out to a macro. Many of these are the same, e.g. socket buffer. These are workarounds for a specific test environment. I'd rather put the complexity in the macro so the test case reads easier. Please call the macro at the start of the tests like other `CHECK_*` macros.

thom311: updated. How about now?

src/tests/regression/ipset_scale.at
```diff
@@ -16,4 +37,7 @@ FWD_CHECK([--permanent --ipset foobar --add-entries-from-file ./entries], 0, [ig
 
 FWD_RELOAD()
 
+# We only ran a part of the test. Overall, mark the test as skipped.
```
erig0, line 40: Why bother with the above if we're just going to skip here? Just call `SKIP_IF_FW_IN_CONTAINER_WITH_NFTABLES()` at the start of this test.

erig0 (changes requested):
Otherwise, lgtm.

src/tests/regression/ipset_scale.at
thom311: The test might still do something interesting and testworthy with a reduced configuration.

Sure, it intentionally creates a large input (of mostly redundant input data). We cannot test *that* aspect. But otherwise, all steps with a smaller configuration should pass.

Skipping the test entirely, without running the parts that should pass, seems a wasted opportunity for coverage. As the hardest part was already done: to write the test.

contrib/fw-in-container/fw-in-container
```diff
@@ -235,6 +235,7 @@ journalctl -u firewalld.service --since '3 min ago'
 m
 make
 make install
+make check TESTSUITEFLAGS="-j4"
```
erig0, line 238: Could do:
```
make check TESTSUITEFLAGS="-j$(nproc)"
```

erig0 (changes requested):

contrib/fw-in-container/fw-in-container
thom311: This line is in the pre-generated bash history of the container, and the point is to type `ARROWUP`, `PAGEUP` or `CTRL+R` to find the useful command. Usually, followed by editing the line found there, e.g. by adding a `-k FILTER` option.

Anyway. I care that there is an example at hand, how to run the testsuite in parallel. Changed.

src/tests/regression/ipset_scale.at
thom311: changed.

erig0 (approved):
Thanks for your patience.

PR #1192, opened 2023-08-10
fix(nftables): restart with CleanupOnExit=no causes duplicate rules

erig0:
https://bugzilla.redhat.com/show_bug.cgi?id=2222044


src/firewall/core/nftables.py
```diff
@@ -399,13 +399,14 @@ def flush_rules(self, log_denied):
...
-        if TABLE_NAME in self.created_tables["inet"]:
-            rules.append({"delete": {"table": {"family": "inet",
-                                               "name": TABLE_NAME}}})
-            self.created_tables["inet"].remove(TABLE_NAME)
-
-        self.set_rules(rules, log_denied=log_denied)
+        try:
+            self.set_rule({"delete": {"table": {"family": "inet", "name": TABLE_NAME}}},
+                          log_denied=log_denied)
+        except ValueError:
```
thom311, line 405: Please split the patch.

Most of the patch is about replacing `self.created_tables` with `self.policy_table_active`.
As such, here we should do:

```python
if self.policy_table_active:
    self.policy_table_active = False
    self.set_rule(...)
```

The change with `ValueError` may still be good, but please do that in a follow-up patch, with a proper commit message.

--- 

In particular, the patch is relatively large but has a `Fixes: rhbz2222044` comment. As most of the change is an internal refactoring, the relevant part (of always calling `self.set_rule()`) should be in a separate/smaller patch.

src/firewall/core/ipXtables.py
```diff
@@ -576,7 +576,7 @@ def build_flush_rules(self):
             # Set counter to zero: -Z
             for flag in [ "-F", "-X", "-Z" ]:
                 rules.append(["-t", table, flag])
-        return rules
+        return self.set_rules(rules, log_denied=log_denied)
```
thom311, line 579: the `return` can be dropped.

src/firewall/core/ebtables.py
```diff
@@ -229,7 +229,7 @@ def build_flush_rules(self):
             # Set counter to zero: -Z
             for flag in [ "-F", "-X", "-Z" ]:
                 rules.append(["-t", table, flag])
-        return rules
+        return self.set_rules(rules, log_denied=log_denied)
```
thom311, line 232: the `return` can be dropped.

src/firewall/core/nftables.py
```diff
@@ -404,7 +404,8 @@ def build_flush_rules(self):
             rules.append({"delete": {"table": {"family": "inet",
                                                "name": TABLE_NAME}}})
             self.created_tables["inet"].remove(TABLE_NAME)
-        return rules
+
+        self.set_rules(rules, log_denied=log_denied)
```
thom311: the later commit adds a `try:` around the `self.set_rules()` call.

Even if firewalld tracks some state (to know that it configured the table earlier), there is no guarantee that the table wasn't deleted in the meantime (or any other reason, why the nftables command might fail). It seems that an unchecked except at this point is fatal, as it backs out in the middle of a larger operation.

OK, the next commit fixes that, by adding the try-catch. But what about the other backends? An invocation of iptables may fail at any time, it seems fragile to not catch exceptions. Maybe the set-rules methods shouldn't signal a problem via an exception (return code instead), because it's cumbersome to handle them. Or as they do, we really should always handle `ValueError` for every call.

src/firewall/core/nftables.py
thom311: also, before (after?) deleting the table, we should always set `self.policy_table_active = False`

src/firewall/core/nftables.py
thom311: > iptables may fail at any time, it seems fragile to not catch exceptions

The fragile part is, that optimally we would not trust /sbin/iptables or libnftables to always succeed and handle/anticipate failures. Letting an exception escape from deep down during `flush()`, probably messes badly with the internal state of firewalld.

src/firewall/core/fw.py
```diff
@@ -482,36 +482,22 @@ def _start_load_direct_rules(self):
...
 
         # complete reload: unload modules also
         if reload and complete_reload:
             log.debug1("Unloading firewall modules")
             self.modules_backend.unload_firewall_modules()
 
-        self.apply_default_tables(use_transaction=transaction)
-        transaction.execute(True)
-        transaction.clear()
+        self.apply_default_tables()
```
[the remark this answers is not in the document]
erig0: > It's not clear to me why that is done. Could you elaborate in the commit message?

Do you need something beyond the below? (the commit message)

```
chore(fw): make flush an isolated operation

This is preparation for the next commit which addresses a limitation of
the nftables backend.
```

src/firewall/core/nftables.py
erig0: I will attempt to split it, although I think you're misunderstanding the original bug. This is not just a refactor. The bug is _caused_ by `created_tables`. Anyways. I think I can simplify it a bit.

src/firewall/core/ipXtables.py
erig0: Yes. Thanks fro catching.

src/firewall/core/nftables.py
erig0: What we would we do if this fails? IMO failure to flush() is fatal anyways.

thom311: ok. then it would be handled by the try/except in `src/firewall/core/fw.py:start()`?

I just find it ugly, that we are in the middle of doing something (internal state was already partly updated, and in the middle something went wrong). It's not clear that backing out in the middle doesn't leave broken state behind. But possibly it does not?

src/firewall/core/nftables.py
```diff
@@ -488,7 +487,6 @@ def build_default_tables(self):
         default_tables = []
         default_tables.append({"add": {"table": {"family": "inet",
                                                  "name": TABLE_NAME}}})
-        self.created_tables["inet"].append(TABLE_NAME)
```
thom311, line 488: `self.policy_table_active = True` ?

src/firewall/core/nftables.py
```diff
@@ -402,7 +402,6 @@ def flush_rules(self, log_denied):
         try:
             self.set_rule({"delete": {"table": {"family": "inet", "name": TABLE_NAME}}},
                           log_denied=log_denied)
-            self.created_tables["inet"].remove(TABLE_NAME)
         except ValueError:
```
thom311, line 405: `self.policy_table_active = False`? (outside the try/except block)

src/firewall/core/nftables.py
thom311: not sure about what I just said... here we only make up the plan to add the table (didn't execute it yet successfully). Optimally, we set the flag TRUE a bit later, when we are confident that the table exists.

Is this flag still useful?

Optimally, we just delete/add the table (which does nothing if the tables is already deleted/added). Like recent `nft destroy table` command.

I guess, the problem is that we append the "delete" command to a list of commands, and we cannot indicate that that command can fail. I thought the workaround for no "destroy" is "add+delete" together? (since "add" does not fail if the table already exists).

thom311: we could add a "Nftables.has_capability()" method, so we can ask it whether "destroy" is supported. If yes, use it. Otherwise, do "add+delete".

erig0: `policy_table_active` is used for the _policy_ table, i.e. `firewalld_policy_drop`. This code is for the main `firewalld` table.

src/firewall/core/nftables.py
erig0: In most changes, i.e. transactions, they changes are accepted or rejected in whole. In those case the failure doesn't mangle the internal state.

However, in this case the failure is _expected_ in many cases, i.e. clean startup. IMO it makes sense to handle here.

It's true that a `flush` command could fail, but if it does then the firewall backend is broken and we're beyond hope anyways. I don't see value in adding extra error handling. The startup code will attempt to fallback to safe defaults on error; if the defaults also fail then we shut down completely.

src/firewall/core/nftables.py
```diff
@@ -399,13 +399,15 @@ def flush_rules(self, log_denied):
...
         self.policy_dispatch_index_cache = {}
 
-        rules = []
-        if TABLE_NAME in self.created_tables["inet"]:
-            rules.append({"delete": {"table": {"family": "inet",
-                                               "name": TABLE_NAME}}})
+        try:
+            self.set_rule({"delete": {"table": {"family": "inet", "name": TABLE_NAME}}},
+                          log_denied=log_denied)
             self.created_tables["inet"].remove(TABLE_NAME)
```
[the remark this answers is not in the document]
erig0: I'd rather do this:
```python
if TABLE_NAME in self.created_tables["inet"]:     
    self.created_tables["inet"].remove(TABLE_NAME)
```

thom311 (approved):
I still think `self.policy_table_active` should be dropped (and avoid the "table-does-not-exist" error during delete via add+delete.

That aside, lgtm.

thom311 (approved):

erig0:
> I still think `self.policy_table_active` should be dropped (and avoid the "table-does-not-exist" error during delete via add+delete.

Are you suggesting always doing an add/delete for the table when doing a flush() or setting the policy to ACCEPT ?

I suppose that would work. It's odd to add a table if we're going to delete it right away, but it would make the transition it `destroy` easy.

thom311 (approved):
lgtm

PR #1194, opened 2023-08-14
[th/reload-policy] add ReloadPolicy to not drop packets during reload

thom311:
https://bugzilla.redhat.com/show_bug.cgi?id=2149039


src/firewall/core/ipXtables.py
```diff
@@ -587,7 +587,14 @@ def build_set_policy_rules(self, policy):
             if table == "nat":
                 continue
             for chain in BUILT_IN_CHAINS[table]:
-                rules.append(["-t", table, "-P", chain, _policy])
+                if table == "filter" and policy == "DROP":
+                    p = drop_details[chain]
+                    if p == "REJECT":
+                        # Reject is not implemented.
+                        p = "DROP"
```
erig0: Why is REJECT not implemented?

src/firewall/core/nftables.py
```diff
@@ -411,20 +412,19 @@ def build_flush_rules(self):
...
+            {add_del: {"rule": {"family": "inet",
+                                         "table": TABLE_NAME_POLICY,
+                                         "chain": "%s_%s" % ("filter", hook),
+                                         "expr": [{"match": {"left": {"ct": {"key": "state"}},
+                                                             "op": "in",
+                                                             "right": {"set": ["established", "related"]}}},
+                                                  {"accept": None}]}}}
+        ]
+
+    def build_set_policy_rules(self, policy, drop_details):
```
erig0, line 427: This refactor looks unnecessary. I presume it's because you're trying to avoid the rule if the ReloadPolicy is ACCEPT.. but that's not worth the complexity. Just keep the rule in all scenarios.

src/firewall/core/io/firewalld_conf.py
```diff
@@ -331,3 +341,42 @@ def write(self):
...
+                    if k not in result:
+                        valid = False
+                        continue
+                    v = a2[1].strip().upper()
+                    if v not in ("ACCEPT", "REJECT", "DROP"):
+                        valid = False
+                        continue
+                    result[k] = v
+
+        return result, valid
```
erig0, line 378: Please don't return multiple values. If it's invalid raise `FirewallError`.

erig0 (changes requested):

src/firewall/core/ipXtables.py
thom311: just a glitch. The patch wrongly moved to the second commit. Fixed now.

The question I have is whether this is the correct way to do it.  E.g. the `if table == "filter" and policy == "DROP":` check.

src/firewall/core/nftables.py
thom311: Which part looks unnecessary? The loop in `build_set_policy_rules()` needs to iterate over the hooks (`("INPUT", "FORWARD", "OUTPUT")`), as the policy is per-hook. What is your exact suggestion?

(I adjusted this part of the patch somewhat, that `_build_set_policy_rules_ct_rule()` now only returns one rule and not a list of rules. But the gist is the same).

src/firewall/core/io/firewalld_conf.py
thom311: changed.

src/firewall/core/ipXtables.py
erig0: I'm not sure you need to match `table == "filter"`. IIRC the policy can/should be set on the `raw` table as well. Historically, I think the iptables backend blocked _all_ traffic during reload. The nftables only blocks new connections.

I'm okay with changing the iptables behavior to match nftables, but it needs to be a conscious change and probably in a separate commit.

erig0 (changes requested):

src/firewall/core/ipXtables.py
thom311: so should it be `table in ("filter", "raw")` ? Because there are also tables  `mangle` and `security` (`nat` is handled above).

> I'm okay with changing the iptables behavior to match nftables,

So I don't do that now, right? (I didn't plan to).

src/tests/regression/rhbz1940928.at
```diff
@@ -5,6 +5,8 @@ CHECK_IPTABLES
 dnl test triggers a limitation in iptables-restore
 dnl
 AT_CHECK([sed -i 's/^IndividualCalls.*/IndividualCalls=no/' ./firewalld.conf])
+AT_CHECK([sed -i 's/^ReloadPolicy=.*/ReloadPolicy=INPUT:REJECT,FORWARD:ACCEPT/' ./firewalld.conf])
+FWD_RELOAD
```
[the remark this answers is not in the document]
thom311: it's there intentionally, because the reload triggers the use of the ReloadPolicy.

I'll add a code comment.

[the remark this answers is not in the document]
thom311: actually. With your other comment, I no longer hijack this tests. This now moved to another, new test.

src/tests/regression/rhbz1940928.at
```diff
@@ -5,6 +5,8 @@ CHECK_IPTABLES
 dnl test triggers a limitation in iptables-restore
 dnl
 AT_CHECK([sed -i 's/^IndividualCalls.*/IndividualCalls=no/' ./firewalld.conf])
+AT_CHECK([sed -i 's/^ReloadPolicy=.*/ReloadPolicy=INPUT:REJECT,FORWARD:ACCEPT/' ./firewalld.conf])
```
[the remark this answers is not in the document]
thom311: ok.

src/firewall/core/io/firewalld_conf.py
```diff
@@ -63,6 +63,7 @@ def set_defaults(self):
         self.set("AutomaticHelpers", config.FALLBACK_AUTOMATIC_HELPERS)
         self.set("FirewallBackend", config.FALLBACK_FIREWALL_BACKEND)
         self.set("FlushAllOnReload", "yes" if config.FALLBACK_FLUSH_ALL_ON_RELOAD else "no")
+        self.set("ReloadPolicy", self._unparse_reload_policy(self._parse_reload_policy(config.FALLBACK_RELOAD_POLICY)))
```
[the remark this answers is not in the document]
thom311: yes, it's to ensure the value is normalized.

I added a unit tests instead, that `config.FALLBACK_RELOAD_POLICY` is normalized.

src/firewall/core/io/firewalld_conf.py
```diff
@@ -196,6 +197,20 @@ def read(self):
                             config.FALLBACK_FLUSH_ALL_ON_RELOAD)
             self.set("FlushAllOnReload", str(config.FALLBACK_FLUSH_ALL_ON_RELOAD))
 
+        value = self.get("ReloadPolicy")
+        valid = True
+        try:
+            parsed= self._parse_reload_policy(value)
```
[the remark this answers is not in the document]
thom311: fixed

src/firewall/core/io/firewalld_conf.py
```diff
@@ -318,3 +333,47 @@ def write(self):
             raise IOError("Failed to create '%s': %s" % (self.filename, msg))
         else:
             os.chmod(self.filename, 0o600)
+
+    RELOAD_POLICY_DEFAULT = {
+            "INPUT": "DROP",
+            "FORWARD": "DROP",
+            "OUTPUT": "DROP",
```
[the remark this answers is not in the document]
thom311: dropped.

src/firewall/core/ipXtables.py
erig0: Correct. Don't do it now.

src/firewall/core/ipXtables.py
```diff
@@ -684,7 +684,14 @@ def build_set_policy_rules(self, policy):
...
             for chain in BUILT_IN_CHAINS[table]:
-                rules.append(["-t", table, "-P", chain, _policy])
+                if table == "filter" and policy == "DROP":
+                    p = drop_details[chain]
+                    if p == "REJECT":
+                        rules.append(["-t", table, "-A", chain, "-j", "REJECT"])
+                        p = "DROP"
+                else:
+                    p = _policy
+                rules.append(["-t", table, "-P", chain, p])
```
erig0, line 694: In general, I think we should always pass valid `drop_details`. This allows us to make simpler decisions down the chain, e.g. this section could be rewritten as:

```python
                if table == "filter" and policy == "DROP" \
                   and drop_details[chain] == "REJECT":
                    rules.append(["-t", table, "-A", chain, "-j", "REJECT"])
                rules.append(["-t", table, "-P", chain, drop_details[chain]])
```

src/firewall/core/fw.py
```diff
@@ -1090,7 +1090,20 @@ def flush(self, use_transaction=None):
...
+        assert policy in ("ACCEPT", "DROP", "PANIC")
+        if policy == "DROP":
+            if drop_details is None:
+                drop_details = {
+                    "INPUT": "DROP",
+                    "OUTPUT": "DROP",
+                    "FORWARD": "DROP",
+                }
+        else:
+            drop_details = None
```
erig0, line 1103: At this point, lets guarantee that `drop_details` is valid. That allows simpler logic down the call chain.

src/firewall/core/io/firewalld_conf.py
```diff
@@ -247,6 +249,20 @@ def read(self):
...
+            parsed = self._parse_reload_policy(value)
+        except ValueError:
+            value2 = config.FALLBACK_RELOAD_POLICY
+            log.warning(
+                "ReloadPolicy '%s' is not valid, using default " "value %s",
+                value if value else "",
+                value2,
+            )
+        else:
+            value2 = self._unparse_reload_policy(parsed)
```
erig0, line 263: This else block is unnecessary. Just put the value2 assignment inside the try.

src/firewall/core/io/firewalld_conf.py
```diff
@@ -247,6 +249,20 @@ def read(self):
                 )
             self.set("FlushAllOnReload", str(config.FALLBACK_FLUSH_ALL_ON_RELOAD))
 
+        value = self.get("ReloadPolicy")
+        try:
+            parsed = self._parse_reload_policy(value)
```
erig0, line 254: assigning to parsed is not needed. We don't use the value. Bare `self._parse_reload_policy(value)` is sufficient.

Or even:
```python
value2 = self._unparse_reload_policy(self._parse_reload_policy(value))
```

erig0 (changes requested):
This is close to ready. Thanks!

src/tests/features/reloadpolicy.at
```diff
@@ -0,0 +1,13 @@
+FWD_START_TEST([check ReloadPolicy])
+AT_KEYWORDS(reloadpolicy rhbz2149039)
+CHECK_IPTABLES
```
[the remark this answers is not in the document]
thom311: dropped.

src/firewall/core/nftables.py
```diff
@@ -574,38 +570,78 @@ def build_set_policy_rules(self, policy):
...
-            rules.append(
-                {"add": {"table": {"family": "inet", "name": TABLE_NAME_POLICY}}}
-            )
+        elif policy in ("ACCEPT", "DROP"):
+            table_added = False
+            for hook in ("INPUT", "FORWARD", "OUTPUT"):
+                if policy == "ACCEPT":
+                    d_policy = "ACCEPT"
+                else:
+                    d_policy = drop_details[hook]
```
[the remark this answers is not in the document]
thom311: the variable is called `drop_details`.

I guess, with `ACCEPT` we could pass a dictionary `{ "INPUT": "ACCEPT", "FORWARD": "ACCEPT", "OUTPUT": "ACCEPT"}`. But that doesn't seem an improvement.

For one, I would have to rename the variable `drop_details` (maybe to `policy_details`). But more importantly, the purpose of the variable is then different and larger. With `policy=ACCEPT`, it doesn't really make sense to pass any `policy_details` that don't do ACCEPT. So the `policy_details` are redundant, and it's unclear what the purpose of the parameter is in that case. Redundancy is bad.

src/firewall/core/ipXtables.py
thom311: I don't think it's simpler. It extends the purpose of the variable, making it harder to understand.

`drop_details` is simple. If the `policy` is `"DROP"`, then these are the details (the settings from `ReloadPolicy=`).

If the argument also applies to `ACCEPT`, what does it mean there?

src/firewall/core/io/firewalld_conf.py
thom311: the try/except should only guard the code that we expect to fail. We don't expect that `self._unparse_reload_policy(parsed)` can ever fail. It MUST not fail. If this ever would raise an exception, this would be a bug we want to catch.

src/firewall/core/io/firewalld_conf.py
thom311: in the current code the variable `parsed` is used (in `else:`).

src/firewall/core/ipXtables.py
erig0: So rename it to `policy_details`. It's then generic in all cases and we don't need an special casing.

src/firewall/core/fw.py
erig0: The cognitive load argument is only because of the variable name. Pick a better name.

src/firewall/core/ipXtables.py
thom311: with `policy="PANIC"|"ACCEPT"`, what are callers supposed to set the `policy_details` to?

erig0: For PANIC it's all "DROP". For accept is all "ACCEPT".

erig0 (approved):

erig0:
@thom311, please take a look.

thom311:
lgtm

erig0 (approved):

PR #1195, opened 2023-08-16
improvement(build): rename the "master" branch to "main"

thom311:
The default firewalld branch will be renamed from "master" to "main". This is part of the conscious language effort. Many projects did such a change, and with time, "main" is (becoming?) the most popular name for the default branch.

There are little actual changes necessary, except pushing the "main" branch, deleting the "master" branch and fixing a bit of fall-out.

This commit changes a few references of "master" in the source tree.

See-also: https://github.com/github/renaming/#renaming-the-default-branch-from-master


erig0:
I rename master on main. That means all PRs also got re-targeted for `main`. I guess we'll discover the fallout soon. :)

erig0 (approved):

PR #1198, opened 2023-08-17
[th/xml-schema] restore XSD XML schemas for validating config files

thom311:
While firewalld command line tools now have an option `--check-config`,
it's still useful to have a scheme for our XML files.
    
Revert the removal of the XSD schemas.
    
This replaces/obsoletes pull request [1], because I think we should
restore the previous schema, instead of writing a new one.
    
This reverts the removal from commit 81d3d6a7d84bc34a18ddd4ab6153f0dafb2fa17e.

[1] https://github.com/firewalld/firewalld/pull/492

Fixes: https://github.com/firewalld/firewalld/issues/488

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -28,6 +28,14 @@ dist_dbus_policy_DATA = FirewallD.conf
 gsettings_in_file = org.fedoraproject.FirewallConfig.gschema.xml.in
 gsettings_SCHEMAS = $(gsettings_in_file:.xml.in=.xml)
 
+xmlschemadir = $(prefixlibdir)/xmlschema
+dist_xmlschema_DATA = \
+	xmlschema/icmptype.xsd \
+	xmlschema/ipset.xsd \
+	xmlschema/service.xsd \
+	xmlschema/zone.xsd
+dist_xmlschema_SCRIPTS = xmlschema/check.sh
+
 BUILT_SOURCES = \
 	$(desktop_DATA) \
 	$(appdata_DATA) \
@@ -374,6 +382,7 @@ EXTRA_DIST = \
 	$(polkit1_action_FILES) \
 	$(gsettings_in_file) \
 	$(CONFIG_FILES) \
+	$(dist_xmlschema_DATA) \
 	lockdown-whitelist.xml.in \
 	firewalld.init \
 	firewalld.logrotate.in \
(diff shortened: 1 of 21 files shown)


config/services/cratedb.xml
```diff
@@ -2,7 +2,7 @@
 <service>
   <short>CrateDB</short>
   <description>CrateDB is a distributed SQL database management system that integrates a fully searchable document oriented data store.</description>
+  <include service="postgresql"/>
```
erig0, line 5: Is this change meant to be part of this commit?

config/services/minidlna.xml
```diff
@@ -2,6 +2,6 @@
 <service>
   <short>MiniDLNA</short>
   <description>MiniDLNA is a simple media server software with the aim to be fully compliant with DLNA/UPNP-AV clients. Enable this service if you run minidlna service.</description>
-  <port protocol="tcp" port="8200"/>
   <include service="ssdp"/>
+  <port protocol="tcp" port="8200"/>
```
erig0, line 6: ditto

config/services/plex.xml
```diff
@@ -2,8 +2,10 @@
 <service>
   <short>PLEX</short>
   <description>Plex Media Server (PMS) is the back-end media server component of Plex. It organizes content from personal media libraries and streams it to the network.</description>
-  <port protocol="tcp" port="32400"/><port protocol="udp" port="32400"/> <!-- Plex media server access (required)> -->
-  <port protocol="tcp" port="32469"/><include service="ssdp"/> <!-- Plex DLNA -->
+  <include service="ssdp"/> <!-- Plex DLNA -->
```
erig0, line 5: ditto

config/Makefile.am
erig0, line 1: I assume the intention here is to get extra test coverage for xmlschema by adding these. I don't think we should worry about synchronizing these with their upstream (libvirt, et al.) counterparts. In that regard, then I think it's fine to add these.

erig0 (changes requested):
Please split the commit `fix(config): adjust XML schemas to allow previously unsupported parts` so it does not modify the XML and the schema in the same commit. AFIAICS, there is no reason to do them in the same commit.

Otherwise, lgtm.

config/Makefile.am
thom311: exactly. I extended the commit message about that.

thom311:
> Please split the commit `fix(config): adjust XML schemas to allow previously unsupported parts` so it does not modify the XML and the schema in the same commit. AFIAICS, there is no reason to do them in the same commit.

split.

Branch repushed.

erig0 (approved):

PR #1199, opened 2023-08-21
fix(firewall-config): escape markup stored in bindings store

yrro:
Fixes: #499

### DIFF
--- a/src/firewall-config.in
+++ b/src/firewall-config.in
@@ -614,11 +614,11 @@ class FirewallConfig:
         self.bindingsView.append_column(
             Gtk.TreeViewColumn(_("Bindings"), Gtk.CellRendererText(), markup=0))
         self.connectionsIter = self.bindingsStore.append(
-            None, [ _("Connections"), "", "" ])
+            None, [ escape(_("Connections")), "", "" ])
         self.interfacesIter = self.bindingsStore.append(
-            None, [ _("Interfaces"), "", "" ])
+            None, [ escape(_("Interfaces")), "", "" ])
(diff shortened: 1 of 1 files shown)


thom311 (approved):
lgtm, thanks

thom311:
> Are these strings causing an issue? AFAICS, none of these strings need to be escaped. Or do the translated strings need escaped?

the strings come from translation and are not hardcoded to the program itself. They go through a `_()` call.

It would require translators (or the maintainers) to review those strings, and make sure that everything is markup-safe. 
Which is not the case, see #499 or try `G_DEBUG=fatal-warnings LANG=en_GB.UTF8 firewall-config`

yrro:
> Are these strings causing an issue? AFAICS, none of these strings need to be escaped.

Run `firewall-config` with `LANG=en_GB.UTF-8` and you'll see the problem:

image

The tree view on the left is supposed to have a row "Interfaces", but because the cell renderer doesn't like the invalid markup (logging `Failed to set text from markup due to error parsing markup: Unknown tag 'interface' on line 1 char 20`) it doesn't render and instead whenever you move your mouse over the bugged entry the rendered text changes to be a copy of one of the other cells.

erig0:
I follow you now. So this translation is triggering the bug:
https://github.com/firewalld/firewalld/blob/57f584b97ed5878aee9bfddfab04ed2f44c0df52/po/en_GB.po#L147-L148

I suppose we should just escape these given the translation problem.

erig0 (approved):

PR #1201, opened 2023-08-29
[th/contrib-fedora-rpm] import firewalld.spec file from fedora and start unifying the spec files

thom311:
- import firewalld.spec from Fedora rawhide/f39, although f38 is in fact very similar to that (I checked that the differences as they exist are desirable).

- add `./contrib/fedora/rpm/build.sh` script. You can build a Fedora-like RPM by running it. Pass "-h" for options.

- unify the upstream spec file in `./firewalld.spec` with the Fedora spec file under `./contrib/fedora/rpm/firewalld.spec`. This is work in progress and not finished. I also compared the changes with the spec files in the various RHEL (7,8,9) releases. The end goal is to have one spec file that is suitable to build an RPM from upstream git (or from a recent release tarball) which is suitable for Fedora 38+ and RHEL8+. In particular, the goal will be, when we rebase the firewalld tarball in Fedora/RHEL, that we can also use the upstream spec file (with minor modifications).

Another goal will be to use the build script for setting up copr repositories to build RPMs from `main` branch.


thom311:
there are still steps missing.

- To further unify the upstream specfile `./firewalld.spec` and the imported Fedora spec file. It's not clear how far that can go however, because the upstream spec file does intentionally not carry downstream-isms like the distro variants. Why downstream differs so much from upstream, should be questioned and the difference reduced further.

- to also import the RHEL spec file, and unify contrib/fedora/rpm with the RHEL variant.

(WIP).

erig0 (commented):
It looks like we end up with two spec files. `./firewalld.spec` and `contrib/fedora/rpm/firewall.spec`.

I was hoping they would be completely unified.

Edit: NM, I see now that this is unfinished.

thom311:
> It looks like we end up with two spec files. `./firewalld.spec` and `contrib/fedora/rpm/firewall.spec`.
> 
> I was hoping they would be completely unified.
> 
> Edit: NM, I see now that this is unfinished.

I think it's basically finished (at least for now. Followups are of course always possible, but not planned).

`./firewalld.spec` and contrib/fedora are synced to the point, when you look at a diff, you see very clear blocks, where the difference makes sense. Meaning, they are different, but in a way that can be understood by looking at the diff. Also, they are now both in git and side-by-side, so they are easy to keep in sync.

The contrib/fedora files are there to be copied to dist-git Fedora/RHEL on the next package rebase. Of course, they need manual adjustments, like the changelog and the __PLACEHOLDER__. But those differences are again very clear, when you look at the diff. And when we update contrib/fedora, and reimport the file to Fedora/RHEL, it will also be clear how to handle the diff. That's one purpose. Another purpose is to use in copr, and in general, to build Fedora/RHEL-like packages from `main`.

./firewalld.spec has a slightly different purpose. It also doesn't try to make a Fedora/RHEL-like package.

The real "problem" is that there are different downstream variants. If possible, those differences should be reduced. Having a common spec file to see them, helps with that. I think the remaining differences are there for reasons, so it's not easy to fix.

contrib/fedora also requires a `build.sh` script while ./firewalld.spec just does
```
rpm test-rpm: dist
    @rpmbuild -ta $(PACKAGE_NAME)-$(PACKAGE_VERSION).tar.bz2
```
I think the `build.sh` script is useful and necessary, but it makes contrib/fedora only useable via the `build.sh` script.


IMO, ./firewalld.spec should be deleted. All useful parts are taken from it and synced with contrib/fedora. Otherwise, I elaborted what's the use of contrib/fedora. What's the remaining purpose of ./firewalld.spec? I don't mind keeping them both...

erig0:
> I think the build.sh script is useful and necessary, but it makes contrib/fedora only useable via the build.sh script.

I don't agree that the script is necessary. I think we can pass macro values and with values via `--define` and `--with`. That would make unifying the spec files easier.

I realize the script does other things, but I don't think they're necessary. One example is changelog. The spec file changelog is about the _package_. I don't think adding firewalld's git log makes sense.

thom311:
Could you showcase how it could be done without a build script?

thom311:
> Could you showcase how it could be done without a build script?

Or maybe can you clarify what the problem with the script is?

Is the task "solve this without adding a script"? Why is that? Scripts can be very useful tools to make repetitive tasks easier to understand and execute.

---

I can only guess, but is the problem the templates in the `firewalld.spec` file? And that you cannot really use the specfile directly without running the script? What exactly were you trying to do, where this caused problems?

thom311:
btw, while you cannot use the spec file directly, it can be easily generated `./contrib/fedora/rpm/build.sh --sources`. Or the SRPM can be generated via `./contrib/fedora/rpm/build.sh --srpm`.

contrib/fedora/rpm/2001-RHEL-only-Add-cockpit-by-default-to-some-zones.patch
erig0, line 1: I know it's the only way to make "one spec file to rule them all", but I'm rather uncomfortable with carrying these downstream patches in the upstream repository.

contrib/fedora/rpm/build.sh
```diff
@@ -0,0 +1,329 @@
...
+        cat <<- EOF
+	* $(LC_TIME=C date '+%a %b %d %Y') $FW_BUILD_USERNAME - %{epoch_version}:%{version}-%{release_version}%{?snap}
+	- build of firewalld ($DATE, uuid: $UUID, git: $COMMIT_FULL)
+	$(git log -n20 --date=local --format='- %h %s [%an] (%ci)')
+	- ...
+	EOF
+    else
+        echo "$CHANGELOG"
+    fi > "$BUILDDIR/SOURCES/CHANGELOG"
+}
```
erig0, line 247: Please omit changelog support. spec file changelogs are supposed to be about _packaging_. It does not make sense to fill them with the git history.

contrib/fedora/rpm/build.sh
```diff
@@ -0,0 +1,329 @@
...
+if [ "$BUILDTYPE" != "sources" ]; then
+    if [ "$BUILDTYPE" = "srpm" ]; then
+        RPM_BUILD_OPTION=-bs
+    else
+        RPM_BUILD_OPTION=-ba
+    fi
+
+    list_add_arg ARGS_WITH_X    "--with"    "${ARGS_WITH[@]}"
+    list_add_arg ARGS_WITHOUT_X "--without" "${ARGS_WITHOUT[@]}"
+
```
erig0, line 284: Do we anticipate building a `fedora_variant` on a non-fedora OS? If then answer is "no", then we can omit all this with logic entirely as it's detected in the spec file itself.

contrib/fedora/rpm/firewalld.spec
```diff
@@ -0,0 +1,412 @@
...
+%endif
+
+%if 0%{?rhel}
+# The patches starting from 2001+ are downstream-only patches only for RHEL.
+# They always apply, also after are rebase to a newer release tarball.
+Patch2001: 2001-RHEL-only-Add-cockpit-by-default-to-some-zones.patch
+%else
+Source2001: 2001-RHEL-only-Add-cockpit-by-default-to-some-zones.patch
+%endif
+
```
erig0, line 64: Unfortunately, we also often backport features. Maybe we need a placeholder for that as well. In theory it could go in the 9001+ range; so maybe it's not necessary.

contrib/fedora/rpm/firewalld.spec
```diff
@@ -0,0 +1,412 @@
...
+# workstation firewalld.conf
+cp -a %{buildroot}%{_sysconfdir}/firewalld/firewalld-standard.conf \
+    %{buildroot}%{_sysconfdir}/firewalld/firewalld-workstation.conf
+sed -i 's|^DefaultZone=.*|DefaultZone=FedoraWorkstation|g' \
+    %{buildroot}%{_sysconfdir}/firewalld/firewalld-workstation.conf
+
+rm -f %{buildroot}%{_datadir}/polkit-1/actions/org.fedoraproject.FirewallD1.policy
+%endif
+
+%find_lang %{name} --all-name
```
erig0, line 236: I have no idea what this does. Can you explain it?

erig0 (changes requested):
Please also remove the top-level `firewalld.spec` and have `make rpm` build using the one in `contrib`. Having two spec files in the repository is a no-go.

Lastly, I'm not convinced this is more convenient than spec file variants for each distro (RHEL, fedora). It adds a lot of complexity to the spec file and also requires an external build script. In my experience, maintaining the distribution spec files isn't a big deal. You end up diffing them to select the changes which we'll still have to do every time we update the package. I think the distribution nuances belongs in the distribution, not the upstream repository.

So I'll ask, what are we gaining with this PR?

contrib/fedora/rpm/firewalld.spec
thom311: I don't know. You find it both in [Fedora]( It comes from https://src.fedoraproject.org/rpms/firewalld/c/4d63fa14ad0ef462713fd3b8fee80a68adca78e0) and [RHEL](https://gitlab.com/redhat/centos-stream/rpms/firewalld/-/blob/dcf3c6db0375bd2f041674259423fe92d7a0dc74/firewalld.spec#L131).

Maybe: https://fedoraproject.org/wiki/PackagingDrafts/find_langt

contrib/fedora/rpm/firewalld.spec
thom311: When doing backports, I don't think there needs to be a distinction between bugfix and features. In general, I find this distinction often unclear. It's just another patch. 

I would add them in 9001+ range. Also, because it's slightly more cumbersome, to add a patch at 3001 when there are already patches on 9001+. It's easier to just append new patches to the list at 9001+.

But sure, a range for "features" would work too...

contrib/fedora/rpm/2001-RHEL-only-Add-cockpit-by-default-to-some-zones.patch
thom311: I don't mind, because this is all under `contrib/fedora` directory. That makes it clear that this is some non-core extension.

contrib/fedora/rpm/build.sh
thom311: the main reason why `%bcond_with fedora_variant` exists, is so that the various places have a check
```
%if %{with fedora_variant}
```
instead of
```
%if 0%{?fedora}
```

That means, the value of this variable is that you can grep for `fedora_variant`, and find all places related to that.

At that point, making it a bcond variable is a small extension. Yes, it allows to build a fedora-variant on RHEL and vice versa. Nice.

---

More interesting is `firewalld_test`. Maybe `firewalld_test` package should *always* be build (even on RHEL), and we should exclude the `firewalld-test` via other means from installation. But as it is for now, whether that package gets build is a bcond switch. It would seem a loss, if I cannot build a `firewalld-test` package on RHEL.

thom311:
> You end up diffing them to select the changes which we'll still have to do every time we update the package

It becomes trivial, to diff the unified spec file (and spot the difference). Contrary to diffing the diverged spec files.

> Lastly, I'm not convinced this is more convenient than spec file variants for each distro (RHEL, fedora). [...]
>
> So I'll ask, what are we gaining with this PR?

The main benefit is the ability to easily build a RHEL/Fedora-style RPM from upstream git (the script!).

Which further leads to the ability to provide a FHEL/Fedora-style firewalld in copr. Which makes it even easier to use latest firewalld from git. Which means users can easier test latest git (and report issues!!) or that RH QA could run their CI against firewalld from git.

To build firewalld in copr, it needs a build script. This branch, adds the build script.

A unification of the spec file is another bonus.

> In my experience, maintaining the distribution spec files isn't a big deal. 

The downstream spec files are widely diverged, which shows clearly that the previous approach fails to avoid that.

I am not sure what you think "isn't a big deal" here. To hack the spec files and let them diverge? Yes, that was easy indeed. The outcome is however subpar.

- when looking at a diff between SPEC files it's overwhelming and hard to reason which piece is there why, and which piece would be the preferable one.
- presumably, any change to a downstream spec file was done to improve something. In various cases, the other distro was left behind. Meaning, the other distro didn't benefit from the work (our efforts don't reach their full potential), and we have less exposure to users (less testing and feedback), and we have different flavors (different issues, behaviors and testing).
- the spec files widely diverge, with downstream-only adjustments. Those changes should have never made downstream-only in the first place. The upstream project failed to provide a consumable artifact, that worked for the distros without tweaking. Having a unified spec file would have made it clear that some of the divergences are a bad idea.

When upgrading to RHEL10, do you plant to just take what comes from Fedora? Clearly various pieces from the Fedora spec file are not applicable to RHEL. So you either go through the effort of RHEL-ifying the Fedora spec files. Or you keep the RHEL9 spec file, and give up on every reconciling Fedora and RHEL again. Because *this* PR contains the reconciliation now, and if this shouldn't be done, there is no point reconciling in future future.

contrib/fedora/rpm/2001-RHEL-only-Add-cockpit-by-default-to-some-zones.patch
erig0: Some of these patches come from RHEL, not fedora.

erig0:
> > You end up diffing them to select the changes which we'll still have to do every time we update the package
> 
> It becomes trivial, to diff the unified spec file (and spot the difference). Contrary to diffing the diverged spec files.

IMO it's no more or less trivial than the current situation of diffing the existing upstream firewalld.spec with the downstream one.

> > Lastly, I'm not convinced this is more convenient than spec file variants for each distro (RHEL, fedora). [...]
> > So I'll ask, what are we gaining with this PR?
> 
> The main benefit is the ability to easily build a RHEL/Fedora-style RPM from upstream git (the script!).
> 
> Which further leads to the ability to provide a FHEL/Fedora-style firewalld in copr. Which makes it even easier to use latest firewalld from git. Which means users can easier test latest git (and report issues!!) or that RH QA could run their CI against firewalld from git.
> 
> To build firewalld in copr, it needs a build script. This branch, adds the build script.

It adds two scripts. A spec file generator and a COPR build script. The latter is what's needed for COPR, not the former.

> A unification of the spec file is another bonus.
> 
> > In my experience, maintaining the distribution spec files isn't a big deal.
> 
> The downstream spec files are widely diverged, which shows clearly that the previous approach fails to avoid that.

The distributions do what they want; often without discussing it with upstream. I disagree that having all of the distribution nuances in the upstream repository will prevent that.

> I am not sure what you think "isn't a big deal" here. To hack the spec files and let them diverge? Yes, that was easy indeed. The outcome is however subpar.
> 
>     * when looking at a diff between SPEC files it's overwhelming and hard to reason which piece is there why, and which piece would be the preferable one.

If it's in downstream.spec and not upstream.spec then that means it's unique to downstream. I don't think that's hard to reason about.

>     * presumably, any change to a downstream spec file was done to improve something. In various cases, the other distro was left behind. Meaning, the other distro didn't benefit from the work (our efforts don't reach their full potential), and we have less exposure to users (less testing and feedback), and we have different flavors (different issues, behaviors and testing).

Hard disagree. Fedora made some very opinionated choices which IMO were bad, e.g. `FedoraWorkstation` zone, opening up all ports.

erig0 (changes requested):
I will not merge changes adding `contrib/fedora/rpm`. I'm opposed to adding downstream patches to the upstream repository. The distributions are free to make their own decisions; misguided as they may be.

I will merge changes that improve the existing firewalld.spec and enables COPR builds (without downstream patches).

thom311:
ok. Closing then.

PR #1202, opened 2023-08-29
Update Warpinator service config to include required DNS (5353/udp) rule

xsub:
Hi, 

According to: https://github.com/linuxmint/warpinator#secure-mode

_If you are planning to use the Flatpak version to connect to other machines using Flatpak, you need to open UDP port 5353 as well as those mentioned above._ 

This change addresses that requirement. 

Thanks,
xsub


erig0:
Hrm. Is this only relevant to flatpak users then? Do we really want to open the port for all users?

erig0:
No response to needinfo. Closing.

PR #1208, opened 2023-09-07
Add submission service for SMTP over port 587

viniciusferrao:
* This commit adds the correct service name for SMTP mail submission for clients/users using tcp/587 and udp/587 ports.
* It resolves the naming confusion between the service name on firewalld and the defaults on `/etc/services`:
https://access.redhat.com/articles/1761
* It adheres to the defined rules in RFC6409, which can be on: https://www.rfc-editor.org/rfc/rfc6409.html
* It aligns with the service name and port reservations listed by IANA:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=11

This PR does not removes the misleading `smtp-submission` service name to avoid breaking already existing software that relies on the old service name.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -323,6 +323,7 @@ CONFIG_FILES = \
 	services/steam-streaming.xml \
 	services/stellaris.xml \
 	services/stronghold-crusader.xml \
+	services/submission.xml \
 	services/supertuxkart.xml \
 	services/svdrp.xml \
 	services/svn.xml \
--- a/config/services/smtp-submission.xml
+++ b/config/services/smtp-submission.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <service>
   <short>Mail (SMTP-Submission)</short>
-  <description>SMTP-Submission allows remote users to submit mail over port 587.</description>
-  <port protocol="tcp" port="587"/>
+  <description>This service is deprecated. Please use the "submission" service.</description>
+  <include service="submission"/>
 </service>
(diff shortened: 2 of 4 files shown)


erig0:
Hrm. Is UDP really necessary? I don't think so. In which case you could use
```
<include service="smtp-submission"/>
```
in the XML as this is really just an alias to `smtp-submission`.

viniciusferrao:
> Hrm. Is UDP really necessary? I don't think so. In which case you could use
> 
> ```
> <include service="smtp-submission"/>
> ```

Yes I never used 587/udp for Submission, however it is reserved on IANA so I think we should follow their reservation: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=587

> 
> in the XML as this is really just an alias to `smtp-submission`.

There's differences on the text of the service because, at least in what I think, the text for `smtp-submission` is not that great and explanatory.

And since I would also consider that `smtp-submssion` should be deprecated on the future I'm proposing it in full form, instead of becoming an alias.

erig0:
> And since I would also consider that `smtp-submssion` should be deprecated on the future I'm proposing it in full form, instead of becoming an alias.

Then we could mark `smtp-submission` as deprecated and alias it to this new service via `include`.

erig0:
> > And since I would also consider that `smtp-submssion` should be deprecated on the future I'm proposing it in full form, instead of becoming an alias.
> 
> Then we could mark `smtp-submission` as deprecated (in the description) and alias it to this new service via `include`.

I pushed a change that does this and drops the UDP port. Please take a look.

Issue #1163, opened 2023-07-10
Missing support for Steam's "Game File Transfer over Local Network"


StefanBruens:
Current versions of the Steam Client support downloading game files from a device in the local network, instead of downloading everything from the remote Steam servers.

For this to work, the device supplying the files must have TCP port 27040 opened.

StefanBruens:
[steam-filetransfer.xml.txt](https://github.com/firewalld/firewalld/files/12008578/steam-filetransfer.xml.txt)

devurandom:
> [steam-filetransfer.xml.txt](https://github.com/firewalld/firewalld/files/12008578/steam-filetransfer.xml.txt)

Confirmed: Loading this definition into firewalld on the *receiving* side (I already had a manually created `nft` rule on the sending side, i.e. the machine that already had the files) allows file transfers over the local network.

```
# firewall-cmd --permanent --new-service-from-file=steam-filetransfer.xml
# firewall-cmd --permanent --add-service=steam-filetransfer
# firewall-cmd --reload
```

erig0:
The steam support forums [1] say:

> The network transfer happens on TCP port 27040 and it needs to be allowed by local firewall software. You also need to open UDP ports 27031-27036 for client discovery.

We can definitely add a service with `27040/tcp`. The question is, do we need to also allow `27031-27036/udp`?

[1]: https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43

devurandom:
Yes, according to the docs those are also needed for discovery. But they are also included in https://github.com/firewalld/firewalld/blob/385fbbf0edc2626a812385abd15c965a3714eff4/config/services/steam-streaming.xml. Would that be a problem, two services containing the same ports?

erig0:
No. Not a problem.

devurandom:
> The steam support forums [1](https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43) say:
> 
> > The network transfer happens on TCP port 27040 and it needs to be allowed by local firewall software. You also need to open UDP ports 27031-27036 for client discovery.
> 
> We can definitely add a service with `27040/tcp`. The question is, do we need to also allow `27031-27036/udp`?

@koloved [tested this](https://github.com/firewalld/firewalld/pull/1216#issuecomment-2143409846): Yes, these ports are also needed.

erig0:
Fixed via #1216

PR #1216, opened 2023-09-25
Added Steam LAN transfer service

DanMan:
Implements #1163 

https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43

https://github.com/ValveSoftware/steam-for-linux/issues/9938

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -328,6 +328,7 @@ CONFIG_FILES = \
 	services/ssh.xml \
 	services/statsrv.xml \
 	services/steam-streaming.xml \
+	services/steam-lan-transfer.xml \
 	services/stellaris.xml \
 	services/stronghold-crusader.xml \
 	services/submission.xml \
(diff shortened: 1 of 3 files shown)


brinkabeler (approved):
seems legit

erig0:
Hi, the steam support docs [1] say this:

> The network transfer happens on TCP port 27040 and it needs to be allowed by local firewall software. You also need to open UDP ports 27031-27036 for client discovery.

Maybe we also need the UDP ports?

[1]: https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43

erig0:
Before this can merge we needinfo about the above question.

> Hi, the steam support docs [1](https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43) say this:
>
> > The network transfer happens on TCP port 27040 and it needs to be allowed by local firewall software. You also need to open UDP ports 27031-27036 for client discovery.
>
> Maybe we also need the UDP ports?

koloved:
> Before this can merge we needinfo about the above question.

i tried download game without the UDP 27031-27036, its not working, then turns its on and its working, tried two times

koloved:
@DanMan can you add that ports please ?

DanMan:
The same UDP port range is already part of the Steam Streaming config. Does it still need to be included here?

erig0 (approved):

PR #1233, opened 2023-11-07
Performance: nftables reload

jrchamp:
Preface: I'm not a Python expert, so I'm hoping you can test to make sure my suggestions are not breaking something. My tests produced identical JSON output, but my tests are not exhaustive.

### Speed up set_rules() using marshal
This one instance of deepcopy() uses 85% of the processing time and deepcopy() is about 15 times slows than marshal in my tests.

### Other potential speedups
Switch `_deduplicated_rules` from a list to a dict and use the rule_key as the key. Avoids another call to `_get_rule_key()` and avoids duplicating rules.
```diff
-        for rule in _deduplicated_rules:
+        for rule_key in _deduplicated_rules:
             index += 1  # +1 due to metainfo
-            rule_key = self._get_rule_key(rule)
+            rule = _deduplicated_rules[rule_key]
```

Pre-calculate ipset information and pass it through build_set_add_rules()
```diff
-    def _set_entry_fragment(self, name, entry):
+    def _set_entry_fragment(self, obj, type_format, entry):
         # convert something like
         #    1.2.3.4,sctp:8080 (type hash:ip,port)
         # to
         #    ["1.2.3.4", "sctp", "8080"]
-        obj = self._fw.ipset.get_ipset(name)
-        type_format = obj.type.split(":")[1].split(",")
         entry_tokens = entry.split(",")
         if len(type_format) != len(entry_tokens):
```
```diff
+        ipset = self._fw.ipset.get_ipset(set_name)
+        type_format = obj.type.split(":")[1].split(",")
+
         # avoid large memory usage by chunking the entries
         chunk = 0
         for entry in entries:
-            rules.extend(self.build_set_add_rules(set_name, entry))
+            rules.extend(self.build_set_add_rules(ipset, type_format, entry))
```

Move the marshal/deepcopy outside of the rules loop (I'm not as confident about this one, because I don't fully understand the need for a deepcopy in the first place if we're not reusing the objects)
```diff
-        for rule in rules:
+        _rules = marshal.loads(marshal.dumps(rules))
+        for _rule in _rules:
```

If you would like any of these additional changes added to this pull request, please let me know or feel free to add the edits directly to my branch.

### DIFF
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -333,13 +333,9 @@ class nftables:
     def _get_rule_key(self, rule):
         for verb in ["add", "insert", "delete"]:
             if verb in rule and "rule" in rule[verb]:
-                rule_key = copy.deepcopy(rule[verb]["rule"])
-                for non_key in ["index", "handle", "position"]:
-                    if non_key in rule_key:
-                        del rule_key[non_key]
                 # str(rule_key) is insufficient because dictionary order is
                 # not stable.. so abuse the JSON library
-                rule_key = json.dumps(rule_key, sort_keys=True)
+                rule_key = json.dumps(rule[verb]["rule"], sort_keys=True)
                 return rule_key
         # Not a rule (it's a table, chain, etc)
         return None
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -348,6 +348,7 @@ class nftables:
         _valid_verbs = ["add", "insert", "delete", "flush", "replace"]
         _valid_add_verbs = ["add", "insert", "replace"]
         _deduplicated_rules = []
+        _deduplicated_rules_keys = []
         _executed_rules = []
         rich_rule_priority_counts = copy.deepcopy(self.rich_rule_priority_counts)
         policy_dispatch_index_cache = copy.deepcopy(self.policy_dispatch_index_cache)
@@ -407,8 +408,9 @@ class nftables:
                 )
 
             _deduplicated_rules.append(rule)
+            _deduplicated_rules_keys.append(rule_key)
 
-            _rule = copy.deepcopy(rule)
+            _rule = rule
             if rule_key:
                 # filter empty rule expressions. Rich rules add quite a bit of
                 # them, but it makes the rest of the code simpler. libnftables
@@ -464,8 +466,8 @@ class nftables:
 
         index = 0
         for rule in _deduplicated_rules:
+            rule_key = _deduplicated_rules_keys[index]
             index += 1  # +1 due to metainfo
-            rule_key = self._get_rule_key(rule)
 
             if not rule_key:
                 continue
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -410,38 +410,37 @@ class nftables:
             _deduplicated_rules.append(rule)
             _deduplicated_rules_keys.append(rule_key)
 
-            _rule = rule
             if rule_key:
                 # filter empty rule expressions. Rich rules add quite a bit of
                 # them, but it makes the rest of the code simpler. libnftables
                 # does not tolerate them.
-                _rule[verb]["rule"]["expr"] = list(
-                    filter(None, _rule[verb]["rule"]["expr"])
+                rule[verb]["rule"]["expr"] = list(
+                    filter(None, rule[verb]["rule"]["expr"])
                 )
 
                 if self._fw._nftables_counters:
                     # -1 inserts just before the verdict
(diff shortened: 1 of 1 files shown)


erig0:
It's interesting that `deepcopy()` is so slow; possibly because it has to consider recursive references. I wonder if a custom version of `deepcopy()` for rules would be even faster than `marshal`. In the nftables backend rules are a dictionary of dictionaries/lists/strings/ints.

That being said, maybe `marshal` is good enough.

---

As for the other suggestions.

1. Use `_deduplicated_rules` instead of `_get_rule_key()`

This does not work because hashing a dictionary is not stable. The keys may hash in arbitrary order. This is why we're using `json_dumps(..., sort_keys=True)`

2. Pre-calculate ipset information and pass it through build_set_add_rules()

This should be fine. Note that there is an ipset scale test in the testsuite that uses ~65k entries. So I'm not sure if there is a major bottleneck here.

3. Move the marshal/deepcopy outside of the rules loop

This is possible, but would increase the peak memory usage (do it all at once) for less frequent allocations. Whether that's a good trade or not would need to be investigated.

erig0:
> Can you investigate a custom deepcopy() ?

I pushed a change that does just that. Please take a look. I'm curious how it profiles for you vs `marshal`.

jrchamp:
> Please take a look. I'm curious how it profiles for you vs `marshal`.

Unmodified: 41503786 function calls (34903759 primitive calls) in 12.009 seconds
`_deepcopy_rule`: 16003660 function calls (11503648 primitive calls) in 4.861 seconds
`marshal`: 4003615 function calls in 2.312 seconds

A custom deepcopy is much more efficient than the original deepcopy, but still seems slower than marshal. However, I'm not sure we need deepcopy or marshal for this part.

Here's my idea:

1. The `rules` that are passed in to `set_rules()` are completely within our control. No other code is going to be using the objects in `rules`.
2. If that's true, then the only code that receives the unmodified `rule` is the `_deduplicated_rules` list. Everything else is using the modified deepcopy `_rule`.
3. `_deduplicated_rules` only uses the unmodified `rule` for two things:
   1. To get the `rule_key` by using `_get_rule_key`
   2. To detect if the verb `"delete"` is part of the rule
4. If we only need those two pieces of information within `_deduplicated_rules`, we can store those instead of using `deepcopy`

No copy: 3103606 function calls in 1.315 seconds
```diff
diff --git a/nftables.py b/nftables.py
--- a/nftables.py
+++ b/nftables.py
@@ -406,9 +406,9 @@ class nftables:
                     f"rule ref count bug, missing ref count: rule_key '{rule_key}'",
                 )

-            _deduplicated_rules.append(rule)
+            _deduplicated_rules.append({"rule_key": rule_key, "rule": rule})

-            _rule = copy.deepcopy(rule)
+            _rule = rule
             if rule_key:
                 # filter empty rule expressions. Rich rules add quite a bit of
                 # them, but it makes the rest of the code simpler. libnftables
@@ -463,9 +463,10 @@ class nftables:
         self.rule_ref_count = rule_ref_count

         index = 0
-        for rule in _deduplicated_rules:
+        for rule_item in _deduplicated_rules:
             index += 1  # +1 due to metainfo
-            rule_key = self._get_rule_key(rule)
+            rule = rule_item["rule"]
+            rule_key = rule_item["rule_key"]

             if not rule_key:
                 continue
```

erig0:
I pushed another update that completely avoids `deepcopy()` for rules. Please take another look. :)

src/firewall/core/nftables.py
```diff
@@ -465,7 +463,7 @@ def set_rules(self, rules, log_denied):
         index = 0
         for rule in _deduplicated_rules:
             index += 1  # +1 due to metainfo
-            rule_key = self._get_rule_key(rule)
+            rule_key = _deduplicated_rules_keys[index - 1]
```
jrchamp, line 466: Would this be easier to read?
```diff
+            rule_key = _deduplicated_rules_keys[index]
             index += 1  # +1 due to metainfo
-            rule_key = _deduplicated_rules_keys[index - 1]
```

jrchamp (commented):
This looks great! I had one idea, but it's completely optional.

src/firewall/core/nftables.py
erig0: Yes. Thanks for catching.

erig0:
I'll merge this, can you give me an name + email for a `Co-authored-by` tag for the first commit?

jrchamp:
[Jonathan Champ &lt;jrchamp@ncsu.edu&gt;](https://delta.ncsu.edu/about-delta/people/name/jonathan-champ/)

erig0 (approved):

PR #1238, opened 2023-11-13
improvement(build): add ".git-blame-ignore-revs" file for "blame.ignoreRevsFile" git config

thom311:
If you want to use git-blame and are bothered by the large diff due to the reformatting with python-black, consider ignoring the reformatting commit by using ".git-blame-ignore-revs" file.

---

this is a follow-up to the reformatting with python black.

The `.git-blame-ignore-revs` file contains a comment that explains how to use it (or its purpose).


erig0 (approved):
lgtm. Bummer git doesn't parse this by default.

PR #1244, opened 2023-11-18
clean up the created nmclient when the reload is complete

lbo874:
After the reload command is executed, a large number of routes are added, which increases the memory usage of firewalld. This may not be caused by firewalld. However, firewalld is affected by this. It is recommended that the nmclient be released after the reload is complete.

### DIFF
--- a/src/firewall/core/fw_nm.py
+++ b/src/firewall/core/fw_nm.py
@@ -22,6 +22,7 @@ else:
     except (ImportError, ValueError, GLib.Error):
         _nm_imported = False
 _nm_client = None
+_nm_client_timeout = None
 
 from firewall import errors
 from firewall.errors import FirewallError
@@ -46,9 +47,28 @@ def nm_get_client():
     """Returns the NM client object or None if the import of NM failed
     @return NM.Client instance if import was successful, None otherwise
     """
+
+    def _release():
+        """
+        Release the client to avoid excess memory usage when libnm pushes
+        irrelevant (to firewalld) updates.
+        """
+        global _nm_client
+        global _nm_client_timeout
+        _nm_client = None
+        _nm_client_timeout = None
+
     global _nm_client
+    global _nm_client_timeout
+
     if not _nm_client:
         _nm_client = NM.Client.new(None)
+    else:
+        # refresh timer
+        GLib.source_remove(_nm_client_timeout)
+
+    _nm_client_timeout = GLib.timeout_add_seconds(5, _release)
+
     return _nm_client
 
 


thom311:
could you fix the formatting, by running `black .` ? (using python-black)

Also, could you wrap lines in the commit message?


Otherwise, patch lgtm.

lbo874:
> could you fix the formatting, by running `black .` ? (using python-black)
> 
> Also, could you wrap lines in the commit message?
> 
> Otherwise, patch lgtm.

Modified, thank you for your advice

lbo874:
@erig0

erig0:
This fixes the `--reload` case, but the daemon also interacts with NM when `--runtime-to-permanent` is called. Furthermore, I expect this issue affects long lived GUIs, e.g. `firewall-config`. I suspect the better "fix" is to handle the libnm connect/disconnect in `src/firewall/core/fw_nm.py`. This would trade the overhead of re-establishing libnm (plus dbus) for peak memory usage.

I think we need to investigate if these other areas are affected before we merge this.

thom311:
I think we could just attach an idle handler, which cleans the nm-client instance, like:

```diff
diff --git i/src/firewall/core/fw_nm.py w/src/firewall/core/fw_nm.py
index aec184c856dc..600bdb93463a 100644
--- i/src/firewall/core/fw_nm.py
+++ w/src/firewall/core/fw_nm.py
@@ -44,16 +44,22 @@ def nm_is_imported():
 
 def nm_get_client():
     """Returns the NM client object or None if the import of NM failed
     @return NM.Client instance if import was successful, None otherwise
     """
     global _nm_client
     if not _nm_client:
         _nm_client = NM.Client.new(None)
+
+        def release():
+            global _nm_client
+            _nm_client = None
+
+        GLib.idle_add(release)
     return _nm_client
 
 
 def nm_get_zone_of_connection(connection):
     """Get zone of connection from NM
     @param connection name
     @return zone string setting of connection, empty string if not set, None if connection is unknown
     """
```

erig0:
> I think we could just attach an idle handler, which cleans the nm-client instance, like:

I also thought about that, but it breaks UIs that rely on NM dbus signals. See `nm_signal_receiever()` in `firewall-config` and `firewall-applet`.

thom311:
> Edit: Actually, listening to the signal does not depend on libnm. As such, maybe this would be fine. I'll try it out.

Right... except, if we get a lot of `PropertiesChanged` signals, each time we will (synchronously) create a `NMClient` instance (which is time consuming). Of course, if we get a lot of signals, we may already have a problem in this code...

src/firewall/core/fw_nm.py
```diff
@@ -49,6 +50,22 @@ def nm_get_client():
...
+            irrelevant (to firewalld) updates.
+            """
+            global _nm_client
+            global _nm_client_timeout
+            _nm_client = None
+            _nm_client_timeout = None
+
+        global _nm_client_timeout
+        if _nm_client_timeout:
+            GLib.source_remove(_nm_client_timeout)
```
thom311, line 66: I think `if _nm_client_timeout` cannot be True here.

The only question is, whether we want to restart the timer with full 5 seconds, each time when `nm_get_client()` is called. Probably not, otherwise can just just move these few lines after the `if not _nm_client` block.

erig0: rewrote, please review again.

src/firewall/core/fw_nm.py
```diff
@@ -47,8 +48,26 @@ def nm_get_client():
...
+        """
+        global _nm_client
+        global _nm_client_timeout
+        _nm_client = None
+        _nm_client_timeout = None
+
     if not _nm_client:
         _nm_client = NM.Client.new(None)
+        _nm_client_timeout = GLib.timeout_add_seconds(5, _release)
+    elif _nm_client_timeout:
```
thom311, line 66: I think it still holds that at this place `_nm_client_timeout` cannot be zero.

But of course, it doesn't hurt much.

Patch lgtm.

erig0: Why? Wouldn't repeated calls to `nm_get_client()` hit this `else`?

erig0: Erm, I see what you mean. I had already pushed an `s/elif/else/` before I saw you comment. ;)

PR #1250, opened 2023-11-29
[th/icmpv6-mdl] add support for ICMPv6 Multicast Listener Discovery (MDL) types (nftables backend only)

thom311:
- https://issues.redhat.com/browse/RHEL-13252
- https://access.redhat.com/solutions/7011275
- https://www.rfc-editor.org/rfc/rfc4890

This only adds support for 4 new ICMPv6 types:
- "mld-listener-done"
- "mld-listener-query"
- "mld-listener-report"
- "mld2-listener-report"

I think these are only supported by nftables, so no ip6tables backend.

There is still the question, whether any of these ICMP types should be allowed by default. This branch only adds support for them. It's basically, about whether I am in the right direction, future work may address this.

also add some tests first. I used that to better unstand the code. It makes it also harder to accidentally add a new ICMP type, and forget adjusting some pieces.


src/tests/unit/test_icmp.py
erig0, line 1: It's odd to have unit tests load the XML files. Is there a reason you chose this way over a test case in autotest testsuite that exercises all the ICMP types?

src/tests/unit/test_icmp.py
```diff
@@ -312,3 +318,110 @@ def _iptables_parse_supported_icmp_types_6():
...
+            l = [xmlobj for xmlobj in xmlobjs if xmlobj.name == icmp_type]
+            assert len(l) == 1
+            xmlobj = l[0]
+            assert ipx in _get_destination(xmlobj)
+
+    # firewall.core.icmp is only used by src/firewall/core/io/ipset.py, and
+    # hard-codes valid ICMP types. It's not used anywhere else.
+    #
+    # It should still correspond to our XML files, where every entry here has
+    # an XML file, but some XML files are not found here.
```
erig0, line 392: I didn't even know we supported ICMP in the `port` field of an ipset. 🤷

src/firewall/core/nftables.py
```diff
@@ -212,6 +212,10 @@ def _icmp_types_fragments(protocol, type, code=None):
         "ttl-zero-during-transit": _icmp_types_fragments("icmpv6", "time-exceeded", 0),
         "unknown-header-type": _icmp_types_fragments("icmpv6", "parameter-problem", 1),
         "unknown-option": _icmp_types_fragments("icmpv6", "parameter-problem", 2),
+        "mld-listener-query": _icmp_types_fragments("icmpv6", "mld-listener-query"),
+        "mld-listener-report": _icmp_types_fragments("icmpv6", "mld-listener-reprot"),
+        "mld-listener-done": _icmp_types_fragments("icmpv6", "mld-listener-done"),
+        "mld2-listener-report": _icmp_types_fragments("icmpv6", "mld2-listener-report"),
```
erig0, line 218: nit: please add them sorted in the list.

erig0 (commented):
This mostly looks fine.

Re: enabling them by default
If ip6tables doesn't support these types, then I'm not sure we can enable it by default. As it would fail with FirewallBackend=iptables

src/tests/unit/test_icmp.py
thom311: > Is there a reason you chose this way over a test case in autotest testsuite that exercises all the ICMP types?

The reason is that unit tests can check for things that would be hard/impossible otherwise. "Unit tests" in the sense, that they can directly call into the Python code in question, instead of running firewalld daemon and exercise some code paths via its API.

autotest suite is usually more high level, running all of firewalld, with `unshare`. Surely, all of those tests will load the XMLs too. Of course, you could also have an autotest test, that only runs `python -c ...`, but such tests already exists, and are our `pytest` based unit tests. So, I wouldn't add a unit test via autotest.

There is no conflict/overlap, between having a unit test and an integration test.

What the unit test does, is to check that the code `firewall.core.io.icmptype.icmptype_reader()` passes, with all the XML files we have. Then, it performs some sanity checks on the loaded objects. Those checks are not done, if you just run `firewalld` daemon.

The directory layout of the project is well-known and well-defined. The unit test is allowed to know that the icmptypes XMLs are at a certain place -- relative to itself. Those XML files are anyway part of the installation and static (that is, commited to git, in a fixed state). No, this does not seem an antipattern to me.

src/tests/unit/test_icmp.py
thom311: https://github.com/firewalld/firewalld/pull/1172 touches that area too (commit 2d1f4b3d2c9842cde5b8fb6c25c4e4b7363311d1)

src/firewall/core/nftables.py
thom311: done.

erig0 (changes requested):
Please fix the CI issues. Commit message introducing the MLD types should be `feat` not `improvement`.

Otherwise, lgtm.

src/tests/unit/test_icmp.py
thom311: > autotest has the advantage that it can test the running/installed firewalld

Well, it's a unit test. It's gonna be harder to run, on the complete installation (because the individual units are not easily accessible). Btw, if that were an interesting usage, it's of course possible (just adjust the PYTHONPATH to load the installed sources instead).

> autotest also verifies the nftables/iptables backends actually support those ICMP types.

Yes, it's a unit test. That does not replace integration-tests (or whatever we call that). Nor is that the point, rather the contrary. The point is that it does *not* test the entire system down to the backends.

> Most of what they are testing could be covered by autotest as well, e.g. --add-icmp-type for every --get-icmp-types. The only thing not covered by that is

You say "Most...could be covered ... The only thing not covered by..."

The thing(*) that could be covered is *also* not covered exhaustively for all icmp types.

(*) i.e. a test that checks that every icmptypes from XML files can be configured with nftables backend.

---

Anyway, there is no conflict between having a unit test testing something very specific, vs. having another test that does end-to-end (running firewalld, see that icmptypes files are loaded, configure icmptypes in zones, activate the zone, check for expected nftables rules). Optimally, we have them all.

erig0: > > autotest has the advantage that it can test the running/installed firewalld
> 
> Well, it's a unit test.

Only because you made it a unit test. You could have made it an end to end test.

erig0 (dismissed):
CI still unhappy because it needs python3-nftables, which is not installed during the source code checks.

Edit: I noticed your top commit has a typo.
`feat(icmp): add ICMPv6 Multicast Listener Discovery (MDL) types`. s/MDL/MLD/

erig0 (dismissed):

erig0:
Before I merge this.. to support ip6tables, couldn't we use `--icmpv6-type type/code` ?

In theory we could fully move to that and thus support the iptables/nftables backends in the same way. I don't think we have to use the "named" icmp-types on the iptables backend.

This could be a follow up series. It would remove the dependency on parsing the iptables `--help` output.

thom311:
> Before I merge this.. to support ip6tables, couldn't we use --protocol ipv6-icmp -m icmp6 --icmpv6-type type/code ?

Yes, that's a good plan.

But it anyway would a follow-up to this merge request. Can we merge this?

erig0:
> But it anyway would a follow-up to this merge request. Can we merge this?

Yes.

erig0 (dismissed):

erig0 (dismissed):

erig0 (approved):

PR #1255, opened 2023-12-05
[th/unittest-richrule] add unit tests for parsing rich rules

thom311:
dito

### DIFF
--- a/src/firewall/core/rich.py
+++ b/src/firewall/core/rich.py
@@ -33,6 +33,10 @@ class _Rich_EntryWithLimit(_Rich_Entry):
     def __init__(self, limit=None):
         self.limit = limit
 
+    def check(self, family=None):
+        if self.limit is not None:
+            self.limit.check(family=family)
+
 
 class _Rich_Element(_Rich_Entry):
     pass
@@ -73,6 +77,27 @@ class Rich_Source(_Rich_Entry):
             return ret + 'ipset="%s"' % self.ipset
         raise FirewallError(errors.INVALID_RULE, "no address, mac and ipset")
 
+    def check(self, family=None):
+        if self.addr is not None:
+            if family is None:
+                raise FirewallError(errors.INVALID_FAMILY)
+            if self.mac is not None:
+                raise FirewallError(errors.INVALID_RULE, "address and mac")
+            if self.ipset is not None:
+                raise FirewallError(errors.INVALID_RULE, "address and ipset")
+            if not functions.check_address(family, self.addr):
+                raise FirewallError(errors.INVALID_ADDR, str(self.addr))
+        elif self.mac is not None:
+            if self.ipset is not None:
+                raise FirewallError(errors.INVALID_RULE, "mac and ipset")
+            if not functions.check_mac(self.mac):
+                raise FirewallError(errors.INVALID_MAC, str(self.mac))
+        elif self.ipset is not None:
+            if not check_ipset_name(self.ipset):
+                raise FirewallError(errors.INVALID_IPSET, str(self.ipset))
+        else:
+            raise FirewallError(errors.INVALID_RULE, "invalid source")
+
 
 class Rich_Destination(_Rich_Entry):
     def __init__(self, addr, ipset, invert=False):
@@ -93,6 +118,20 @@ class Rich_Destination(_Rich_Entry):
         elif self.ipset is not None:
             return ret + 'ipset="%s"' % self.ipset
         raise FirewallError(errors.INVALID_RULE, "no address and ipset")
+
+    def check(self, family=None):
+        if self.addr is not None:
+            if family is None:
+                raise FirewallError(errors.INVALID_FAMILY)
+            if self.ipset is not None:
+                raise FirewallError(errors.INVALID_DESTINATION, "address and ipset")
+            if not functions.check_address(family, self.addr):
+                raise FirewallError(errors.INVALID_ADDR, str(self.addr))
+        elif self.ipset is not None:
+            if not check_ipset_name(self.ipset):
+                raise FirewallError(errors.INVALID_IPSET, str(self.ipset))
+        else:
+            raise FirewallError(errors.INVALID_RULE, "invalid destination")
 
 
 class Rich_Service(_Rich_Element):
@@ -214,8 +253,7 @@ class Rich_Log(_Rich_Log):
         ]:
             raise FirewallError(errors.INVALID_LOG_LEVEL, self.level)
 
-        if self.limit is not None:
-            self.limit.check(family=family)
+        super().check(family=family)
 
 
 class Rich_NFLog(_Rich_Log):
@@ -251,8 +289,7 @@ class Rich_NFLog(_Rich_Log):
                 "nflog 'queue-size' must be an integer between 0 and 65535.",
             )
 
-        if self.limit is not None:
-            self.limit.check(family=family)
+        super().check(family=family)
 
 
 class Rich_Audit(_Rich_EntryWithLimit):
@@ -297,6 +334,8 @@ class Rich_Reject(_Rich_Action):
                     "Wrong reject type %s.\nUse one of: %s." % (self.type, valid_types),
                 )
 
+        super().check(family=family)
+
 
 class Rich_Drop(_Rich_Action):
     def __init__(self, limit=None):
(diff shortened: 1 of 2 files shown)


src/firewall/core/rich.py
```diff
@@ -13,7 +13,8 @@
 
 
 class _Rich_Entry:
-    pass
+    def check(self, family=None):
+        pass
```
erig0, line 17: I don't think we need to pass `family` here. AFAICS, the only `check()` using it is `Rich_Reject`, but I don't see any code actually passing family to the constructor.

src/firewall/core/rich.py
```diff
@@ -236,8 +278,7 @@ def check(self, family=None):
                 "nflog 'queue-size' must be an integer between 0 and 65535.",
             )
 
-        if self.limit is not None:
-            self.limit.check(family=family)
+        super().check(family=family)
 
 
 class Rich_Audit(_Rich_Entry):
```
erig0, line 333: `audit` is a type of log. This could derive `_Rich_Log`.

src/firewall/core/rich.py
```diff
@@ -778,6 +827,10 @@ def _import_from_string(self, rule_str):
 
             index = index + 1
 
+    def _check_entry(self, entry):
+        if entry is not None:
+            entry.check(family=self.family)
+
```
erig0: IMO it's not useful to abstract the if statement for a "null check". Just do it at the call sites.

e.g.
```python
if self.source is not None:
  self.source.check()
```

src/firewall/core/rich.py
```diff
@@ -871,50 +872,43 @@ def check(self):
         self._check_entry(self.source)
         self._check_entry(self.destination)
 
-        # service
         if isinstance(self.element, Rich_Service):
+            # service
```
erig0: Just remove the comment entirely. It's clear from the object name this is a `Rich_Service`. The comment is not providing any value.

erig0 (changes requested):
Minor suggestions. For the most part this looks good. Thanks!

src/firewall/functions.py
```diff
@@ -670,3 +670,15 @@ def wrong_args_for_callable(fcn, *a, **kw):
...
+# Dummy class for EOL singleton instance. It's a class, so we
+# can overwrite __repr__().
+class _EOLType:
+    def __repr__(self):
+        # The string representation is the full name of the instance.
+        return "firewall.functions.EOL"
+
+
+# A EndOfLine token, used by tokenizer.
+EOL = _EOLType()
```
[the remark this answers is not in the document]
thom311: I'd rather not. `firewall.functions.EOL` is a general thing. It's a token that indicates EOL, it can be used for any kind of parsing.

For good reasons, `firewall.functions` provides serveral API which is only used at one place:
- `firewall.functions.stripNonPrintableCharacters` only in `rich.py`
- `firewall.functions.enable_ip_forwarding` only in `fw_policy.py`.
- `firewall.functions.firewalld_is_active` only in `firewalld.in`
- `firewall.functions.get_nf_conntrack_short_name` only in `fw_policy.py`

Whether something is in `firewall.functions` should depend on whether it's a self-contained concept, that makes sense on it's own (independent from the actual users).

For example, `EOL` is so trival that it doesn't have any tests. But if it were not as trivial, we should add tests for it. And for testing it, `rich.py` is not relevant.

src/firewall/core/rich.py
```diff
@@ -372,6 +368,19 @@ def __init__(self, family=None, rule_str=None, priority=0):
...
+            self.priority = priority
+
+        if family is not None:
+            family = str(family)
+            if self.family is None:
+                self.family = family
+            elif self.family != family:
+                raise FirewallError(errors.INVALID_FAMILY, family)
+
+        if rule_str is not None:
```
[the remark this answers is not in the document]
thom311: Then a plain
```python
rule = firewall.core.rich.Rich_Rule()
```
would fail the check and raise an exception.

There are two modes of operation. One creating a rich-rule from a string (in which we call `check()`). The other is to create an empty rich-rule instance, and programmatically set the values. In the latter case, calling `check()` is pointless but also won't work, because the instance is not yet full initialized nor valid.

src/firewall/core/rich.py
thom311: Various "entry" types honor the family during check. For example, `Rich_Source.check()`, `Rich_Destination.check()`, `Rich_ForwardPort.check()`, `Rich_Reject.check()`. 

The relevant part is, while not all of them honor a family, some do. Consequently, for `_Rich_Entry.check()` the family must be an argument, to perform the necessary validation.

> but I don't see any code actually passing family to the constructor.

"constructor"? The `_Rich_Entry` instances don't actually have an address family property. Instead, the `Rich_Rule` has a family, and it passes it down to the `check()` method.

src/firewall/functions.py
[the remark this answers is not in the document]
erig0: I still disagree. We'll never use this for other things. If in the future we do, then sure; lets move it to a common area.

src/firewall/core/rich.py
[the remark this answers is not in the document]
erig0: ACK.

src/firewall/core/rich.py
thom311: I chose the base types based on how the entries are assigned to the rich_rule.

For example,
- `rich_rule.audit` property holds  instance derived from `_Rich_Audit`
- `rich_rule.log` property holds instance derived from `_Rich_Log`
- `rich_rule.action` property holds instance derived from `_Rich_Action`

Under that consideration, `Rich_Audit` "is not a" `_Rich_Log`.

Instead, I added `_Rich_EntryWithLimit`, which is now a base class of `Rich_Audit`, `_Rich_Action` and `_Rich_Log`.

src/firewall/core/rich.py
thom311: I'd rather not.

Your example is incorrect. The full replacement also requires to pass the family, so it's
```python
if self.source is not None:
    self.source.check(family=self.family)
```
which then needs to be copied at 6 places, and adjusted for `self.{source,destination,element,log,audit,action}`.

That redundancy and verbosity is avoided by `_check_entry()`.

src/firewall/core/rich.py
thom311: dropped

src/firewall/functions.py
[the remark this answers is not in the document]
thom311: moved. It's now `firewall.core.rich.EOL`.

src/firewall/core/rich.py
erig0: > "constructor"? The _Rich_Entry instances don't actually have an address family property. Instead, the Rich_Rule has a family, and it passes it down to the check() method.

Okay. I think we can improve this.

If `family` is important to the element, e.g. `Rich_Source`, then it should be known to the object, i.e. `self.family`. Or we need a reference to the containing `Rich_Rule` so the element knows it's own context.

Many of the validation checks are "was family specified?". Only a few do real checks based on the family.

e.g.
```
            if family is None:
                raise FirewallError(errors.INVALID_FAMILY)
```

erig0: Another example of why this would be useful.

This code is in `Rich_Rule` instead of `Rich_Audit` because it needs `self.action`.

```
        if self.audit is not None:
            if type(self.action) not in [Rich_Accept, Rich_Reject, Rich_Drop]:
                raise FirewallError(errors.INVALID_AUDIT_TYPE, type(self.action))
```

If we had a reference to the containing `Rich_Rule` we could have this code is `Rich_Audit.check()`.

```
        if type(self.parent.action) not in [Rich_Accept, Rich_Reject, Rich_Drop]:
                raise FirewallError(errors.INVALID_AUDIT_TYPE, type(self.parent.action))
```

`parent` may not be the best name.

src/firewall/core/rich.py
erig0: > Your example is incorrect.

Sure. I forgot passing `family` to `check()` because above I suggested removing it entirely. The point I was making is still stands.

> which then needs to be copied at 6 places, and adjusted for

It would only be copying the `if foo is not None:`. I don't care enough about this to discuss further. What you have is okay.

erig0 (approved):

PR #1261, opened 2023-12-12
[th/rich-rule-check-service] fix(rich): validate service name of rich rule

thom311:
Previously, validation of valid service names was not done. That meant:
```
  $ firewall-cmd --add-rich-rule='rule priority="-100" family="ipv4" source address="10.0.0.10" service name="listen" accept' --permanent
  success
  $ firewall-cmd --reload
  Error: INVALID_SERVICE: listen
```
which left firewalld in a bad state.

Now:
```
  $ firewall-cmd --add-rich-rule='rule priority="-100" family="ipv4" source address="10.0.0.10" service name="listen" accept' --permanent
  Error: INVALID_SERVICE: Zone 'public': 'listen' not among existing services
```
It's still not all golden. If you now have to happen an XML file with an unknown service on disk, on restart we get tracebacks and a degraded state.
```
  $ firewall-cmd --add-rich-rule='rule priority="-100" family="ipv4" source address="10.0.0.10" service name="listen" accept' --permanent
  Error: RUNNING_BUT_FAILED: Changing permanent configuration is not allowed while firewalld is in FAILED state. The permanent configuration must be fixed and then firewalld restarted. Try `firewall-offline-cmd --check-config`.
  $ firewall-offline-cmd --check-config
  Configuration error: INVALID_SERVICE: Zone 'public': 'listen' not among existing services
```
But that is a general problem, which probably should be handled better. For example, by not failing badly and making the best of what we have. This while possibly loosing configuration by normalizing errors away, but still exposing the configuration problem to the user. Anyway, this is not new. What is new is that this error condition now triggers the failure.

https://issues.redhat.com/browse/RHEL-5790

### DIFF
--- a/src/firewall/core/io/policy.py
+++ b/src/firewall/core/io/policy.py
@@ -570,6 +570,14 @@ def common_check_config(obj, config, item, all_config, all_io_objects):
                             log.debug1("{} (unsupported)".format(ex))
                         else:
                             raise ex
+            elif isinstance(obj_rich.element, rich.Rich_Service):
+                if obj_rich.element.name not in all_io_objects["services"]:
+                    raise FirewallError(
+                        errors.INVALID_SERVICE,
+                        "{} '{}': '{}' not among existing services".format(
+                            obj_type, obj.name, obj_rich.element.name
+                        ),
+                    )
 
 
 def common_writer(obj, handler):
--- a/src/tests/features/rich_rules.at
+++ b/src/tests/features/rich_rules.at
@@ -46,6 +46,10 @@ FWD_CHECK([--permanent --policy foobar --add-rich-rule='rule family=ipv4 priorit
 FWD_CHECK([--permanent --policy foobar --add-rich-rule='rule family=ipv4 priority=0  source address=10.10.10.13 drop'], 0, ignore)
 FWD_CHECK([--permanent --policy foobar --add-rich-rule='rule family=ipv4 priority=-1 source address=10.10.10.14 accept'], 0, ignore)
 FWD_CHECK([--permanent --policy foobar --add-rich-rule='rule family=ipv4 priority=1  source address=10.10.10.15 accept'], 0, ignore)
+
+dnl Invalid service name is rejected.
+FWD_CHECK([--permanent --policy foobar --add-rich-rule='rule priority="-100" family="ipv4" source address="10.0.0.10" service name="bogusservice" accept'], 101, ignore, ignore)
+
 FWD_RELOAD
 NFT_LIST_RULES([inet], [filter_IN_policy_foobar_pre], 0, [dnl
     table inet firewalld {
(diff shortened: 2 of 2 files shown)


src/firewall/core/io/policy.py
```diff
@@ -570,6 +570,16 @@ def common_check_config(obj, config, item, all_config, all_io_objects):
                             log.debug1("{} (unsupported)".format(ex))
                         else:
                             raise ex
+            elif isinstance(obj_rich.element, rich.Rich_Service):
+                if "services" in all_io_objects:
```
erig0, line 574: nit: The `if` guard is not necessary. The `all_io_objects` dict will always have a "services" key. The `existing_services` temporary variable is also unnecessary, but okay if you think it improves readability.

src/tests/features/rich_rules.at
```diff
@@ -319,4 +323,4 @@ IP6TABLES_LIST_RULES([filter], [IN_foobar_post], 0, [dnl
     ACCEPT 0 -- ::/0 ::/0
 ])
 
-FWD_END_TEST([-e '/ERROR: INVALID_ZONE:/d'])
+FWD_END_TEST([-e '/ERROR: INVALID_ZONE:/d' -e "/ERROR: INVALID_SERVICE: Policy 'foobar': 'listen' not among existing services/d"])
```
erig0, line 326: You can avoid the long line by using `dnl`.
```m4
FWD_END_TEST([-e '/ERROR: INVALID_ZONE:/d' dnl
              -e "/ERROR: INVALID_SERVICE: Policy 'foobar': 'listen' not among existing services/d"])
```

erig0 (commented):
lgtm. A couple suggestions that aren't strictly required. Please take a look.

erig0:
> For example, by not failing badly and making the best of what we have.

Hard disagree. This is "limping along". If the configuration is invalid then we should complain loudly. Older firewalld would ignore the invalid rule and move on. That leaves the user with only part of their desired rules applied. IMO that's much worse that falling back to the default configuration and complaining.

src/firewall/core/io/policy.py
thom311: I was thinking so too and wanted to avoid the check, but the block above has a similar check.

If I can rely on that, all the better. Will change...

src/tests/features/rich_rules.at
thom311: will change.

thom311:
>  If the configuration is invalid then we should complain loudly.

The effect is, that firewalld becomes unusable and requires the user to fix XML files (which requires different permissions than the D-Bus permissions). It means, a user can get into a situation by using D-Bus, from which they cannot recover without resorting to editing files. Even if they only get there due to a bug of some sort, bugs happen and are not recoverable.

This may be under the assumption, that a detectable problem in XML is obviously invalid and the worst thing. But note that there are infinitely many ways to misconfigure the firewall, with similar bad effects (e.g. wrong port number). But those misconfigurations are not caught by the consistency check. Failing in the most painful way (and non-recoverable) is not helpful, when many subtle misconfigurations cannot be detected. The user MUST not misconfigure the firewall either way, and they are well advised to fix misconfigurations. Forcing them by bricking the system is not best.

Anyway. The branch changes nothing about that. Except a comment in the commit message how this bug can be used to brick firewalld (by first using an old version, and upgrading to a fixed version). I'll reword that paragraph.

erig0:
Most of what you're describing is only possible if a user directly edits the XML. Don't do that. Use the CLI.

erig0 (approved):
Thanks!

PR #1262, opened 2023-12-13
[th/rich-rule-limit-burst] add support for "burst" argument to "limit" of rich rules

thom311:
For `iptables`, this is `-m limit --limit rate/suffix --limit-burst burst`.

For nftables, this is `limit rate [over] packet_number / TIME_UNIT [burst packet_number packets]`

https://issues.redhat.com/browse/RHEL-9316

not implemented in `src/firewall-config.in`.

This is currently on top of !1255, therefore it's still a draft.


thom311:
tests are also still missing... will be addressed, when the general direction is agreed.

src/firewall/core/rich.py
```diff
@@ -458,11 +460,19 @@ def check(self, family=None):
...
+        return rate, mult, duration
+
+    def value_normalized(self, check=True):
+        try:
+            rate, mult, duration = self.value_parse()
+        except FirewallError:
+            if check:
+                raise
+            return self.value
+        return f"{rate}/{duration}"
```
erig0, line 474: Why don't we just always make `self.value` normalized? I don't follow why we would not want that. I don't see the value in `value_normalized()` and `value_parse()` when we could just always guarantee that `foo.value` is normalized. If you need the rate the set `self.rate`.

src/firewall/core/rich.py
```diff
@@ -471,8 +473,35 @@ def value_normalized(self, check=True):
...
+        try:
+            burst = self.burst_parse()
+        except FirewallError:
+            if check:
+                raise
+            return self.burst
+        if burst is None:
+            return None
+        return f"{burst}"
+
```
erig0, line 499: Just set `self.burst` to a normalized value when the object is created. No need for these functions.

erig0 (changes requested):
I'll re-review after your other MR merges and you rebase.

src/firewall/core/rich.py
thom311: So far, the `_Rich_Entry` objects are plain, stateless data holders. They get initialized once, and are not mutated afterwards.

While `Rich_Rule._import_from_string()` necessarily does some parsing and validation, once it got the basic tokens it just creates the `Rich_Rule` and `_Rich_Entry` objects. The entries in there are mostly unvalidated yet (aside what was necessary while parsing the string). Also, the GUI constructs entries, but doesn't yet perform a full validation (that happens during `check()`).

a) It feels wrong, that the constructor of a `_Rich_Entry` would already perform validation. That is, `Rich_Limit(value='foo')` should not raise an exception, that happens later during `check()`. Also, if you set `Rich_Limit(value=' 1 /m')` it feels wrong to mutate that to `entry.value='1/m'` before calling `check()`.

b) these `_Rich_Entry` instances should be immutable after being created. Immutable things are great, it means that entries don't have an unchecked and checked state. Normalizing `self.value` e.g. during `Rich_Limit.check()` seems very wrong, because then the instance mutates.

So the normalization shouldn't happen in `__init__()` and it shouln't happen in `check()`. There is no other good place. Note that the amount of code is still the same. It's not that normalizing `self.value` would safe any code. It might safe multiple parsing, if you are concerned about the overhead of that...

src/firewall/core/rich.py
thom311: the need is to have stateless data holders, that can contain the input values of the caller (without normalization).

src/firewall/core/rich.py
erig0: In this PR you argue we should not be normalizing, but you also have PR #1290 that is all about normalizing configuration.

I'm pushing back on this PR. Just normalize when the object is created. I have not seen any reasoning or justification about the benefits of not normalizing when the object is created.

Even the standard library throws errors at object creation.

```
>>> import ipaddress.IPv4Address as IPv4Address
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'ipaddress.IPv4Address'; 'ipaddress' is not a package
>>> from ipaddress import IPv4Address
>>> IPv4Address('not an address')
  File "<stdin>", line 1, in <module>
  File "/Users/egarver/pkg-2023Q3/lib/python3.11/ipaddress.py", line 1315, in __init__
    self._ip = self._ip_int_from_string(addr_str)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/egarver/pkg-2023Q3/lib/python3.11/ipaddress.py", line 1202, in _ip_int_from_string
    raise AddressValueError("Expected 4 octets in %r" % ip_str)
ipaddress.AddressValueError: Expected 4 octets in 'not an address'
>>> IPv4Address('10.10.10.10')
IPv4Address('10.10.10.10')
```

thom311: > In this PR you argue we should not be normalizing, but you also have PR #1290 that is all about normalizing configuration.

This branch also did normalization. Except, it didn't do it in `Rich_Rule.__init__()`.

Changed now. The constructor normalizes the values.

>  Even the standard library throws errors at object creation.

I didn't say that `__init__()` never fails. However, in this case it seems not best, because there is a `check()` function that fails at a more appropriate moment. That was the case already before the patch, and it still is.

src/firewall/core/rich.py
thom311: changed.

src/firewall/core/rich.py
```diff
@@ -425,51 +425,70 @@ def check(self, family=None):
...
+    "s": 1,
+    "m": 60,
+    "h": 60 * 60,
+    "d": 24 * 60 * 60,
+}
+
+
 class Rich_Limit(_Rich_Entry):
     def __init__(self, value):
         self.value = value
```
erig0, line 438: Should this be `self._value = value` ?

src/firewall/core/rich.py
```diff
@@ -425,51 +425,70 @@ def check(self, family=None):
...
         if rate == 1 and duration == "d":
             # iptables (v1.4.21) doesn't accept 1/d
-            raise FirewallError(errors.INVALID_LIMIT, "%s too slow" % self.value)
+            raise FirewallError(errors.INVALID_LIMIT, "%s too slow" % (value,))
 
-    def __str__(self):
-        return 'limit value="%s"' % (self.value)
+        return rate, duration
+
+    def value_parse(self):
```
erig0, line 507: Why do we need both `value_parse()` and `_value_parse()` ? This wrapper doesn't seem helpful, just rename `_value_parse()` to `value_parse()`.

src/firewall/core/nftables.py
```diff
@@ -1594,15 +1594,12 @@ def _rich_rule_limit_fragment(self, limit):
             "d": "day",
         }
 
-        try:
-            i = limit.value.index("/")
-        except ValueError:
-            raise FirewallError(INVALID_RULE, "Expected '/' in limit")
+        rate, duration = limit.value_parse()
```
erig0, line 1597: This usage makes me think `value_parse()` is the wrong name for the function. If I read your commit message right the value is already parsed and normalized.

These make more sense to me:
```
rate, duration = tuple(limit)
```
or
```
rate, duration = limit.tuple()
````

Then later when you add burst the tuple size increases, but python is smart.
```
>>> foo, bar = (1,2,3)
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
```

Individual values are still available via `obj.value` and `obj.burst`.

---

> Now parse and stringify the value, so that it is normalized. Invalid
values are left unnormalized, and Rich_Limit.__init__() still does not
fail the operation (like before). For that we have check().

File: src/firewall/core/rich.py [hunk 6]
```diff
@@ -434,11 +434,13 @@ def check(self, family=None):
 
 
 class Rich_Limit(_Rich_Entry):
-    def __init__(self, value):
+    def __init__(self, value, burst=None):
         self.value = value
+        self.burst = burst
```
erig0, line 439: `self._burst = burst`

erig0 (changes requested):

src/firewall/core/rich.py
thom311: because `value_parse()` operates on `self._value`.

And `_value_parse()` is static and takes the value. That is needed by the `@value.setter` to do the normalization.

src/firewall/core/nftables.py
thom311: >  If I read your commit message right the value is already parsed and normalized.

The values is normalized (if and only if it's normalizable). Normalization requires parsing. But it's still a string, that cannot be understood without parsing it.

some callers want to decompare the value into their parts. They want to parse it. Hence, there is still a parse, even after having the values normalized.

Yes, as the Rich_Rule setter already parses the value to normalize it, it could also keep around the parsed values. Like setting `self.duration` & `self.rate`. However, that means all those object, that just lie around in memory for the most part, also hold the redundant, parsed values. Instead, the parsing result is not kept. You have to re-request it. That is intentional, to avoid storing those values. It could of course be different. It trades some CPU vs. Memory. The usage is pretty much the same.

What a separate parse also gives, is that a Rich_Limit can also hold invalid value, and only blow up during check(). That is simpler, if the parsing is always done anew.

> But that assumes you fully parsed/normalized in the constructor.

The constructor does normalization via `self.value = value`. That implies it also does a parsing. However, it doesn't keep those parse-results around...

src/firewall/core/rich.py
thom311: no. This is intentional. so that the constructor does the normalization. As requested.

src/firewall/core/nftables.py
```diff
@@ -1594,14 +1594,12 @@ def _rich_rule_limit_fragment(self, limit):
...
-
         d = {
-            "rate": rate,
-            "per": rich_to_nft[duration],
+            "rate": limit.rate,
+            "per": rich_to_nft[limit.duration],
         }
 
-        burst = limit.burst_parse()
+        burst = limit.burst
```
thom311, line 1602: burst should be an integer, it needs to be parsed.

It also doesn't get normalized, like it got previously.

It would be good to add a unit test (but that is already true for the patches I had previously). Should be done.

src/firewall/core/rich.py
```diff
@@ -433,104 +435,69 @@ def check(self, family=None):
...
+    rate: int = field(init=False)
+    duration: str = field(init=False)
 
-    def check(self, family=None):
-        self.value_parse()
-        self.burst_parse()
+    def __post_init__(self):
+        rate, duration = self._value_parse()
+        object.__setattr__(self, "rate", rate)
+        object.__setattr__(self, "duration", duration)
```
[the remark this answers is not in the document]
thom311: now that calling `Rich_Limit()` can fail with an exception, did you check all callers that they handle that correctly?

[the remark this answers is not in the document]
erig0: `self.value` does not, but we also don't use it. It's only used to set `self.rate` and `self.duration`.

`self.burst` is verified to be an int and range. What else needs to be done to "normalize" it ?

src/firewall/core/nftables.py
erig0: It's guaranteed to be an int by the `Rich_Limit` constructor, see `_burst_check()`.

src/firewall/core/rich.py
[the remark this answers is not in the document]
erig0: > now that calling Rich_Limit() can fail with an exception, did you check all callers that they handle that correctly?

No. I didn't intend my commit to be complete; just illustrating other options.
I'm guessing that in most cases we want to propagate the error, i.e. it came from the UI or startup.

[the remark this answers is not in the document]
erig0: Fixed remaining usage of `limit.value`.

src/firewall/core/nftables.py
thom311: > It's guaranteed to be None or an int by the Rich_Limit constructor, see _burst_check().

Ah. I didn't know how dataclasses work. It's parsed to be an int. That's nice.

erig0: IMO, normalizing `limit.value` is nonsensical as "normal" varies depending on firewall backend and XML. Better to breakdown `limit.value` into `limit.rate` and `limit.duration`.

For the rich language, normalization is achieved with `str(rich_limit_obj)`. But internally, if we make all these Rich_Rule objects hashable, we won't have to use this string internally at all and thus won't care about the normalized string much.

src/firewall/core/rich.py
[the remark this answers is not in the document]
thom311: > self.value does not, but we also don't use it

Isn't it used for the hashing/equality implementation?

There are probably various places, that do care about the string representation of the value (preferably, in a normalized, expected form). For example, `ip4tables._rule_limit()`.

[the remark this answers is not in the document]
erig0: I'd actually like to consider making `limit.value` throw an exception when it's read. That would make sure nothing uses it. :)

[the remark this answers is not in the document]
thom311: I think it's sensible that `ip4tables._rule_limit()` uses `limit.value`, and doesn't have to go through to construct it from `f"{limit.rate}/{limit.duration}"`.

Sure, that means, `iptables` backend is fine with the default string format that `Rich_Rule.value` provides. Other users may not (nftables), and may need to implement their own thing based on the more fundamental values.

But leaving `value` unnormalized there, and then require `ip4tables._rule_limit()` to reimplement it, is not useful. It also means, that the constructed rule contains a newly allocated string, instead of sharing the one from `Rich_Rule.value`.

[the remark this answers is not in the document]
thom311: in other words, `value` has a sensible form to which we can normalize. And "coincidentally", that is also what `ip4tables._rule_limit()` needs. Win-Win.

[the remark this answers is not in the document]
erig0: pushed one line change to normalize `self.value`. I would still prefer ipXtables to not rely on this.

src/firewall/core/nftables.py
erig0: > Ah. I didn't know how dataclasses work. It's parsed to be an int. That's nice.

Actually it's not true. The types are hints and not enforced. I changed the function to `_burst_parse()` again to make sure it's normalized to an int.

thom311:
patch lgtm. I pushed 2 fixup commits for consideration.

erig0:
I squashed the code changes, moved unit tests to a separate commit, and added checks to verify `Rich_Limit` is hashable.

Will merged after CI.

erig0 (approved):

Issue #1266, opened 2023-12-18
Deprecate and Drop Lockdown Feature


thom311:
Deprecate and Drop lockdown feature.

The lockdown feature restricts access for local processes to change firewalld configuration. 

I tried to understand why this feature was added. It was introduced in commit 4cf6dae0f196 and first released in v0.3.0 (dbe0f39a297e) in March 2013. I find no clear justification in bugzilla.redhat.com ([[*]](https://bugzilla.redhat.com/buglist.cgi?classification=Red%20Hat&classification=Fedora&columnlist=product%2Ccomponent%2Cassigned_to%2Cbug_status%2Cshort_desc%2Cchangeddate%2Cbug_severity&component=firewalld&f0=OP&f1=OP&f2=product&f3=component&f4=alias&f5=short_desc&f6=status_whiteboard&f7=CP&f8=CP&j1=OR&longdesc=lockdown&longdesc_type=allwordssubstr&o2=substring&o3=substring&o4=substring&o5=substring&o6=substring&order=assigned_to%2C%20id%2C%20&product=Fedora&product=Red%20Hat%20Enterprise%20Linux%207&product=Red%20Hat%20Enterprise%20Linux%208&product=Red%20Hat%20Enterprise%20Linux%209&product=Red%20Hat%20Enterprise%20Linux%206&query_format=advanced&target_milestone=---&target_release=---&v2=firewalld&v3=firewalld&v4=firewalld&v5=firewalld&v6=firewalld)) or the mailing list ([[*]](https://lists.fedorahosted.org/archives/list/firewalld-devel@lists.fedorahosted.org/2013/4/)).

A local user/process may have permissions to change system configuration. In case of firewalld, access is mainly guarded by Polkit. The lockdown mode enable an alternative mode for granting access (`--lockdown-on`), with different rules.

If this were a useful feature, then Polkit should provide a convenient way to (temporarily?) override permissions. It already has, you can drop `/etc/polkit-1/rules.d/*.rules` files. If the existing Polkit functionality is not powerful/useful enough, maybe something else should be added. For example, firewalld's lockdown supports `--list-lockdown-whitelist-contexts`, while arguably Polkit rules cannot access the subject's SELinux context. Maybe that should be improved (in Polkit). However, firewalld providing an alternative authentication mode is not the way.

[...]

PR #1269, opened 2024-01-10
[th/deprecate-lockdown-pt1] deprecate Lockdown API (part 1)

thom311:
No change in behavior, just documentation update and D-Bus annotation (`org.freedesktop.DBus.Deprecated`).

https://github.com/firewalld/firewalld/issues/1266

A future part will remove part of Lockdown code.

Also try to improve how D-Bus methods are deprecated. The ability to annotate methods/signals (with `org.freedesktop.DBus.Deprecated`) really should have been improved in dbus-python project from the start. Maybe I should still send a patch for that. Regardless, even if dbus-python gets the feature today, we couldn't rely on it from firewalld yet.


thom311:
[Here](https://gitlab.freedesktop.org/dbus/dbus-python/-/merge_requests/30) dbus-python would gain support for an `is_deprecated` argument. If we ever could rely on this, we could drop code from firewalld.

doc/xml/firewalld.dbus.xml
erig0, line 1: Please add the CLI help text deprecation notice to `firewall-offline-cmd` as well.

erig0 (changes requested):
This looks great. Add the missing change to `firewall-offline-cmd` and I'll merge.

doc/xml/firewalld.dbus.xml
thom311: done (yes?)

erig0 (approved):

PR #1270, opened 2024-01-10
feat: call dispatcher scripts for dbus signals

erig0:
TODO: find out if there are SELinux implications.

This PR adds firewalld dispatchers. Dispatchers are executables that may be called when firewalld does something interesting. They're called for every dbus signal that firewalld emits.

From the new man page:
```man
FIREWALLD.DISPATCH(5)                            firewalld.dispatcher                            FIREWALLD.DISPATCH(5)

NAME
       firewalld.dispatcher - firewalld dispatcher

SYNOPSIS
       /usr/lib/firewalld/dispatcher.d

       /etc/firewalld/dispatcher.d

DESCRIPTION
       Firewalld may call dispatcher scripts whenever it performs an action, e.g. Reloaded, that may be interesting to
       external entities. These actions correspond to dbus signals as documented in firewalld.dbus(5). This is useful
       to entities that can't listen for dbus signals. This includes daemonless and dbus unaware software.

DISPATCHERS
       Dispatchers are an executable in one of two locations:

       /usr/lib/firewalld/dispatcher.d

       /etc/firewalld/dispatcher.d

       Executables from both locations will be called. If both locations contain the same executable name then both
       will be called. One does not override or replace the other.

       Dispatchers are called via systemd-run. They are transient systemd services with a randomly generated name.
       They are run with the systemd property TimeoutSec=60 to guarantee they don't result in long lived processes.
       There is no guarantee about execution order. Dispatchers may run simultaneously.

DISPATCHER ARGUMENTS
       Every dispatcher is called with the below arguments. Refer to firewalld.dbus(5) for details.

       DBUS interface

       DBUS signal

       DBUS signal arguments

       For example, upon a user executing "firewall-cmd --zone internal --add-service https" a dispatcher located at
       /etc/firewalld/dispatcher.d/foobar would be called thusly:

       /etc/firewalld/dispatcher.d/foobar org.fedoraproject.FirewallD1.zone ServiceAdded internal https 0

```


src/tests/unit/test_rich_rule.py
```diff
@@ -24,7 +24,7 @@ def _exp_rule_check_2(exp_rule, rule):
         == 'rule family="ipv6" source address="1:2:3:4:6::" icmp-block name="redirect" log prefix="redirected: " level="info" limit value="4/m"'
     )
     s = rule.element
-    assert type(s) == firewall.core.rich.Rich_IcmpBlock
+    assert type(s) is firewall.core.rich.Rich_IcmpBlock
```
thom311, line 27: This warning happens because latest `pip install flake8 -U` gives you 7.0.0, while the github action seems to get something older. That's somewhat surprising, because we don't set a fixed version in pip (which we should).

This is of course a recurring topic. I also run Fedora 39's `black==23.7.0`, which formats some code differently than the blessed `black==22.12.0`. But luckily, I can find a format that works with both versions.

May point is, very soon we shouldupdate the github action to use newer `black`/`flake8`... I will open an PR for that.

thom311: hm... while it makes sense to tie `black==22.12.0` (as it affects the chosen coding style), it makes a bit less sense to tie `flake8==7.0.0`. Nothing to do here.

I am only surprised that `pip install flake8` on github doesn't give 7.0.0. Dunno...

src/firewall/server/decorators.py
```diff
@@ -115,26 +115,31 @@ def decorator(func):
...
+            *self.dbus_args,
+            **self.dbus_kwargs,
         )
-        return dbus_decorator(func)
+        @functools.wraps(func)
+        def _impl(*args, **kwargs):
+            pass
 
-    return decorator
+        return _impl
```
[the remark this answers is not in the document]
thom311: also, it seems that `_impl()` should call `func()`.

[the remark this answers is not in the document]
thom311: ah, it doesn't matter usually, because the methods for D-Bus signals usually don't do anything. But they do something, so `func()` should be called.

src/firewall/server/decorators.py
```diff
@@ -137,7 +140,9 @@ def __call__(self, func):
         )
         @functools.wraps(func)
         def _impl(*args, **kwargs):
-            pass
+            GLib.idle_add(
+                run_dispatcher, self.dbus_interface, func.__name__, (args[1:])
```
thom311, line 144: run_dispatcher already spawns the process in the background. It's not clear why this needs to happen on an idle handler. The call does not block, so it can happen right away.

Spawning it in the background, also means that all processes run in parallel and their order is non-deterministic. `GLib.idle_add()` does not help with that. We could help with that, by `run_dispatcher()` building a list of jobs, and only spawning one job at a time. But `GLib.idle_add()` doesn't achieve that.

src/firewall/server/decorators.py
```diff
@@ -137,7 +140,9 @@ def __call__(self, func):
         )
         @functools.wraps(func)
         def _impl(*args, **kwargs):
-            pass
+            GLib.idle_add(
```
thom311, line 143: it doesn't seem right that this dispatcher runs for *every* D-Bus signal.

There are *a lot* of signals, and having a dispatcher for all of them seems very expensive, but also unnecessary.

It's also not clear that this should be tied to D-Bus signals at all. Just define events and points in time when a dispatcher script is invoked, and explicitly run it at that point. Yes, sometimes a D-Bus signal and a dispatcher script will be invoked together, but most often not.

E.g. the `PropertiesChanged` signal seems not a useful signal for also calling a dispatcher. But aside from a few examples where it's not useful, it's mainly about how expensive it is to spawn that many processes, over and over. And commiting to do such expensive actions that in form of a new API.

src/firewall/server/decorators.py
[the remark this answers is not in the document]
erig0: Yikes. It should definitely call `func()`. Thanks.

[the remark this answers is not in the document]
thom311: > it's not clear to me, what the advantage of this is. Could you elaborate in the commit message?

I meant, it's unclear what is the advantage of converting the decorator from a closure to a class.

src/firewall/server/decorators.py
thom311: ok, maybe the advantage is to emit the D-Bus signal faster.

spawning a process, even if it's async, may still be somewhat slow...

src/firewall/server/decorators.py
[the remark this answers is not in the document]
erig0: > it's not clear to me, what the advantage of this is. Could you elaborate in the commit message?

Was just to match the style of the other decorators and avoid triple nested function declarations. I'll leave it as-is.

src/firewall/server/decorators.py
erig0: > The call does not block, so it can happen right away.

It does not block, but still takes time to fork+exec and wait on `systemd-run` to start the real process.

> Spawning it in the background, also means that all processes run in parallel and their order is non-deterministic

Right, this is noted in the man page. To be order deterministic firewalld would have to wait for each one to run to completion before starting the next. Could use `systemd-run --wait` and `Popen.poll()`. We would just keep adding the work to the mainloop with `idle_add()`.

That means building the list as you say, then calling each one from the mainloop, polling for completion, then moving on to the next dispatcher. Definitely more work than fire and forget. It does mean firewalld could report the exit status of the dispatcher.

If we use `systemd-run --wait` then we could instead directly call the dispatchers from firewalld and avoid the dependency on `systemd-run`. I was only using `systemd-run` because we are doing "fire and forget".

If we think that it's important to have in-order and non-parallel execution of dispatchers, then I'll make this change. I see the argument either way.

src/firewall/server/decorators.py
erig0: > it doesn't seem right that this dispatcher runs for every D-Bus signal.

I could see omitting some stuff, e.g. `PropertiesChanged`, that you point out below, but I think we want to see all runtime and permanent configuration changes.

> It's also not clear that this should be tied to D-Bus signals at all. 

Both dbus signals and dispatcher are a "something interesting" happened. Why maintain two different concepts or sets of "events"? Reusing the dbus signals gives us dispatchers for cheap.

> But aside from a few examples where it's not useful, it's mainly about how expensive it is to spawn that many processes, over and over. And commiting to do such expensive actions that in form of a new API.

This costs only exists if there are dispatchers to call.

src/firewall/server/dispatcher.py
```diff
@@ -0,0 +1,36 @@
...
+    def _dispatchers(directories):
+        for directory in directories:
+            if not os.path.isdir(directory):
+                continue
+            for file in os.listdir(directory):
+                _path = os.path.abspath(directory + os.sep + file)
+                if os.path.isfile(_path) and os.access(_path, os.X_OK):
+                    yield _path
+
+    env = {"LANG": "C"}
```
thom311, line 20: doesn't `systemd-run` sanitize the environment? I think we would want that...

I think, if we want to set environment variables, we need to pass them as `-E` arguments to `systemd-run`.


systemd-run ends up logging lines like
```
# journalctl -u run-ref10f3923f304b379bfcf732e79d2987.service
Jan 19 15:48:07 rh1 systemd[1]: Started run-ref10f3923f304b379bfcf732e79d2987.service - /usr/bin/echo hallo world.
Jan 19 15:48:07 rh1 echo[11487]: hallo world
Jan 19 15:48:07 rh1 systemd[1]: run-ref10f3923f304b379bfcf732e79d2987.service: Deactivated successfully.
```

On the one hand, this is great. Built-in-logging for the scripts! On the other hand, it's also again that if we spawn too many scripts, it will spam the logs in a bad way.

src/firewall/server/dispatcher.py
```diff
@@ -0,0 +1,36 @@
...
+    prog_args1 = [COMMANDS["systemd-run"], "--no-block", "--property", "TimeoutSec=60"]
+    prog_args2 = [interface, signal] + [str(a) for a in args]
+
+    for file in _dispatchers([FIREWALLD_DISPATCHER, ETC_FIREWALLD_DISPATCHER]):
```
thom311, line 24: should be `list(_dispatchers(...))` to build a list of files upfront.


While `subprocess.Popen()` doesn't block, it still is maybe a bit expensive to fork and take a bit. I think we should try to build the list of scripts at once (of course, it's not atomic either, but at least do it fast).

src/firewall/server/dispatcher.py
```diff
@@ -0,0 +1,36 @@
...
+from firewall.core.logger import log
+from firewall.config import FIREWALLD_DISPATCHER, ETC_FIREWALLD_DISPATCHER, COMMANDS
+
+
+def run_dispatcher(interface, signal, args):
+    def _dispatchers(directories):
+        for directory in directories:
+            if not os.path.isdir(directory):
+                continue
+            for file in os.listdir(directory):
```
thom311, line 15: there is a race, that `directory` just got deleted after you check for it. Or maybe you can even get an exception when the permissions are lacking.

`_dispatchers()` shouldn't raise an exception.

Wrap a try/except around listdir().

erig0:
This feature will not be merged. It's better for other software to listen to the dbus events directly.

PR #1274, opened 2024-01-12
[th/various-for-deprecated-lockdown] various cleanup in preparation for dropping Lockdown feature

thom311:
These patches are a dependency for an upcoming branch to drop the Lockdown feature.

As that branch may take a while to be merged (e.g. on when we deem it right to break the feature), these patches can go in earlier.

### DIFF
--- a/src/firewall/command.py
+++ b/src/firewall/command.py
@@ -43,31 +43,35 @@ class FirewallCommand:
     def get_verbose(self):
         return self.verbose
 
-    def print_msg(self, msg=None):
-        if msg is not None and not self.quiet:
-            sys.stdout.write(msg + "\n")
+    def print_msg(self, msg=None, *, force=False, file=None, exit_code=None):
+        if msg is not None and (not self.quiet or force):
+            if file is None:
+                file = sys.stdout
+            file.write(msg + "\n")
+        if exit_code is not None:
+            sys.exit(exit_code)
 
-    def print_error_msg(self, msg=None):
-        if msg is not None and not self.quiet:
-            sys.stderr.write(msg + "\n")
+    def print_error_msg(self, msg=None, *, force=False, file=None, exit_code=None):
+        if file is None:
+            file = sys.stderr
+        self.print_msg(msg=msg, force=force, file=file, exit_code=exit_code)
 
-    def print_warning(self, msg=None):
-        FAIL = "\033[91m"
-        END = "\033[00m"
-        if sys.stderr.isatty():
+    def print_warning(self, msg=None, *, exit_code=None):
+        if msg is not None and sys.stderr.isatty():
+            FAIL = "\033[91m"
+            END = "\033[00m"
             msg = FAIL + msg + END
-        self.print_error_msg(msg)
+        self.print_error_msg(msg, exit_code=exit_code)
 
     def print_and_exit(self, msg=None, exit_code=0):
         # OK = '\033[92m'
         # END = '\033[00m'
         if exit_code > 1:
-            self.print_warning(msg)
+            self.print_warning(msg, exit_code=exit_code)
         else:
             # if sys.stdout.isatty():
             #   msg = OK + msg + END
-            self.print_msg(msg)
-        sys.exit(exit_code)
+            self.print_msg(msg, exit_code=exit_code)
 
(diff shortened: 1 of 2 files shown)


thom311:
to be clear, I am not particularly happy about some of the patches here.

However, I also think that the current way how D-Bus properties are handled is not great, and should be improved. This is an attempt in that direction... Granted, since dbus-python is deprecated for more than a decade, maybe what really should be done is to rewrite this all with GDBus+pygobject...

src/firewall/command.py
erig0, line 1: I don't see an improvement here.

> Tackling different behaviors by adding several functions is often less
flexible than having one function, that can be parametrized (via
arguments) for all those behaviors.

Sure that's accomplished by making them more complex. So flexibility is gained by adding optional arguments (complexity).

> For example, I want to print a message to stderr, but quit with
exit_code=0 afterwards.

When would you want to print to stderr, but return success?

src/firewall/command.py
```diff
@@ -43,31 +43,35 @@ def set_verbose(self, flag):
...
-
-    def print_error_msg(self, msg=None):
-        if msg is not None and not self.quiet:
-            sys.stderr.write(msg + "\n")
-
-    def print_warning(self, msg=None):
-        FAIL = "\033[91m"
-        END = "\033[00m"
-        if sys.stderr.isatty():
+    def print_msg(self, msg=None, *, force=False, file=None, exit_code=None):
```
erig0, line 46: AFACIS, `force` is not used in this PR or the PR for lockdown removal. Why is it being added? Same for `file`.

IMO, this is change for change's sake.

erig0 (changes requested):
Otherwise, lgtm.

IMO, we can drop the last commit that changes the print functions. I don't see it doing anything useful.

src/firewall/command.py
thom311: that will happen from `firewall-cmd` with any of the removed lockdown arguments. It prints a message to stderr that the command does nothing, but also exit with success.

src/firewall/command.py
thom311: ah, force was supposed to be used by the lockdown removal. Will add it there.

That print by default honors `self.quiet` is at least questionable, because quiet mode should not mean that all messages are suppressed, but rather that certain verbose output is suppressed.

Regardless of whether `force=False` is good to be the default value. The message that the Lockdown command does nothing, is not something that should be suppressed by `--quiet`.

```bash
$ ./src/firewall-cmd --lockdown-on -q
Lockdown feature is deprecated and removed.
```

src/firewall/command.py
thom311: > Sure that's accomplished by making them more complex. So flexibility is gained by adding optional arguments (complexity).

Calling
```python
    def print_msg(self, msg=None, *, force=False, file=None, exit_code=None):
        if msg is not None and (not self.quiet or force):
            if file is None:
                file = sys.stdout
            file.write(msg + "\n")
        if exit_code is not None:
            sys.exit(exit_code)
```
complex seems not agreeable.

It has a small number of arguments, with reasonable defaults and the name of those arguments is in a way that it makes sense. The implementation is literally 6 lines of code.

When you read `cmd.print_msg(msg, file=sys.stderr)`, then I guess it's clear what that means.

What these arguments however allow, is that callers modify the behavior in concise (but clear?) ways.

I mean, instead of
```python
cmd.print_error_msg(
    "Lockdown feature is deprecated and removed.", exit_code=0, force=True
)
```
it would be:
```python
cmd.print_error_msg("Lockdown feature is deprecated and removed.")
sys.exit(0)
```

But that doesn't support the `force=True`. And it makes a direct call to `sys.exit()` (because there is no `cmd.exit()`). Currently `cmd.exit()` of course only does `sys.exit()`. But optimally, all our calls to `sys.exit()` are at one place, so we could add a hook there (e.g. some printf debugging).


I don't claim that this patch is a major improvement. Of course, you can also do things differently. But it's IMO rather nice. When I complained about not being happy about this branch, I meant patch 2/3...

erig0: I think it's a lateral move, but at least you can support `force` now.

erig0 (approved):

PR #1276, opened 2024-01-14
feat(service): add LocalSend

MikeWalrus:


erig0:
Hrm, It looks like localsend upstream is considering changing the default port.
https://github.com/localsend/protocol/issues/12

Perhaps we should hold off on adding this service.

PR #1282, opened 2024-01-18
fix(io.ipset): raise exception if entries exceed limit

### LINKED ISSUE #1280: ipset: config checks should verify that the number of elements does not exceed ipset size
### What happened

Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=2258244

### What you expected to happen

firewall-offline-cmd --check-config should complain if there are more elements than "size" allows.

### How to reproduce it (as minimally and precisely as possible)

_No response_

### Anything else we need to know?

_No response_

### Firewalld Version

1.3.4

### Firewalld Backend

_No response_

### Linux distribution

_No response_

### Linux kernel version

_No response_

### Other information

_No response_

TorontoMedia:
Closes #1280

### DIFF
--- a/src/firewall/core/io/ipset.py
+++ b/src/firewall/core/io/ipset.py
@@ -322,6 +322,12 @@ class IPSet(IO_Object):
                             errors.INVALID_VALUE,
                             "Option '%s': Value '%s' is negative" % (key, config[key]),
                         )
+                    if key == "maxelem" and len(self.entries) > int_value:
+                        raise FirewallError(
+                            errors.INVALID_IPSET,
+                            "the number of ipset entries exceeds the maximum set size of %d"
+                            % int_value,
+                        )
                 elif key == "family" and config[key] not in ["inet", "inet6"]:
                     raise FirewallError(errors.INVALID_FAMILY, config[key])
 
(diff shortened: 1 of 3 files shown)


src/firewall/core/io/ipset.py
```diff
@@ -430,6 +430,16 @@ def ipset_reader(filename, path):
...
+
+    if "maxelem" in ipset.options and len(ipset.entries) > int(
+        ipset.options["maxelem"]
+    ):
+        raise FirewallError(
+            errors.INVALID_IPSET,
+            "the number of ipset entries exceeds the maximum set size of %s"
+            % ipset.options["maxelem"],
+        )
+
```
erig0, line 442: I think this more correctly belongs in `_check_config()`.

erig0 (changes requested):
Otherwise, lgtm.

src/firewall/core/io/ipset.py
TorontoMedia: You are correct. I have made the necessary changes and modified gh881 maximum element limit.

erig0 (approved):
Thank you!!

PR #1289, opened 2024-01-26
test(check-container): fix running of polkit tests

erig0:
In order to run polkit tests inside the container we must start systemd,
dbus-broker, and polkit. In order to do this we have to rework how the
container is built and start them with init.

### DIFF
--- a/src/tests/Makefile.am
+++ b/src/tests/Makefile.am
@@ -74,9 +74,18 @@ check-container-debian-sid-image: check-container-%-image:
 	echo "RUN apt-get install -y autoconf automake pkg-config intltool libglib2.0-dev \
 	                             xsltproc docbook-xsl docbook-xml iproute2 iptables ipset ebtables \
 	                             nftables libxml2-utils libdbus-1-dev libgirepository1.0-dev \
-	                             python3-dbus python3-gi python3-nftables \
-	                             procps network-manager gir1.2-nm-1.0 time" && \
-	echo "COPY . /tmp/firewalld"; \
+	                             python3-dbus python3-gi python3-nftables sudo \
+	                             procps network-manager gir1.2-nm-1.0 time policykit-1 systemd" && \
+	echo "ENV container docker" && \
+	echo "VOLUME [ \"/sys/fs/cgroup\" ]" && \
+	echo "CMD [\"/usr/sbin/init\"]" && \
+	echo "RUN mkdir -p /run/dbus" && \
+	echo "COPY . /root/firewalld" && \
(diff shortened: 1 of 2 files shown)


src/tests/Makefile.am
```diff
@@ -74,9 +74,18 @@ check-container-debian-sid-image: check-container-%-image:
 	echo "RUN apt-get install -y autoconf automake pkg-config intltool libglib2.0-dev \
```
thom311, line 74: it's still not clear, why you cannot manually start dbus-borker and polkit in the docker-init-less container.

Apparently it already works, with NetworkManager (and dbus-broker)?

Anyway. Using systemd inside the container is probably a better choice anyway. So ok. Would just have been nice to understand what the problem is.

src/tests/Makefile.am
```diff
@@ -74,9 +74,18 @@ check-container-debian-sid-image: check-container-%-image:
...
+	echo "ENV container docker" && \
+	echo "VOLUME [ \"/sys/fs/cgroup\" ]" && \
+	echo "CMD [\"/usr/sbin/init\"]" && \
+	echo "RUN mkdir -p /run/dbus" && \
+	echo "COPY . /root/firewalld" && \
+	echo "WORKDIR /root/firewalld" && \
+	echo "RUN ./autogen.sh" && \
+	echo "RUN ./configure --sysconfdir=/etc PYTHON=\"/usr/bin/python3\" am_cv_python_pythondir=/usr/lib/python3/dist-packages" && \
+	echo "RUN make" && \
+	echo "RUN make install" ; \
```
thom311, line 88: these containerfiles are constructed by shell commands, inside the Makefile.am.

It would have been better, if this were a separate script (and/or a separate containerfile) that is commited to git. That makes the functionality re-usable without make. It also makes the content easier to review, as there are not multiple laters of escaping. The script of course can still be called from make.

thom311 (approved):

src/tests/Makefile.am
erig0: I don't think there is a use case outside of the build/make/test? I did think about commonizing these dockerfiles, but they're actually different enough (packages, configure, alternatives) that it didn't seem useful. It would be more work to mix-in the differences.

src/tests/Makefile.am
erig0: > it's still not clear, why you cannot manually start dbus-borker and polkit in the docker-init-less container.

You could. Or we could be more like the actual distribution and just start systemd. Polkit also does not like to run without being dbus activated. Instead of hacking it up I decided to just run like the actual distro.

PR #1294, opened 2024-01-30
Add iperf{2,3} services

cyqsimon:
Add service definitions for `iperf2` and `iperf3`.

- [ ] Should `iperf2` simply be called `iperf` instead? Personally I think `iperf2` is clearer, but all major distros package `iperf2` as simply `iperf`. So I'm a bit torn.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -204,6 +204,8 @@ CONFIG_FILES = \
 	services/ident.xml \
 	services/imaps.xml \
 	services/imap.xml \
+	services/iperf2.xml \
+	services/iperf3.xml \
 	services/ipfs.xml \
 	services/ipp-client.xml \
 	services/ipp.xml \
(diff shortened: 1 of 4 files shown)


erig0:
> Should iperf2 simply be called iperf instead? Personally I think iperf2 is clearer, but all major distros package iperf2 as simply iperf. So I'm a bit torn.

Anecdotally, I've only ever used iperf3. So to me that's the "default" iperf.

I'm good with having iperf2 and iperf3 as separate service.

erig0:
This PR needs to update `config/Makefile.am` and `po/POTFILES.in`.

thom311 (approved):
I think the follow-up patch (that updates Makefile.am) should be squashed into the first one. It's not a self-contained change, and should not be a separate commit.

Otherwise, lgtm.

cyqsimon:
> I think the follow-up patch (that updates Makefile.am) should be squashed into the first one. It's not a self-contained change, and should not be a separate commit.
> 
> Otherwise, lgtm.

Github does provide squash merge functionality so we can just use that. Hope you're okay with this.

PR #1296, opened 2024-01-31
ci(github): update badges and workflows

TorontoMedia:
- a3e10a0: show test results of all branches and releases.
- ad1d676: update workflow action versions for both action/checkout and action/setup-python.


README.md
```diff
@@ -1,6 +1,6 @@
-[![Build status](https://github.com/firewalld/firewalld/workflows/testsuite/badge.svg)](https://github.com/firewalld/firewalld/actions/workflows/testsuite.yml?query=event%3Apush+workflow%3Atestsuite)
-[![Integration status](https://github.com/firewalld/firewalld/workflows/integration%20testsuite/badge.svg)](https://github.com/firewalld/firewalld/actions?query=event%3Apush+branch%3Amaster+workflow%3Aintegration-testsuite)
```
erig0, line 2: I was reading about the badges and AFAICS if you omit the `branch=` it will use the "default branch", i.e. main - or whatever we change it to in the future.

What do you think?

TorontoMedia: We could emit the branch from the badge icon, but removing it from the workflow URL would show the release results too.

erig0: Yes. Make sense to me. @thaller agree?

TorontoMedia: I agree and have removed the branch from the badges.

erig0 (approved):

PR #1301, opened 2024-02-06
[th/test-keywords] automatically add the filename to AT_KEYWORDS and show it

thom311:
To have it easier to run a single test, add the filename to AT_KEYWORDS.

`-k features/forward_ports.at:199` then works.

Also, show that name as part of the test name, otherwise it's cumbersome to see the right name that can be used to run the test.

### DIFF
--- a/src/tests/functions.at
+++ b/src/tests/functions.at
@@ -97,6 +97,12 @@ m4_define([FWD_RESTART], [
 m4_define([FWD_START_TEST], [
     AT_SETUP([$1])
     AT_KEYWORDS(FIREWALL_BACKEND)
+    AT_KEYWORDS(__file__:__line__)
+    AT_KEYWORDS(__file__)
+    AT_KEYWORDS(AT_LINE)
+    AT_KEYWORDS([m4_bregexp(__file__, [^.*/\([^/]*\)$], [\1])])
+    AT_KEYWORDS([m4_bregexp(__file__, [^.*/\([^/]*\)\.at$], [\1])])
+    AT_KEYWORDS([m4_bregexp(__file__, [^\(.*\)/[^/]*$], [\1])])
 
     dnl Default values for things that should be defined in atlocal. If atlocal
     dnl can't be found it's likely because the testsuite is run "standalone" and


src/tests/functions.at
erig0, line 1: I like showing this file name. This is favoring showing the filename over the test name though. Is there anyways to print them both? Even on a second line?

erig0: Line numbers may also not work for some test cases that are wrapped with `m4_if()`. Some are conditionally wrapped because they only apply to a specific backend.

erig0: Can we add the entire test group name as a keyword?

thom311: > I like showing this file name. This is favoring showing the filename over the test name though. Is there anyways to print them both? Even on a second line?

The patch abbreviates the test-name, if the both don't fit. It could simply not do the abbreviation, but then the "ok" lines are all mis-aligned. That seems bad. I think the width for the alignment of the "ok" is hardcoded in autoconf, and rather narrow.

> Line numbers may also not work for some test cases that are wrapped with m4_if(). Some are conditionally wrapped because they only apply to a specific backend.

Really? Do you have an example of what doesn't work well?

> Can we add the entire test group name as a keyword?

You mean,
```
    AT_KEYWORDS([$1])
```

That doesn't work well, because the test name usually contains spaces, like `[policy with mixed family zone source]`. So each word in the sentence becomes a keyword, like `-k mixed`. But that doesn't seem to useful.

I think the `__file__:__line__` is really the best choice for selecting a test with `-k`. It is now also shown as part of the test output.

See the commit message of the first commit, it lists which options you have for the `-k` option.

src/tests/functions.at
```diff
@@ -95,7 +95,9 @@ m4_define([FWD_RESTART], [
 ])
 
 m4_define([FWD_START_TEST], [
-    AT_SETUP([$1])
+    AT_SETUP(m4_if(m4_eval(m4_len([$1]) + m4_len(__file__) + m4_len(__line__) < 45), [1],
+          [__file__:__line__> $1],
+          [__file__:__line__> m4_substr([$1], 0, m4_eval(42 - m4_len(__file__) - m4_len(__line__)))… ]))
```
thom311, line 100: this should be
```
[__file__:__line__> m4_substr([$1], 0, m4_eval(43 - m4_len(__file__) - m4_len(__line__)))…]))
                                               ^^                                        ^^
```
but that doesn't work. Suggestion welcome.

src/tests/functions.at
erig0: > I think the width for the alignment of the "ok" is hardcoded in autoconf, and rather narrow.

It's done by m4 itself not the generated testsuite script. I looked for a bit, but could not find a way to customize it. IMO something wider would be nice, e.g. 120.

> Really? Do you have an example of what doesn't work well?

`src/tests/features/startup_failsafe.at`

> Can we add the entire test group name as a keyword?

As we found offline, this already works.

`make check TESTSUITEFLAGS="-k '<entire test group name with spaces>' "`

src/tests/functions.at
erig0: 🤷

thom311: ok, then we live with the fact that when the name gets abbreviated, it gets abbreviated one character shorter than would be necessary... *sigh*

src/tests/functions.at
thom311: > >  Really? Do you have an example of what doesn't work well?
> `src/tests/features/startup_failsafe.at`

I see.

`-k features/startup_failsafe.at`

gives

```
204: features/startup_failsafe.at:1> startup fails…  ok
```

OK.... not nice. But no solution?

thom311:
OK, then I drop the second commit. To easier find it, I copy the patch here:

```diff
From bc1f9db6ca9cceca6faf24f9267283f86d369ef0 Mon Sep 17 00:00:00 2001
From: Thomas Haller <thaller@redhat.com>
Date: Tue, 6 Feb 2024 16:24:01 +0100
Subject: [PATCH 1/1] improvement(tests): show filename along test name
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

It's most useful to see the filename where the test comes from. Also because
that can be used to re-run the test with the "-k" option.

The filename should be clear and a useful name for the test. If it is not, the
file name should be improved.

If the test name is too long, it gets abbreviated to not misalign the "ok"
output.

Unfortunately, there isn't much space before the "ok" gets misaligned. So the
test name often gets abbreviated. However, the filename seems more important to
see in full.

Result:
    ...
     29: features/rfc3964_ipv4.at:1> RFC3964_IPv4        ok
     32: features/policy.at:5> policy - xml              ok
     35: features/policy.at:109> policy - list           ok
     33: features/policy.at:79> policy - create          ok
     34: features/policy.at:96> policy - name            ok
     39: features/policy.at:822> policy - dispatch       ok
     40: features/policy.at:5351> policy - interfaces/…  ok
     42: features/policy.at:6121> policy - from file     ok
     36: features/policy.at:231> policy - options        ok
     31: features/helpers_custom.at:1> customer helpers  ok
     41: features/policy.at:6072> policy - target        ok
     44: features/policy.at:6300> policy - multiple us…  ok
     30: features/service_include.at:1> service include  ok
     43: features/policy.at:6138> policy - zone drifti…  ok
     50: features/forward_ports.at:279> forward ports …  ok
     45: features/services.at:1> services                ok
     37: features/policy.at:286> policy - priority       ok
     49: features/forward_ports.at:199> forward ports …  ok
     46: features/ports.at:1> ports                      ok
     47: features/source_ports.at:1> source ports        ok
     52: features/protocols.at:1> protocols              ok
     56: features/rich_destination_ipset.at:1> rich de…  ok
     58: features/rpfilter.at:1> rpfilter                ok
     55: features/rich_tcp_mss_clamp.at:5> tcp-mss-clamp ok
     59: features/zone_combine.at:1> zone - combine      ok
     57: features/zone.at:1> zone - target               ok
     54: features/icmp_blocks.at:1> ICMP blocks          ok
     63: features/reloadpolicy.at:1> check ReloadPolicy  ok
     38: features/policy.at:470> policy - zones          ok
     51: features/masquerade.at:1> masquerade            ok
     48: features/forward_ports.at:1> forward ports      ok
     60: features/ipset_defer_native_ipset_creation.at:1> …  ok
    171: features/rfc3964_ipv4.at:1> RFC3964_IPv4        ok
    174: features/policy.at:5> policy - xml              ok
     53: features/rich_rules.at:1> rich rules            ok
    176: features/policy.at:96> policy - name            ok

You can see, that the abbreviated terms waste one character. It should be

     56: features/rich_destination_ipset.at:1> rich des… ok

instead of

     56: features/rich_destination_ipset.at:1> rich de…  ok

Patch welcome to fix that. It would seem the right solution would be

  -          [__file__:__line__> m4_substr([$1], 0, m4_eval(42 - m4_len(__file__) - m4_len(__line__)))… ]))
  +          [__file__:__line__> m4_substr([$1], 0, m4_eval(43 - m4_len(__file__) - m4_len(__line__)))…]))

Except, that breaks the m4 tool. This is utterly infuriating. Suggestion
welcome.
---
 src/tests/functions.at | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git a/src/tests/functions.at b/src/tests/functions.at
index a07c1682a75e..2d680a1cf542 100644
--- a/src/tests/functions.at
+++ b/src/tests/functions.at
@@ -95,7 +95,9 @@ m4_define([FWD_RESTART], [
 ])
 
 m4_define([FWD_START_TEST], [
-    AT_SETUP([$1])
+    AT_SETUP(m4_if(m4_eval(m4_len([$1]) + m4_len(__file__) + m4_len(__line__) < 45), [1],
+          [__file__:__line__> $1],
+          [__file__:__line__> m4_substr([$1], 0, m4_eval(42 - m4_len(__file__) - m4_len(__line__)))… ]))
     AT_KEYWORDS(FIREWALL_BACKEND)
     AT_KEYWORDS(__file__:__line__)
     AT_KEYWORDS(__file__)
-- 
2.43.0

```

erig0 (approved):

PR #1303, opened 2024-02-09
[th/nft-mem] reorder code so we can release temporary object earlier during nft cmd

thom311:
During nft command, we may (temporarily) create a lot of data to track the rules. This increases the memory usage, even if only for a short time.

Rework some code, so that we can drop references earlier when we no longer need them.


thom311:
previously, there was a patch here which dropped the chunking from `nftables.set_restore()`. That doesn't work (yet?). Patch was dropped (for now?).

thom311:
> previously, there was a patch here which dropped the chunking from `nftables.set_restore()`. That doesn't work (yet?). Patch was dropped (for now?).

With patch `improvement(nftables): combine request for entries of ipset`, it works. Patch re-added.

src/firewall/core/nftables.py
```diff
@@ -239,10 +239,26 @@ def __init__(self, fw):
         self._nft_handle.set_json_output(True)
 
     def nft_cmd(self, json_root):
+        if isinstance(json_root, list):
+            # The caller gave us the JSON in a list, so we can drop the
+            # reference to the memory. Unpack the list, and delete the entry.
+            l = json_root
+            (json_root,) = l
+            del l[0]
```
erig0, line 246: I don't see any cases where `json_root` would be a list. It's always a dict.

This code also doesn't work. If json_root is a list then you can't assign `l` to a tuple.

```
>>> foo=[1,2,3,4,5]
>>> (blah,) = foo
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack (expected 1)
```

Or `json_root` is a list of one element, then this does nothing.

src/firewall/core/nftables.py
```diff
@@ -239,10 +239,26 @@ def __init__(self, fw):
...
         json_root_str = json.dumps(json_root)
+        del json_root
```
erig0: I'm not sure this does anything if other areas of the code, e.g. the caller, are holding a reference. It would have to be `json_root.clear()` to remove the dict entries.

src/firewall/core/nftables.py
```diff
@@ -238,17 +239,37 @@ def __init__(self, fw):
...
             # reference to the memory. Unpack the list, and delete the entry.
             l = json_root
             (json_root,) = l
             del l[0]
-        json_root_str = json.dumps(json_root)
-        del json_root
-        json_root_str = json_root_str.encode("utf-8")
-        rc, output, error = self._nft_handle.cmd(json_root_str)
+
+        if is_large and hasattr(self._nft_handle, "cmd_from_file"):
```
erig0, line 251: This `hasattr()` guard is weird. AFAIK, python-nftables has always had this function.

src/firewall/core/nftables.py
```diff
@@ -238,17 +239,37 @@ def __init__(self, fw):
...
+                    filename.encode("utf-8")
+                )
+            finally:
+                if filename is not None:
+                    firewall.functions.removeFile(filename)
+        else:
+            json_root_str = json.dumps(json_root)
+            del json_root
+            json_root_str = json_root_str.encode("utf-8")
+            rc, output, error = self._nft_handle.cmd(json_root_str)
```
erig0, line 272: Instead of adding complexity by sometimes doing it directly and sometimes doing it via a file, why not just always do it from a file?

src/firewall/functions.py
erig0, line 1: This function is unused by the end of this changeset. Please drop this commit.

erig0 (changes requested):

erig0:
There are a lot of changes here that I'm not convinced are providing real benefit. The ipset entry changes are for sure. But some of the explicit `del`s I am not convinced about.

Do you have any benchmarks about the improvements?

src/firewall/functions.py
```diff
@@ -673,6 +673,19 @@ def wrong_args_for_callable(fcn, *a, **kw):
     return True
 
 
+def iter_reiterable(iterable):
```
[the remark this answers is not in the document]
thom311: Most callers probably are just lists, so re-iterating them is fine. It would be wasteful to clone the list for that. But more importantly, on the next review, somebody thinks "oh, this clone is unneecessary". Then I would have to add a code comment explaining why the clone is even there. At this point, I wouldn't clone the list.

But that makes the function less general and the caller needs to make sure not to mess up (in non-obvious ways). The function no longer accepts arbitrary iterables, only iterables that can be iterated more than once. So, maybe I should add a code comment there that indicates that this function can only be called with certain iterables. But all callers currently are lists (probably?), so I wouldn't want to add such a code comment either.

`iter_reiterable()` solves this. More than anything, it clearly expresses the intent of making the implementation safe. The caller no longer need to worry. The reviewer of the code also sees clearly why this is done, and can reason why this is the correct thing to do.

It doesn't matter whether all callers today(!!) pass a `list()`. Using `iter_reiterable()` in a function makes the function more accepting to input and less of a foot gun.

src/firewall/functions.py
```diff
@@ -458,6 +458,13 @@ def firewalld_is_active():
     return False
 
 
+def removeFile(filename):
+    try:
+        os.remove(filename)
+    except OSError:
```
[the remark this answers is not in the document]
thom311: I think not. At least, not for the caller at hand.

The caller is `nft_cmd()`, and that one really does not want to be bothered with this. It certainly shouldn' t just let the exception escape, so, it would need to handle it. But how? Logging a message?

If the caller still needs to handle `OSError` (e.g. because the path is a file, or because the file has wrong permissions), then they should handle `FileNotFoundError` themselves. This function is not for them. This function is for "I don't care about OSError and don't want to write 4 extra lines of code for that" (and that's the only purpose).

src/firewall/core/nftables.py
thom311: I think it works as intended. `(blah,) = foo` is tuple unpacking, and unpacks an interable of one element. It checks that the number of element matches, which is the reason for doing this instead of `blah = foo[0]`.

> Or json_root is a list of one element, then this does nothing.

If the caller passes `json_root`  as a list, then it MUST be a list of one element (otherwise, the tuple unpacking fails). But then it does the right thing. Unpack the list, and clear the element from the list.

> I don't see any cases where json_root would be a list. It's always a dict.

yes, it was previously unused. I now added a two commits
- improvement(nftables): avoid creating JSON string twice before nft_cmd()
- improvement(nftables): free json data earlier during nft_cmd()

src/firewall/core/nftables.py
thom311: I think it will be slower to write a small file. The branch change should not make some cases slower.

src/firewall/core/nftables.py
thom311: it's only "recently" added to the python library, here: https://git.netfilter.org/nftables/commit/?id=448bd104d3530ed2870921a35ebf54589cdb274c

This is recent enough, that github CI tests would break without this.

src/firewall/core/nftables.py
```diff
@@ -233,9 +233,9 @@ def __init__(self, fw):
         self.rich_rule_priority_counts = {}
         self.policy_dispatch_index_cache = {}
 
-        self.nftables = Nftables()
-        self.nftables.set_echo_output(True)
-        self.nftables.set_handle_output(True)
+        self._nft_handle = Nftables()
```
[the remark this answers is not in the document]
thom311: I search for `self._nft_handle` (or whatever it's called).

It's in general good to name different things differently. Extra bonus points for unique names.

Rename to "_nft_ctx".

src/firewall/core/nftables.py
thom311: it would, if `json_root` argument were a list.

I added commit "improvement(nftables): free json data earlier during nft_cmd()" to do that.

thom311:
> Do you have any benchmarks about the improvements from del and clear() commands?

No. You brought up how the ipset transaction has to be split due to memory (peak) usage. The deletes and clears allow it that memory that is no longer needed gets released earlier. I think they are successful at that, because (from review) it is clear that the variables are not aliased (have no other references). Also, CPython uses reference counting (and a GC for reference loops), so once those (loopless) variables get freed right away. Does it measurably help performance? That is hard to say.

Equally important, does it hurt readability/maintainability? I don't think so. A `del` reduces the scope of a variable (which you otherwise cannot do in Python). Reducing the scope is  IMO on its own a good thing (for maintainability), because it becomes clear that a variable shall not be used after that point. For example, `nft_cmd()` function is 41 lines, easy to understand and maintain. Dropping the `del`s there, doesn't make it significantly easier.

Too bad, that `flake8` (currently) cannot catch using a variable after it was deleted. However, `mypy` can (which we should enable).

PR #1308, opened 2024-02-13
[th/ipaddr-parse-pt1] add new function for parsing IP addresses

thom311:
This is the first part of https://github.com/firewalld/firewalld/pull/1172 , which is pending review. Split the branch into smaller pieces.

the overall goal is to have solid functions for handling strings. firewalld mainly deals with (firewall) configuration, where we express data as strings. We need functions that can handle those strings, and thus define (by implementation) how the strings are handled. Such functions in part already exist, however, their implementation and API is lacking. This introduces new functions. In particular:

- functions can handle IPv4/IPv6 similarly (there is now an `addr_family` argument)
- the branch adds unit tests.
- the branch fixes bugs in the existing implementation
- the new functions have a better name. Better, because they establish a pattern how similar function (eg. to parse a MAC address or a port) will be called.

### DIFF
--- a/src/firewall/functions.py
+++ b/src/firewall/functions.py
@@ -84,6 +84,12 @@ def addr_family_bitsize(family):
 ###############################################################################
 
 
+IPAddrZero4 = b"\0\0\0\0"
+
+
+###############################################################################
+
+
 def getPortID(port):
     """Check and Get port id from port string or port id using socket.getservbyname
 
--- a/src/firewall/functions.py
+++ b/src/firewall/functions.py
@@ -5,6 +5,7 @@
 # Authors:
 # Thomas Woerner <twoerner@redhat.com>
 
+import enum
 import socket
 import os
 import os.path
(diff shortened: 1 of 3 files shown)


src/firewall/functions.py
```diff
@@ -26,6 +26,22 @@
 }
 
 
+###############################################################################
+
+
+class BugError(Exception):
```
erig0, line 33: This already exists in `src/firewall/errors.py`. Please drop this commit.

src/firewall/functions.py
```diff
@@ -101,6 +102,94 @@ def addr_family_bitsize(family):
...
+
+    return f
+
+
+def _parse_norm_fcn(parse_fcn, unparse_fcn):
+    def f(*a, **kw):
+        detail = parse_fcn(*a, **kw)
+        return unparse_fcn(*detail)
+
+    return f
```
erig0, line 121: These are only used by functions that I'm asking you to drop. Please also drop these.

src/firewall/functions.py
```diff
@@ -124,6 +124,71 @@ def f(*a, **kw):
...
+def ipaddr_unparse(addrbin, family, *, with_ip6_brackets=False):
+    s = socket.inet_ntop(family, addrbin)
+    if with_ip6_brackets and family == socket.AF_INET6:
+        s = f"[{s}]"
+    return s
+
+
+ipaddr_check = _parse_check_fcn(ipaddr_parse)
+
+ipaddr_norm = _parse_norm_fcn(ipaddr_parse, ipaddr_unparse)
```
erig0, line 186: Unused. Please drop.

erig0 (changes requested):
In the last PR I said that we should not be using `socket`. Instead we should be using `ipaddress`. That still stands.

src/firewall/functions.py
thom311: they are not unused. This is part1 of a a larger set, they are used afterwards (on #1172).

Also, they are useful as they replace for example `normalizeIP6()` and `checkIP()`.

thom311:
> In the last PR I said that we should not be using `socket`. Instead we should be using `ipaddress`. That still stands.

To which I replied.

- `socket.{AF_INET,AF_INET6,AF_UNSPEC}` is used as identifiers for the address family. Yes, it could also be `"ipv4"` or "4" or `functions.IP4_ADDRESS_FAMILY`, but this is a sensible choice, because it's well-known and relevant to API like `inet_ntop()`.

- `socket.inet_ntop()` and `socket.inet_pton()` is used to convert IP addresses to/from strings. Since we throughout the code use strings for IP addresses, that stays. We wouldn't pass around `ipaddress` instances, because that doesn't seem useful and because it would be a larger change (API that currently accepts strings, would need to change). Strings are fine. All we need, is a function to parse/unparse the strings. These functions are `ipaddr_parse()`/`ipaddr_unparse()`, which call `socket.inet_pton()`/`socket.inet_ntop()`.

What does `ipaddress` module provide here?

erig0:
> What does ipaddress module provide here?

It does all the parsing, unparsing, normalization, and family detection. There is no reason for us to reinvent that using custom code + `socket`.

I won't merge PR.

PR #1331, opened 2024-04-23
improvement(service): transmission-client: Add 9091/tcp

luk1337:
Transmission client (as well as daemon) uses 9091 by default for its remote web interface.


luk1337:
bump?

erig0:
AFAICS, the web interface is not enabled by default. If that's true, does it make sense to open in the standard `transmission-client` service? Or do we need `transmission-client-web` service?

erig0:
No response to needinfo. Closing the PR. It can be reopened if the above is answered.

PR #1334, opened 2024-05-04
feat(service): add SlimeVR

ImUrX:


erig0:
AFAICS, only `6969/udp` is used. Do we need 6868?

https://github.com/SlimeVR/SlimeVR-Server/blob/2c146169acfee8071cd2759b2227997258b0e8b2/server/core/resources/firewall.bat#L18

ImUrX:
That commit that changed this is a bit newer than this PR, we can remove the 6868

erig0 (approved):
Thanks. Will merge after CI.

PR #1341, opened 2024-05-16
feat(service): xrootd file server

jcpunk:
This adds a service definition for the `xrootd` file service.


erig0 (changes requested):
IANA calls this `rootd` [1]. That's what we should call it as well.

[1]: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=rootd

jcpunk:
Changed, does that look better?

erig0 (approved):
Thanks!

PR #1342, opened 2024-05-16
feat(service): Add Steam local network transfers service

alosarjos:
In the same way there is a service for In-Home streaming, I find useful to have a service for Local Network Game transfers.

The required ports come from the documentation here: https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43


DanMan:
Already added: https://github.com/firewalld/firewalld/pull/1216

PR #1345, opened 2024-06-03
feat(service): stun and stuns protocols

maddymeows:
References:

STUN: [IETF RFC 8489](https://datatracker.ietf.org/doc/html/rfc8489)
TURN: [IETF RFC 8656](https://datatracker.ietf.org/doc/html/rfc8656)

STUN and TURN use the same ports and are very closely related and often end up used interchangeably.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -331,6 +331,8 @@ CONFIG_FILES = \
 	services/steam-lan-transfer.xml \
 	services/stellaris.xml \
 	services/stronghold-crusader.xml \
+	services/stun.xml \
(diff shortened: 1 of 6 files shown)


erig0 (changes requested):
I think it makes sense to add a service named `turn`. It could use the `include` directive for the `stun` service.

Please also update `po/POTFILES.in` and `config/Makefile.am`.

erig0 (approved):

erig0:
CI failure is almost certainly a change in nftables upstream. Merging!

PR #1347, opened 2024-06-04
fix(firewall-applet): Set fallback icons

City-busz:
Qt6 does not fallback to the normal icon if the -shields_up/-shields_down icons are missing, causing missing icons for the applet. Set the normal icon as fallback explicitly.


erig0 (approved):
Thanks!

Issue #312, opened 2018-03-12
firewall-config does not fit screen


hwoarang:
On a display with 1366x768 the firewall-config window is too high so buttons at the bottom are not reachable. The window also doesn't allow itself to be resized making it impossible to use the GUI.

I have attached a screenshot to show the problem

![65888653](https://user-images.githubusercontent.com/45309/37275496-4efbc72a-25d7-11e8-84bc-53ad10b5252a.png)

PR #1350, opened 2024-06-08
fix(firewall-config): Add long labels into a scrolled window

City-busz:
This allows the main window to fit into smaller screens, regardless of the size of the labels.

This is a better and more general fix for https://github.com/firewalld/firewalld/issues/312.


erig0 (approved):
Thanks!

PR #1358, opened 2024-06-23
feat(service): add MPD (Music Player Daemon)

frozencemetery:
Service description taken from its website <https://www.musicpd.org/>.


erig0 (approved):
I updated the build files. Will merge after CI.

PR #1388, opened 2024-09-22
fix: Fix translating appdata file

City-busz:
It needs `@INTLTOOL_XML_RULE@` instead of `@INTLTOOL_XML_NOMERGE_RULE@` to actually work.


erig0 (approved):
Thanks.

Issue #1389, opened 2024-09-22
Add services for socks proxy with TCP 1080 and for http proxy with TCP 3128.


13ilya:
### What would you like to be added

I noticed that the list of services does not include proxy TCP 3128 and socks TCP 1080, which are very popular and quite standardized.
Then I noticed that proxy still exists under the name squid, which I think is not correct, as there are many other proxy servers besides squid, for example 3proxy, and it would be more logical to bind not to the name of the proxy server, but to the purpose, for example as in hhtp, https, http3.
I have 3 variations of the more obvious and preferred names so far.
```
proxy.xml socks.xml
proxy-http.xml proxy-socks.xml
http-proxy.xml socks-proxy.xml
```

### Why is this needed

Proxy servers, especially with the socks5 protocol, are still very relevant.
There are a huge number of them, and they should have a convenient and obvious service out of the box.

PR #1393, opened 2024-10-05
feat(service): add proxy-http socks

13ilya:
https://github.com/firewalld/firewalld/issues/1389

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -290,6 +290,7 @@ CONFIG_FILES = \
 	services/prometheus-node-exporter.xml \
 	services/prometheus.xml \
 	services/proxy-dhcp.xml \
+	services/proxy-http.xml \
 	services/ps2link.xml \
 	services/ps3netsrv.xml \
 	services/ptp.xml \
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -327,6 +327,7 @@ CONFIG_FILES = \
 	services/snmp.xml \
 	services/snmptls-trap.xml \
 	services/snmptls.xml \
+	services/socks.xml \
 	services/spideroak-lansync.xml \
 	services/spotify-sync.xml \
 	services/squid.xml \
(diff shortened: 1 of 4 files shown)


IlyaIndigo:
How do you pass Semantic PR?

erig0:
The commit message (first line) must match one of the keywords [specified in the config](https://github.com/firewalld/firewalld/blob/main/.github/semantic.yml).

e.g. `feat(service): add http proxy`

13ilya:
Can you please tell me what is wrong with my PR?
Why is it being ignored and not checked?

config/services/proxy.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>proxy</short>
+  <description>HTTP/HTTPS proxy server</description>
+  <port protocol="tcp" port="3128"/>
```
erig0, line 5: Port `3128` belongs to `ndl-aas` according to ICANN. Furthermore the service name `proxy` is too generic.

Where are you getting port `3128` ?

erig0 (changes requested):

config/services/proxy.xml
13ilya: 1 Port 3128 is used by default by all proxy servers I know. 
2 I suggested other options in #1389, such as proxy-http and proxy-socks, and I'm fine with those too.
3 https://github.com/firewalld/firewalld/blob/main/config/services/squid.xml This in no way prevented you from adding the squid service with an even worse name, even though squid is far from being the only proxy server.

erig0: I'd like to see it renamed to `proxy-http`.

erig0: `socks` is okay. That's the IANA assigned name.

erig0 (approved):
I split your commit into two. I'll merge after CI.

Issue #1399, opened 2024-10-11
firewalld is unusable with ~60 interfaces added to a zone. 


alisarctl:
### What happened

Adding ~60 interfaces to a zone, then running firewall-cmd --reload takes long time and the dbus call itself times out.

### What you expected to happen

The old firewall-0.9.3 works with no issue

### How to reproduce it (as minimally and precisely as possible)

for ((i = 1; i < 70; ++i)); do
  ip link add tun${i} type dummy      
  firewall-cmd --permanent --zone=public --add-interface=tun${i}
done

#firewall-cmd --reload
ERROR:dbus.proxies:Introspect error on :1.18:/org/fedoraproject/FirewallD1: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
Error: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.

### Anything else we need to know?

On firewalld-2.0.1 (and later version) 81063 rules are created (versus 875 rules only with firewalld-0.9.3). And apply these long list of rules times causing dbus timeout.

I'm dumped the rules created, there are loads of 
'filter_FORWARD_POLICIES', 'nat_POSTROUTING_POLICIES' created for each combination of tunX-tunY, generating a huge amount of rules that take long time to apply.

These rules are being created in
_interface_or_source_update_policies_derived_from_zone

erig0:
It generates a very dumb list of rules. Largely to deal with unwritten rules in zone/policy dispatch, e.g. sources before interfaces, sources sorted by zone name.

The rule generation needs to be optimized. For nftables, we could use anonymous sets. For iptables it's trickier. We might be able to use ipset. If we don't come up with a solution for iptables, then IMO, this is justification for dropping iptables support.

Alix82:
Thank you, glad that this is treated with high priority, as this affects lots of setups.

fargburger:
I have tried this with ipsets (with the hash:net,iface type), it actually appears to work for iptables (although I'm not confident that it is matching the interface), but nftables has a curious rule in nftables.py 

https://github.com/firewalld/firewalld/blob/d3052a9584ccb0369e5a3287c87468e31cd94625/src/firewall/core/nftables.py#L2822

[...]

PR #1406, opened 2024-10-17
fix: match daddr to oifname and saddr to iifname

fargburger:
fix: Fixes this issue where ipsets with IPs and interfaces are mismatched between outbound interface and destination IP address

https://github.com/firewalld/firewalld/issues/1399#issuecomment-2420036920


beninidavide:
I too replicated the same issue; using iptables as backend works correctly, while using nftables I have the issue  

I am testing sending a UDP packet from an external server to my host
(10.100.53.42); on my host the following zone and ipset are defined

zone "test1", in which I add an ipset made of one host.
```bash
# firewall-cmd --zone=test1 --list-all
test1 (active)
  target: default
  ingress-priority: 0
  egress-priority: 0
  icmp-block-inversion: no
  interfaces: eth0
  sources:
  services: dhcpv6-client ssh
  ports:
  protocols:
  forward: no
  masquerade: no
  forward-ports:
  source-ports:
  icmp-blocks:
  rich rules:
	rule family="ipv4" source ipset="test" log prefix="test1 " level="debug" accept
```
the ipset "test":
```bash
# firewall-cmd --get-entries --ipset=test
10.100.52.226/32,eth0
```


## Using iptables as backend
I can see (output of `iptables-save`)
```bash
[...]
-A IN_test1_allow -p tcp -m tcp --dport 22 -j ACCEPT
-A IN_test1_allow -m set --match-set test src,src -j ACCEPT
-A IN_test1_log -m set --match-set test src,src -j LOG --log-prefix "test1 " --log-level 7
```
When sending a packet from the allowed source 10.100.52.226, the packet is accepted; in firewall logs I see:  
```kernel: [    C0] test1 IN=eth0 OUT= MAC=52:54:ee:d4:ee:b8:52:54:ee:54:ce:7d:08:00 SRC=10.100.52.226 DST=10.100.53.42 LEN=35 TOS=0x00 PREC=0x00 TTL=64 ID=11666 DF PROTO=UDP SPT=48073 DPT=9999 LEN=15```

When sending a packet from another IP, the packet is blocked as expected

## Using nftables as backend
firewall logs show that a packet from 10.100.52.226, which was previously passing when using iptables as backend, is now
rejected:
```kernel: [    C0] filter_IN_test1_REJECT: IN=eth0 OUT= MAC=52:54:ee:d4:ee:b8:52:54:ee:54:ce:7d:08:00 SRC=10.100.52.226 DST=10.100.53.42 LEN=35 TOS=0x00 PREC=0x00 TTL=64 ID=2324 DF PROTO=UDP SPT=59530 DPT=9999 LEN=15```

and the output of nft list says:

```bash
# nft -a list chain inet firewalld filter_IN_test1_allow
table inet firewalld {
	chain filter_IN_test1_allow { # handle 340
		tcp dport 22 accept # handle 344
		ip6 daddr fe80::/64 udp dport 546 accept # handle 345
		ip saddr . oifname @test accept # handle 347
	}
}
```

Now I add a rule using iifname:
```bash
# nft insert rule inet firewalld filter_IN_test1_allow position 347 ip saddr . iifname @test accept
# nft -a list chain inet firewalld filter_IN_test1_allow
table inet firewalld {
	chain filter_IN_test1_allow { # handle 340
		tcp dport 22 accept # handle 344
		ip6 daddr fe80::/64 udp dport 546 accept # handle 345
		ip saddr . iifname @test accept # handle 455
		ip saddr . oifname @test accept # handle 347
	}
}
```

Now I try again sending the udp packet and it's accepted (only from the adddress
10.100.52.226, others are blocked)

fargburger:
@beninidavide  thank you for confirming that, I had a very similar test case.  What I couldn't tell was whether it was programmed this way for a good reason for any other cases.  I couldn't get the unit tests to run on my development machine (I wasn't trying that hard) but if the maintainers need me to improve this PR, I'm more than happy to.  

I thought I fixed the Semantic issues too - but maybe that step doesn't get re-checked in github?

erig0:
This fix looks correct. I'm keeping it open until test coverage can be added. I'll try to get to it next week.

erig0 (dismissed):
I added a test. IMO this is ready to merge after CI.

PR #1425, opened 2024-12-05
Update firewalld.conf

jschwender:
see discussion #1424


erig0:
Currently logging goes to the kernel buffer. We can't default to spamming the kernel log buffer. It can also cause serious performance issue if the kernel logs go to a serial console.

If firewalld were to gain better log, e.g. NFLOG support, then perhaps log by default would make sense.

PR #1440, opened 2024-12-23
fix(firewall-config): fix rich rule creation

xzakharov:
It's impossible to create a rich rule because the Rich_Rule constructor is incorrecly invoked on lines 3404-3709 of src/firewall-config.in. Moreover, Rich_Rule attributes are set for its instance, which, due to the frozen state of Rich_Rule, leads to even more errors.

Basically, the fix is quite simple: just create a dictionary of attributes and pass them to the Rich_Rule constuctor at once.

### DIFF
--- a/src/firewall-config.in
+++ b/src/firewall-config.in
@@ -3699,25 +3699,22 @@ class FirewallConfig:
             _("debug"): "debug",
         }  # 7, debug-level messages
 
+        rule_params = {}
+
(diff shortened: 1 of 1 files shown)


erig0:
Thank for the fix. Sorry for the delay and breaking it. :)

I'll merge after CI.

erig0 (approved):

PR #1443, opened 2024-12-24
fix(firewall-config): allow selecting the first service in the list

xzakharov:
When the service selection dialog opens and no service is set, you can't
select the first service in the list right away. You have to choose
another service and then select the first one. This happens due to the
unconditional disabling of the service selection button.

To resolve this issue, we can simply check whether the service is set by
verifying that the `old_service` param of the `service_select_dialog`
function is not empty.


erig0 (approved):
lgtm. Thanks!

Issue #1436, opened 2024-12-20
LogDenied=unicast logs ipv6 multicast traffic


SDedik:
### What happened

I have LogDenied option set to 'unicast'
```
cat /etc/firewalld/firewalld.conf | grep LogDenied=
LogDenied=unicast
```

However, in /var/log/kern I see tons of entries related to ipv6 multicast
```
Dec 20 11:01:41 <redacted> kernel: rpfilter_DROP: IN=eth0 OUT= MAC=33:33:00:00:00:02:34:40:b5:ab:f2:b8:86:dd SRC=fe80:0000:0000:0000:3640:b5ff:feab:f2b8 DST=ff02:0000:0000:0000:0000:0000:0000:0002 LEN=56 TC=0 HOPLIMIT=255 FLOWLBL=148171 PROTO=ICMPv6 TYPE=133 CODE=0
Dec 20 11:01:55 <redacted> kernel: rpfilter_DROP: IN=eth0 OUT= MAC=33:33:00:00:00:fb:6c:b3:11:13:d2:d0:86:dd SRC=fe80:0000:0000:0000:bbb1:96ae:416a:d4b2 DST=ff02:0000:0000:0000:0000:0000:0000:00fb LEN=85 TC=0 HOPLIMIT=1 FLOWLBL=223537 PROTO=UDP SPT=5353 DPT=5353 LEN=45
Dec 20 11:01:55 <redacted> kernel: rpfilter_DROP: IN=eth0 OUT= MAC=33:33:00:00:00:fb:6c:b3:11:13:d2:d0:86:dd SRC=fe80:0000:0000:0000:bbb1:96ae:416a:d4b2 DST=ff02:0000:0000:0000:0000:0000:0000:00fb LEN=85 TC=0 HOPLIMIT=1 FLOWLBL=223537 PROTO=UDP SPT=5353 DPT=5353 LEN=45
Dec 20 11:02:28 <redacted> kernel: rpfilter_DROP: IN=eth0 OUT= MAC=33:33:00:00:00:16:00:50:56:6a:a2:0e:86:dd SRC=fe80:0000:0000:0000:0250:56ff:fe6a:a20e DST=ff02:0000:0000:0000:0000:0000:0000:0016 LEN=76 TC=0 HOPLIMIT=1 FLOWLBL=0 PROTO=ICMPv6 TYPE=143 CODE=0
[...]

PR #1451, opened 2025-02-01
fix(rpfilter): use log denied packet type

TorontoMedia:
Fixes: #1436

### DIFF
--- a/src/tests/features/rpfilter.at
+++ b/src/tests/features/rpfilter.at
@@ -2,6 +2,7 @@ FWD_START_TEST([rpfilter - strict])
 AT_KEYWORDS(rpfilter)
 CHECK_NFTABLES_FIB()
 
+AT_CHECK([sed -i 's/^LogDenied.*/LogDenied=multicast/' ./firewalld.conf])
 AT_CHECK([sed -i 's/^IPv6_rpfilter.*/IPv6_rpfilter=strict/' ./firewalld.conf])
 FWD_RELOAD()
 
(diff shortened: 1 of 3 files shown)


src/tests/features/rpfilter.at
```diff
@@ -2,21 +2,23 @@ FWD_START_TEST([rpfilter - strict])
...
+AT_CHECK([sed -i 's/^LogDenied.*/LogDenied=multicast/' ./firewalld.conf])
 AT_CHECK([sed -i 's/^IPv6_rpfilter.*/IPv6_rpfilter=strict/' ./firewalld.conf])
 FWD_RELOAD()
 
 NFT_LIST_RULES([inet], [filter_PREROUTING], 0, [dnl
     table inet firewalld {
         chain filter_PREROUTING {
             icmpv6 type { nd-router-advert, nd-neighbor-solicit } accept
-            meta nfproto ipv6 fib saddr . mark . iif oif missing drop
+            meta nfproto ipv6 fib saddr . mark . iif oif missing meta pkttype multicast log prefix "rpfilter_DROP: " drop
```
erig0, line 13: Won't this cause it to drop only for multicast packets? Where as `LogDenied` is only about logging.

To support this, I think it would have to be two separate rules.

erig0 (changes requested):

src/tests/features/rpfilter.at
TorontoMedia: Correct, I apologize! I’ve now separated the rule configurations for cases where log_denied is not equal to 'all', and created distinct test cases to specifically validate each scenario.

PR #1472, opened 2025-04-12
feat(service): add wivrn

NikolasK-source:
see https://github.com/WiVRn/WiVRn

### DIFF
--- a/config/services/wivrn.xml
+++ b/config/services/wivrn.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>wivrn</short>
+  <description>
+    The Linux OpenXR streaming application to standalone headsets 
+  </description>
+  <port port="5353" protocol="udp"/>
+  <port port="9757" protocol="udp"/>
+  <port port="9757" protocol="tcp"/>
+</service>
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -372,6 +372,7 @@ CONFIG_FILES = \
 	services/wbem-http.xml \
 	services/wbem-https.xml \
 	services/wireguard.xml \
+	services/wivrn.xml \
 	services/ws-discovery-client.xml \
 	services/ws-discovery-host.xml \
 	services/ws-discovery-tcp.xml \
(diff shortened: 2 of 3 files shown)


config/services/wivrn.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>wivrn</short>
+  <description>
+    The Linux OpenXR streaming application to standalone headsets 
+  </description>
```
devurandom: Looking at other service files, e.g. https://github.com/firewalld/firewalld/blob/05586b0550894f886126905bbb8bba8f4e6536c4/config/services/alvr.xml, I believe the text could be a bit longer:
```suggestion
  <short>WiVRn wirelessly connects a standalone VR headset to a Linux computer</short>
  <description>
    Play PCVR games on the headset while processing is done on the computer. It supports a wide range of headsets such as Quest 1 / 2 / Pro / 3 / 3S, Pico Neo 3, Pico 4, HTC Vive Focus 3, HTC Vive XR elite and most other Android based headsets.
  </description>
```

NikolasK-source:
> I believe adding this to firewalld itself is not necessary. WiVRn already ships with its own firewalld config: https://github.com/WiVRn/WiVRn/blob/6cd65de21d9a8e31192d3af947c7f65b8b1640a1/server/dist/firewalld-wivrn.xml

This firewalld config is not available when WiVRn is installed as a Flatpak (as recommended by the project).

erig0:
> I believe adding this to firewalld itself is not necessary. WiVRn already ships with its own firewalld config: https://github.com/WiVRn/WiVRn/blob/6cd65de21d9a8e31192d3af947c7f65b8b1640a1/server/dist/firewalld-wivrn.xml

The problem here is downstream packaging. If firewalld starts shipping the `wivrn` service then it will cause a package conflict with the actual `wivrn` package as it wants to install the same file, `/usr/lib/firewalld/services/wivrn.xml`.

Some packages prefer to ship their own service definition. In fact, we have _removed_ services in firewalld because the project wanted to ship it themselves.

As such, before we can merged this, we need buy-in from the `wivrn` project. They can decide if they want to ship it themselves or let it be part of firewalld proper.

@NikolasK-source, Please consult with the `wivrn` maintainers.

Issue #1403, opened 2024-10-16
[firewalld 2.2.1] testsuite: 224 failed


mbiebl:
Test environment: debian sid / trixie
Filed also as downstream bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1083210

```
## ---------------------- ##
## Detailed failed tests. ##
## ---------------------- ##

#                             -*- compilation -*-
224. nftables_table_owner.at:1: testing nftables table owner ...
/nftables_table_owner.at:1: if ! cp /etc/firewalld/firewalld.conf ./firewalld.conf; then exit 77; fi
/nftables_table_owner.at:1: sed -i 's/^FirewallBackend.*/FirewallBackend=nftables/' ./firewalld.conf
/nftables_table_owner.at:1: ip netns add fwd-test-${at_group_normalized}
/nftables_table_owner.at:1:             env DBUS_SYSTEM_BUS_ADDRESS="unix:abstract=firewalld-testsuite-dbus-system-socket-${at_group_normalized}" ip netns exec fwd-test-${at_group_normalized} nft -f ./nft_rule_index.nft      
/nftables_table_owner.at:1:             env DBUS_SYSTEM_BUS_ADDRESS="unix:abstract=firewalld-testsuite-dbus-system-socket-${at_group_normalized}" ip netns exec fwd-test-${at_group_normalized} rm ./nft_rule_index.nft      
/nftables_table_owner.at:1:             env DBUS_SYSTEM_BUS_ADDRESS="unix:abstract=firewalld-testsuite-dbus-system-socket-${at_group_normalized}" ip netns exec fwd-test-${at_group_normalized} nft delete table inet firewalld_check_rule_index      
running
exit code = 0
/nftables_table_owner.at:1: sed -i 's/^NftablesTableOwner=.*/NftablesTableOwner=yes/' ./firewalld.conf
/nftables_table_owner.at:1:             env DBUS_SYSTEM_BUS_ADDRESS="unix:abstract=firewalld-testsuite-dbus-system-socket-${at_group_normalized}" ip netns exec fwd-test-${at_group_normalized} firewall-cmd -q --reload      
/nftables_table_owner.at:1:             env DBUS_SYSTEM_BUS_ADDRESS="unix:abstract=firewalld-testsuite-dbus-system-socket-${at_group_normalized}" ip netns exec fwd-test-${at_group_normalized} firewall-cmd -q --state      
[...]

PR #1477, opened 2025-05-02
fix(fw): start: remove ipset probe

erig0:


erig0:
CI failure is the same as report #1403.

PR #1494, opened 2025-07-10
feat: add policy disable flag

erig0:
feat(cli): policy: add disable flag
    
    This feature allows disabling policies so they do not go active.


n0-1:
Hi Eric,

Since there already are pre-defined zones in firewalld, wouldn't it suffice for
regular users if they behaved "as expected"? So, for instance "internal" would
be allowed to route through the machine and "external" would allow only
established (and perform masquerading)? Then "dmz" would freely route with
"external" but blocked from/to "internal".

Assuming one follows that path, I wonder how many zones there are needed. For
instance, maybe users would want a "external-routed" or even
"external-masqueraded" as well, but such details sound a bit like advanced use
and people may still define their own zones with appropriate policies, right?

Cheers, Phil

SirPhuttel:
> > So, for instance "internal" would
> > be allowed to route through the machine and "external" would allow only
> > established (and perform masquerading)? Then "dmz" would freely route with
> > "external" but blocked from/to "internal".
> 
> Firewalld disallows forwarding between zones unless explicitly enabled (save for the trusted zone). Are you suggesting that we change that principle?

Yes, I do.

> Zones are meant to separate the network - not enforce traffic rules/policies.

Zones are not just network separators, there are expected properties. E.g. the
`trusted` zone is expected to accept all incoming traffic. Including the
forwarding policy in this zone description is consistent wrt. expectations (of
people familiar with zones but unfamiliar with firewalld's interpretation of
them). Take `dmz` for instance: `firewalld.zones(5)` reads:

| For computers in your demilitarized zone that are publicly-accessible with
| limited access to your internal network. Only selected incoming connections are
| accepted.

How are these computers publicly-accessible if firewalld does not allow
forwarding between `external`/`public` and `dmz`?

> You can think of it this way: zones ~= interfaces policies ~= chains
> 
> > but such details sound a bit like advanced use
> > and people may still define their own zones with appropriate policies, right?
> 
> Right. Policy sets are meant to provide a skeleton for specific use cases.
> 
> Policy sets will use the existing zones. Enabling them enables certain behavior that the user wants, e.g. act like a home gateway, which may do things like: masquerade, attach conntrack helpers, etc.

Take such sample user setting eth0 into `internal` zone and eth1 into
`external`. Why shouldn't firewalld accept forwarded traffic from `internal`?

One might also define `dmz` as the only zone which is routed to/from with
`external`/`public`, but it's a matter of definition. This in turn seems a bit
sloppy wrt. zone descriptions. E.g. the difference between `work` and `home` is
not clear, even diffing the XMLs does not help much - `work` does not allow
`samba-client` which in turn might be needed for some offices.

My claim is this: With properly defined zones, standard/novice users will have
enough knobs to get their setup right and firewalld behaviour will be
foreseeable for them. Policy sets just add to the confusion, needlessly
increasing the learning curve for them. Advanced users OTOH will define their
own zones to fine-tune behaviour exactly as they need it. They don't need the
aid policy sets are supposed to provide.

erig0:
> > > So, for instance "internal" would
> > > be allowed to route through the machine and "external" would allow only
> > > established (and perform masquerading)? Then "dmz" would freely route with
> > > "external" but blocked from/to "internal".
> > 
> > 
> > Firewalld disallows forwarding between zones unless explicitly enabled (save for the trusted zone). Are you suggesting that we change that principle?
> 
> Yes, I do.

I think we want something similar. We just disagree about how it should be enabled. I don't think it should be enabled by default. Sets have the advantage that we're not changing behavior for existing users.

It's important to highlight that traffic may be assigned to one and only one zone at ingress/egress. I'm weary about creating a bunch of zones that serve a specific purpose. That feels like we're painting ourselves into a corner.

> > Zones are meant to separate the network - not enforce traffic rules/policies.
> 
> Zones are not just network separators, there are expected properties. E.g. the `trusted` zone is expected to accept all incoming traffic. Including the forwarding policy in this zone description is consistent wrt. expectations (of people familiar with zones but unfamiliar with firewalld's interpretation of them). Take `dmz` for instance: `firewalld.zones(5)` reads:

They are supposed to be just network separators.

> | For computers in your demilitarized zone that are publicly-accessible with | limited access to your internal network. Only selected incoming connections are | accepted.
> 
> How are these computers publicly-accessible if firewalld does not allow forwarding between `external`/`public` and `dmz`?

This is referring to the network itself. Meaning firewalld does not make them publicly accessible. The network is publicly accessible out-of-band of firewalld. INPUT traffic from dmz is considered public traffic.

This description was written long before firewalld supported forward/output filtering. At that point in time firewalld was an end-station only firewall (INPUT).

> > You can think of it this way: zones ~= interfaces policies ~= chains
> > > but such details sound a bit like advanced use
> > > and people may still define their own zones with appropriate policies, right?
> > 
> > 
> > Right. Policy sets are meant to provide a skeleton for specific use cases.
> > Policy sets will use the existing zones. Enabling them enables certain behavior that the user wants, e.g. act like a home gateway, which may do things like: masquerade, attach conntrack helpers, etc.
> 
> Take such sample user setting eth0 into `internal` zone and eth1 into `external`. Why shouldn't firewalld accept forwarded traffic from `internal`?

Why should it? Internal doesn't mean trusted.

The trusted zone allows this by default. That's the only zone that does. But that has historical reasons (to not break podman/docker).

> One might also define `dmz` as the only zone which is routed to/from with `external`/`public`, but it's a matter of definition. This in turn seems a bit sloppy wrt. zone descriptions. E.g. the difference between `work` and `home` is not clear, even diffing the XMLs does not help much - `work` does not allow `samba-client` which in turn might be needed for some offices.

I agree the zones names aren't great nor do they represent what they actually do. Again, that's because they were named long before firewalld supported forward/output filtering.

> My claim is this: With properly defined zones, standard/novice users will have enough knobs to get their setup right and firewalld behaviour will be foreseeable for them. Policy sets just add to the confusion, needlessly increasing the learning curve for them. Advanced users OTOH will define their own zones to fine-tune behaviour exactly as they need it. They don't need the aid policy sets are supposed to provide.

A lot of the things you suggest can't be done by only adding zones. They also needs the accompanying policies to do the filtering/rules. You're still going to have the policies whether we call them sets or not.

Sure we could add more descriptive zones and names. I'm not sure that lessens the confusion though.

Maybe it would help to come up with a list of new zones and what extra things, e.g. policies/rules, would be necessary to make them work as expected. I'm guessing that it ends up looking like a zone plus a set of policies.

erig0:
FWIW, I think the "disable" knob is useful outside of policy sets.

PR #1506, opened 2025-08-27
feat: add gateway policy set

erig0:
This PR introduces the idea of "policy sets". Policy sets are a collection of policies that serve as starting configuration for common use cases, e.g. a home router. This allows the user to get started quickly and then modify the set to their specific environment.

In the future, we could add more sets for other things, e.g. RFCs, firewall standards, etc.

```
FIREWALLD.POLICY-(5)                   firewalld.policy-sets                  FIREWALLD.POLICY-(5)

NAME
       firewalld.policy-sets - Policy Sets

DESCRIPTION
   What Are Policy Sets?
       Policy sets are collections of policies that serve as starting configuration for specific
       use cases, e.g. a home router. They provide an easy way to get started. Users may then fine
       tune the configuration for their environment. Every policy set has a dedicated man page to
       explain its use case.

       All policy sets shipped by firewalld are administratively disabled by default. Using them
       is a matter of adding your interfaces to zones and removing the disable.

   Enabling a Policy Set
       Below is a complete example for using the gateway policy set. In this example: eth0 is the
       LAN interface, and eth1 is the uplink to the internet.

                      # firewall-cmd --permanent --zone internal eth0
                      # firewall-cmd --permanent --zone external eth1
                      # firewall-cmd --permanent --policy-set gateway --remove-disable
                      # firewall-cmd --reload

   Enabling Multiple Policy Sets
       Policy sets are guaranteed to interoperate. Multiple policy sets may be enabled
       simultaneously.

   Existing Policy Sets
       gateway
           Provides basic functionality for a gateway, e.g. a home router. See firewalld.policy-
           set-gateway(5).
```

----

```

IREWALLD.POLICY-(5)                                                                Firewalld Policy Set Gateway                                                                FIREWALLD.POLICY-(5)

NAME
       firewalld.policy-set-gateway - Firewalld Policy Set Gateway

DESCRIPTION
   Policy Set: Gateway
       The Gateway policy set is a useful starting point for a home router. It enables masquerading, conntrack helpers, and forwarding between zones.

       Zones used by this set are logically grouped. These groups name are used by the predefined policies.

                       +-----------+                  +-----------+
                       |    LAN    |                  |   WORLD   |
                       |-----------|                  |-----------|
                       | zones:    |                  | zones:    |
                       |  internal |                  |  external |
                       |  home     |                  |  public   |
                       |  trusted  |                  |           |
                       +-----------+                  +-----------+
                             |                              |
                             |                              |
                             |         +-----------+        |
                             +---------|   HOST    |--------+
                                       |-----------|
                                       | zones:    |
                             +---------|  HOST     |--------+
                             |         +-----------+        |
                             |                              |
                             |                              |
                       +-----------+                  +-----------+
                       |   WORK    |                  |    DMZ    |
                       |-----------|                  |-----------|
                       | zones:    |                  | zones:    |
                       |  work     |                  |  dmz      |
                       +-----------+                  +-----------+
                       

   Policies in the Gateway set
       gateway-dmz-to-HOST
           Enables services commonly needed for a gateway, e.g. dns, dhcp.

           File location: /usr/lib/firewalld/policies/gateway-dmz-to-HOST.xml

       gateway-lan-to-work
           Allows all traffic from LAN to work. Enables masquerading and common connection tracking helpers.

           File location: /usr/lib/firewalld/policies/gateway-lan-to-work.xml

       gateway-lan-to-world
           Allows all traffic from LAN to world. If an interface is added to the "external" zone then the traffic will be masqueraded. Also enables connection tracking helpers for common services,
           e.g. ftp.

           File location: /usr/lib/firewalld/policies/gateway-lan-to-world.xml

       gateway-lan-to-HOST
           Enables services commonly needed for a gateway, e.g. dns, dhcp.

           File location: /usr/lib/firewalld/policies/gateway-lan-to-HOST.xml

       gateway-world-to-HOST
           May be used to expose internal/dmz services to the world by adding a forward port to this policy.

           Here is an example for adding a forward port. It forward port 8080 to 10.1.1.42:80.

               # firewall-cmd --permanent --policy gateway-world-to-HOST \
                              --add-forward-port=port=8080:proto=tcp:toport=80:toaddr=10.1.1.42
               # firewall-cmd --reload

           File location: /usr/lib/firewalld/policies/gateway-world-to-HOST.xml
```

### DIFF
--- a/config/policies/gateway-dmz-to-HOST.xml
+++ b/config/policies/gateway-dmz-to-HOST.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<policy target="CONTINUE" priority="1000">
+  <short>Set Gateway - DMZ to HOST</short>
+  <description>
+    Part of the policy set "gateway". Enables services commonly needed for a
+    gateway, e.g. dns, dhcp.
+  </description>
+  <disable />
+
+  <ingress-zone name="dmz" />
+  <egress-zone name="HOST" />
+
+  <service name="dns" />
+  <service name="dhcp" />
+  <service name="dhcpv6-client"/>
+</policy>
--- a/config/policies/gateway-lan-to-HOST.xml
+++ b/config/policies/gateway-lan-to-HOST.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<policy target="CONTINUE" priority="1000">
+  <short>Set Gateway - LAN to HOST</short>
+  <description>
+    Part of the policy set "gateway". A policy for traffic destined to the node
+    running firewalld and originating from: internal, home, trusted. It
+    enables basic services provided by a gateway, e.g. dns, dhcp.
+  </description>
+  <disable />
+
+  <ingress-zone name="internal" />
+  <ingress-zone name="home" />
+  <ingress-zone name="trusted" />
+  <egress-zone name="HOST" />
+
+  <service name="dns" />
+  <service name="dhcp" />
+  <service name="dhcpv6-client"/>
+  <service name="ssh" />
+</policy>
--- a/config/policies/gateway-lan-to-work.xml
+++ b/config/policies/gateway-lan-to-work.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<policy target="ACCEPT" priority="1000">
+  <short>Set Gateway - LAN to work</short>
+  <description>
+    Part of the policy set "gateway". Allows all traffic from LAN to work.
+    Enables masquerading and common connection tracking helpers.
+  </description>
+  <disable />
+
+  <ingress-zone name="internal" />
+  <ingress-zone name="home" />
+  <ingress-zone name="trusted" />
+  <egress-zone name="work" />
+
+  <!-- Some services that provide conntrack helpers -->
+  <service name="ftp"/>
+  <service name="tftp"/>
+</policy>
--- a/config/policies/gateway-lan-to-world.xml
+++ b/config/policies/gateway-lan-to-world.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<policy target="ACCEPT" priority="1000">
+  <short>Set Gateway - LAN to world</short>
+  <description>
+    Part of the policy set "gateway". Allows all traffic from LAN to world. If
+    an interface is added to the "external" zone then the traffic will be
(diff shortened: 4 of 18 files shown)


erig0:
@TorontoMedia, @mbiebl,
Any thoughts on this? Do you think users will find it useful?

TorontoMedia:
> @TorontoMedia, @mbiebl, Any thoughts on this? Do you think users will find it useful?

I really like the idea, but I’m not a fan of the extra complexity from adding the `set-` prefix. I think it’d make more sense to just explain the policy in the config file and on the website, and have it disabled by default. That keeps things simple and easier to follow.

erig0:
> I really like the idea, but I’m not a fan of the extra complexity from adding the `set-` prefix. I think it’d make more sense to just explain the policy in the config file and on the website, and have it disabled by default. That keeps things simple and easier to follow.

So you're suggesting having the sets/groups of policies, but dropping the "set" prefix?

Advantages to "set-" prefix:
1. visually groups the policies for a use case, e.g. "set-gateway"
  - alternatively, could just use "gateway-foo", "gateway-bar", etc.
  - "set-" provides a prefix/namespace to avoid collisions with user defined policies
2. allows the whole set to be enabled an once
  - we could instead support wildcards, e.g. "gateway-*", "foo*"

erig0:
@TorontoMedia, bump.

erig0:
I made the code changes to allow globbing, e.g. `--policy=gateway*`, but it breaks due to shell globbing if the current directory has files that match, e.g. `gateway-file`. In those instances, the user could _escape_ the asterisk.

Do you feel this is worth the risk/trade-off ?

TorontoMedia:
> Edit: Chose to add `--policy-set` instead of supporting wildcard/asterisk. This is more descriptive and avoids the shell issues.
> 
> @TorontoMedia, I made the code changes to allow globbing, e.g. `--policy=gateway*`, but it breaks due to shell globbing if the current directory has files that match, e.g. `gateway-file`. In those instances, the user could _escape_ the asterisk.
> 
> Do you feel this is worth the risk/trade-off ?
> 
> ```

Apologies for the late reply. Just to clarify, what I meant was that instead of hardcoding the configuration directly into the code, we just include the config file as part of the install similar to ```allow-host-ipv6```. This way, packaging tools like rpmconf can handle config updates and allow the user to review and merge changes if needed. This approach keeps the config easily editable and maintainable while still being included by default in installations.

I also like the ```--policy-set``` command implemented in your pull request.

erig0:
I'm merging this PR.

PR #1523, opened 2025-11-09
Modernize AppStream metadata

AsciiWolf:
Modernize/fix the (outdated) AppStream firewall-config metadata:
- Use a proper rDNS app id / file name format
- Change outdated appdata suffix to metainfo
- Change outdated desktop component type to desktop-application
- Add provides tag for the old app id
- Add missing launchable tag
- Add missing name and summary tags
- Add missing developer tag
- Add missing OARS metadata

Also fix the exec path in desktop files to not hardcode `/usr/bin` since that is a standard path + it won't work inside Flatpak and other special configurations.

/cc @jpopelka


AsciiWolf:
Done. :)

PR #1538, opened 2026-01-28
feat(watcher): log warning if configuration changed

erig0:


erig0:
@todoleza 

New changes:
 - Watcher log only trigger once. Watcher disabled after first occurrence.
 - Only `.xml` and `.conf` files trigger the log.
 - test to verify config watcher re-triggers after `--reload`
 - test to verify log rate limit
 - test to verify `.bak`/etc files don't trigger watcher

todoleza:
@erig0
ACK, LGTM.
Not sure if worth changing: the first change triggers one message or another and any following changes do not trigger the other message. A reload may still be hinted by an out-of-date message while the on-disk configuration state is invalid. Unless you know a clean solution to this, I don't think changes are needed.

PR #1544, opened 2026-03-10
fix(nftables): batch ipset elements in set_restore to avoid O(n²) insertion cost

### LINKED ISSUE #933: firewall-cmd times out with large blocklists
Originally filed as downstream bug report https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006544

Package: firewalld
Version: 0.9.3-2

## Observation

I'm noticing that `firewalld-cmd --reload` crashes when it has to deal
with a large drop.xml file, as shown here:

root@debian:/etc/firewalld/zones# time firewall-cmd --
reload                   
ERROR:dbus.proxies:Introspect error on
:1.28084:/org/fedoraproject/FirewallD1: dbus.exceptions.DBusException:
org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible
causes include: the remote applicat
ion did not send a reply, the message bus security policy blocked the
reply, the reply timeout expired, or the network connection was broken.
Error: Message recipient disconnected from message bus without
replying              
                                                     
real    15m2.633s
user    0m0.273s
sys     0m0.045s

root@debian:/etc/firewalld/zones# systemctl status firewalld
● firewalld.service - firewalld - dynamic firewall daemon       
     Loaded: loaded (/lib/systemd/system/firewalld.service; enabled;
vendor preset: enabled)
     Active: failed (Result: signal) since Sun 2022-02-27 10:46:57 CET;
27min
ago                                                                    
                                                                  
       Docs: man:firewalld(1)
    Process: 33724 ExecStart=/usr/sbin/firewalld --nofork --nopid
(code=killed, signal=KILL)  
   Main PID: 33724 (code=killed, signal=KILL)
        CPU: 15min
20.336s                                                                
                                                                       
                                                      

Feb 27 10:31:34 debian systemd[1]: Starting firewalld - dynamic
firewall daemon...
Feb 27 10:31:35 debian systemd[1]: Started firewalld - dynamic firewall
daemon.
 
Feb 27 10:46:57 debian systemd[1]: firewalld.service: Main process
exited, code=killed, status=9/KILL
Feb 27 10:46:57 debian systemd[1]: firewalld.service: Failed with
result 'signal'.
Feb 27 10:46:57 debian systemd[1]: firewalld.service: Consumed 15min
20.336s CPU time.

## Reproducer

Find attached to this email my drop.xml list. I tested this bug in a
fresh VM running Debian 10 with all installed updates.

* Put attached drop.xml into /etc/firewalld/zones/drop.xml
* Run `firewall-cmd --reload`

## Expected behaviour

The same works on a openSUSE Leap 15.3 VM, where the process took up to
20 minutes to complete.

[drop.xml.txt](https://github.com/firewalld/firewalld/files/8366148/drop.xml.txt)

andrewlukoshko:
## Summary

`set_restore()` currently creates one `{"add": {"element": ...}}` nft operation per ipset entry, then sends them in chunks of 1000 to `set_rules()` via `GLib.idle_add`. On older nftables/kernel combinations (e.g. nftables 1.0.4 / kernel 4.18 shipped in RHEL/AlmaLinux 8), each incremental element insertion into an interval set has O(n) cost proportional to the current set size, making the overall complexity O(n²).

With 12,000 ipset entries this causes `firewall-cmd --reload` to take **~80 seconds** on RHEL 8 with the nftables backend. Even on newer systems (nftables 1.0.9 / kernel 5.14), batching yields a ~20% improvement.

This patch accumulates all element fragments and sends batched `"add element"` operations — one per chunk of 1000 elements — instead of one per entry. This reduces nft operations from N to ceil(N/1000), letting nftables handle bulk element insertion efficiently.

## Benchmark (12k ipset entries, nftables backend)

| | Before | After |
|---|---:|---:|
| RHEL/AL 8 (nftables 1.0.4, kernel 4.18) | ~82,600 ms | ~2,600 ms |
| RHEL/AL 9 (nftables 1.0.9, kernel 5.14) | ~1,560 ms | ~1,240 ms |

## Root cause detail

Profiling `json_cmd` calls inside `set_restore()` showed each successive chunk taking progressively longer on RHEL 8:

```
Chunk  1 (1000 entries): 0.29s
Chunk  6 (1000 entries): 4.70s
Chunk 12 (1000 entries): 15.71s
Total:                   80.54s
```

The raw libnftables API on the same machine can add all 12k elements in 0.14s when done in a single batch, confirming the bottleneck is incremental per-element insertion, not the nftables library itself.

Fixes #933

## Test plan

- Tested on AlmaLinux 8.10 (nftables 1.0.4, kernel 4.18) and AlmaLinux 9.7 (nftables 1.0.9, kernel 5.14) with 12,000 random /24 ipset entries
- Verified all entries load correctly after reload via `firewall-cmd --ipset=<name> --get-entries` and `nft list set`
- Verified empty ipsets and small ipsets (100 entries) work correctly

### DIFF
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -2874,9 +2874,14 @@ class nftables:
                 fragment.append(entry_tokens[i])
         return [{"concat": fragment}] if len(type_format) > 1 else fragment
 
-    def build_set_add_rules(self, name, entry):
+    def build_set_add_rules(self, name, entries):
         rules = []
-        element = self._set_entry_fragment(name, entry)
+        elements = []
+        if not isinstance(entries, (list, tuple)):
+            entries = [entries]
+        for element in entries:
+            elements.extend(self._set_entry_fragment(name, element))
+
         rules.append(
             {
                 "add": {
@@ -2884,7 +2889,7 @@ class nftables:
                         "family": "inet",
                         "table": TABLE_NAME,
                         "name": name,
-                        "elem": element,
+                        "elem": elements,
                     }
                 }
             }
@@ -2942,8 +2947,9 @@ class nftables:
         rules.extend(self.build_set_flush_rules(set_name))
         self.set_rules(rules, self._fw.get_log_denied())
 
-        def _idle_set_add_entries(rules):
+        def _idle_set_add_entries(entries):
             try:
+                rules = self.build_set_add_rules(set_name, entries)
                 self.set_rules(rules, self._fw.get_log_denied())
             except Exception as e:
                 log.error("While restoring ipset entries the following Error occurred:")
@@ -2953,14 +2959,5 @@ class nftables:
         # the entries from the GLib main loop when it's idle. This avoids
         # blocking the main loop for too long.
         #
-        chunk = 0
-        rules = []
-        for entry in entries:
-            rules.extend(self.build_set_add_rules(set_name, entry))
-            chunk += 1
-            if chunk >= 1000:
-                GLib.idle_add(lambda x: _idle_set_add_entries(x), rules)
-                rules = []
-                chunk = 0
-        else:
-            GLib.idle_add(lambda x: _idle_set_add_entries(x), rules)
+        for i in range(0, len(entries), 1000):
+            GLib.idle_add(lambda x: _idle_set_add_entries(x), entries[i : i + 1000])


erig0:
Thank you for the contribution! I made some small changes but kept the idea. I wanted to reuse the existing code to add set entries. Please take a look and retest if you have time. The testsuite has coverage and scale tests for ipsets.

The only downside of this change is that it kind of breaks `IndividualCalls=yes`, but I don't think that's a real concern for ipset entries.

I will merge this, but would like you to take a look at my incremental changes. :)

erig0:
Below is the RHEL-8 backport that I tested. It takes the ipset scale test from 45s to 10s.

--->8----

```diff
commit 7dafd6beec79c6b2b7db22131a31279025743dd6
Author: Andrew Lukoshko <andrew.lukoshko@gmail.com>
Date:   Tue Mar 10 13:30:54 2026 +0100

    fix(nftables): batch ipset elements in set_restore to avoid O(n²) insertion cost
    
    set_restore() currently creates one "add element" nft operation per
    ipset entry, then sends them in chunks of 1000 operations via
    GLib.idle_add.  On older nftables/kernel combinations (e.g. nftables
    1.0.4 / kernel 4.18 shipped in RHEL 8) each incremental element
    insertion into an interval set has O(n) cost proportional to the
    current set size, making the overall complexity O(n²).
    
    With 12,000 ipset entries this causes firewall-cmd --reload to take
    ~80 seconds on RHEL 8 with the nftables backend.  Even on newer
    systems (nftables 1.0.9 / kernel 5.14) batching yields a ~20%
    improvement.
    
    This patch accumulates all element fragments and creates batched
    "add element" operations — one per chunk of 1000 elements — instead
    of one per entry.  This lets nftables handle bulk element insertion
    natively and reduces nft operations from N to ceil(N/1000).
    
    Benchmark (12k ipset entries, nftables backend):
      RHEL 8 (nftables 1.0.4, kernel 4.18): ~82,600ms → ~2,600ms
      RHEL 9 (nftables 1.0.9, kernel 5.14):  ~1,560ms → ~1,240ms
    
    Fixes #933

diff --git a/src/firewall/core/nftables.py b/src/firewall/core/nftables.py
index 19a649aaaa71..022c2961337c 100644
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -1867,14 +1867,19 @@ class nftables(object):
                 fragment.append(entry_tokens[i])
         return [{"concat": fragment}] if len(type_format) > 1 else fragment
 
-    def build_set_add_rules(self, name, entry):
+    def build_set_add_rules(self, name, entries):
         rules = []
-        element = self._set_entry_fragment(name, entry)
+        elements = []
+        if not isinstance(entries, (list, tuple)):
+            entries = [entries]
+        for element in entries:
+            elements.extend(self._set_entry_fragment(name, element))
+
         for family in ["inet", "ip", "ip6"]:
             rules.append({"add": {"element": {"family": family,
                                               "table": TABLE_NAME,
                                               "name": name,
-                                              "elem": element}}})
+                                              "elem": elements}}})
         return rules
 
     def set_add(self, name, entry):
@@ -1923,13 +1928,6 @@ class nftables(object):
         rules.extend(self.build_set_flush_rules(set_name))
 
         # avoid large memory usage by chunking the entries
-        chunk = 0
-        for entry in entries:
-            rules.extend(self.build_set_add_rules(set_name, entry))
-            chunk += 1
-            if chunk >= 1000:
-                self.set_rules(rules, self._fw.get_log_denied())
-                rules.clear()
-                chunk = 0
-        else:
+        for i in range(0, len(entries), 1000):
+            rules.extend(self.build_set_add_rules(set_name, entries[i : i + 1000]))
             self.set_rules(rules, self._fw.get_log_denied())
```

andrewlukoshko:
Thanks for quick reaction @erig0 

In your backport there is missing `rules.clear()` or `rules = []` causes
```
COMMAND_FAILED: 'python-nftables' failed: netlink: Error: Could not process rule: No buffer space available
```
on big ipsets (I tested with 12k).
I tested fixed patch for RHEL8 and it shows same performance:
```
---
 src/firewall/core/nftables.py | 23 ++++++++++++-----------
 1 file changed, 12 insertions(+), 11 deletions(-)

diff --git a/src/firewall/core/nftables.py b/src/firewall/core/nftables.py
index 19a649aaaa71..022c2961337c 100644
--- a/src/firewall/core/nftables.py
+++ b/src/firewall/core/nftables.py
@@ -1867,14 +1867,19 @@ class nftables(object):
         fragment.append(entry_tokens[i])
         return [{"concat": fragment}] if len(type_format) > 1 else fragment
 
-    def build_set_add_rules(self, name, entry):
+    def build_set_add_rules(self, name, entries):
         rules = []
-        element = self._set_entry_fragment(name, entry)
+        elements = []
+        if not isinstance(entries, (list, tuple)):
+            entries = [entries]
+        for element in entries:
+            elements.extend(self._set_entry_fragment(name, element))
+
         for family in ["inet", "ip", "ip6"]:
             rules.append({"add": {"element": {"family": family,
                                               "table": TABLE_NAME,
                                               "name": name,
-                                              "elem": element}}})
+                                              "elem": elements}}})
         return rules
 
     def set_add(self, name, entry):
@@ -1923,13 +1928,9 @@ class nftables(object):
         rules.extend(self.build_set_flush_rules(set_name))
 
         # avoid large memory usage by chunking the entries
-        chunk = 0
-        for entry in entries:
-            rules.extend(self.build_set_add_rules(set_name, entry))
-            chunk += 1
-            if chunk >= 1000:
-                self.set_rules(rules, self._fw.get_log_denied())
-                rules.clear()
-                chunk = 0
-        else:
+        for i in range(0, len(entries), 1000):
+            rules.extend(self.build_set_add_rules(set_name, entries[i : i + 1000]))
+            self.set_rules(rules, self._fw.get_log_denied())
+            rules.clear()
+        else:
             self.set_rules(rules, self._fw.get_log_denied())
-- 
```
Could you please check your updated PR patch with big ipsets just in case?

erig0:
> Could you please check your updated PR patch with big ipsets just in case?

The testsuite uses a set with 62500 entries. The full test completes in 15 seconds, but most of that time is spent on other things (startup, reload, etc) and not applying the ipset.

erig0:
> In your backport there is missing rules.clear()

Indeed. No need for the `for .. else` logic. Just unconditionally call it right after `set_rules()`.

e.g.
```diff
+        for i in range(0, len(entries), 1000):
+            rules.extend(self.build_set_add_rules(set_name, entries[i : i + 1000]))
             self.set_rules(rules, self._fw.get_log_denied())
+            rules.clear()
```

andrewlukoshko:
I tested PR patch on Fedora 44 where it applies cleanly and even survived 500k ipset!
Here are the results:

### Environment

| | |
|---|---|
| OS | Fedora 44 (prerelease) |
| firewalld | 2.4.0-2.fc44 |
| nftables | 1.1.6-2.fc44 |
| kernel | 6.19.6-300.fc44 |
| arch | aarch64 |

### Benchmark — `hash:net` ipset, `firewall-cmd --reload`

| Entries | Before | After | Speedup |
|---:|---:|---:|---:|
| 50,000 | 3,243 ms | 2,667 ms | **~18%** |
| 500,000 | 865,423 ms (~14.4 min) | 850,696 ms (~14.2 min) | **~1.7%** |

### Per-chunk timing (50k entries, 1000 elements/chunk)

Clear O(n) per-chunk growth in both cases, but the patched version is consistently faster per-chunk:

| Chunk | Before (1 op/entry) | After (1 op/1000 entries) |
|---:|---:|---:|
| 1 | 12 ms | 7 ms |
| 10 | 27 ms | 23 ms |
| 25 | 60 ms | 50 ms |
| 40 | 94 ms | 85 ms |
| 50 | 122 ms | 106 ms |

### Per-chunk timing (500k entries)

At extreme scale, batching provides minimal benefit — the nftables kernel interval set insertion cost (O(set_size) per operation) dominates:

| Chunk | Before | After |
|---:|---:|---:|
| ~200 | ~980 ms | ~830 ms |
| ~350 | ~2,500 ms | ~2,300 ms |
| ~500 | ~5,000 ms | ~5,100 ms |

### Edge case tests — all pass

- [x] Empty ipset
- [x] Small ipset (100 entries)
- [x] Single entry add/remove (`set_add` / `set_delete`)
- [x] `hash:ip,port` type (concat fragment path, 500 entries)

I believe it's safe to merge now @erig0  :)
Thank you!

erig0:
Thanks again for the collaboration. Merging!

erig0 (approved):

erig0:
@andrewlukoshko, I pushed this to the `stable-0.9` branch to make your backport easier. Keep in mind that release stream has been very dead for years. I also significantly increased the size of the ipset in the test.

See 60bca542d86364c8053e9324dd4d7f515a30c9d5.

PR #1545, opened 2026-03-12
fix: Drop dependency on dbus-x11

### LINKED ISSUE #1543: Clarify/update/remove dependency on `dbus-x11` for `firewall-config` & `firewall-applet`
### What would you like to be added

`firewall-config` and `firewall-applet` currently require `dbus-x11`:
- https://github.com/firewalld/firewalld/blob/main/firewalld.spec#L74
- https://github.com/firewalld/firewalld/blob/main/firewalld.spec#L88

This was added to fix https://bugzilla.redhat.com/show_bug.cgi?id=1281416 but it's not clear why this is really needed. I could not find any call to dbus-launch in the code.

As we are moving away from dbus & X11 tools in Fedora (to dbus-broker & Wayland), it would be good to update this dependency to dbus-broker (which includes `dbus-broker-launch`) or update it to a recommends if it's not strictly needed in the common case, or remove it entirely.

Thanks

### Why is this needed

We want to remove our dependency on legacy dbus tools and X11 windowing systems.

travier:
This was added to fix [1] but it's not clear why this is really needed. Remove it as firewalld only uses the system bus.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1281416

Fixes: https://github.com/firewalld/firewalld/issues/1543

### DIFF
--- a/contrib/fw-in-container/fw-in-container
+++ b/contrib/fw-in-container/fw-in-container
@@ -364,7 +364,6 @@ RUN dnf install -y \\
     cscope \\
     dbus-broker \\
     dbus-devel \\
-    dbus-x11 \\
     desktop-file-utils \\
     dhclient \\
     dhcp-client \\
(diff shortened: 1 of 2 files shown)


erig0 (approved):
lgtm. Thanks!

Issue #1353, opened 2024-06-13
firewall-config brings up INVALID_ZONE dialog for combined zones


AndersBlomdell:
### What happened

When defining combined zones, i.e.
```
  /etc/firewalld/zones/myzone/01_rules.xml
  /etc/firewalld/zones/myzone/02_rules.xml
```
firewall-config brings up `INVALID_ZONE: myzone` dialog. The reason for this is that in `src/firewall/server/config.py` `getZoneByName` never matches `myzone`, since `obj.obj.name` are `myzone/02_rules`  and `myzone/02_rules` which 
doesn't match `myzone`. A simple (but obviously wrong) fix to this is adding:
```
            elif obj.obj.name.split('/')[0] == zone:
                # Quick and dirty, need to combine all parts somewhere...
                return obj
```

### Firewalld Version

firewalld-2.1.2-2.fc40.noarch

### Firewalld Backend

nftables

### Linux distribution

Fedora Linux 40

### Linux kernel version

6.8.7-300.fc40.x86_64

AndersBlomdell:
Added pull request for this issue https://github.com/firewalld/firewalld/pull/1550

erig0:
> firewall-config brings up INVALID_ZONE: myzone dialog. The reason for this is that in src/firewall/server/config.py getZoneByName never matches myzone, since obj.obj.name are myzone/02_rules and myzone/02_rules which
doesn't match myzone. A simple (but obviously wrong) fix to this is adding:

That's how it's suppose to be. In the runtime configuration, the zone pieces are combined into ONE object. In the permanent configuration they are kept separate otherwise you wouldn't be able to configure them separately on-disk.

This is shown in this CLI test:
https://github.com/firewalld/firewalld/blob/6cdb99268335b57ffd6028480e4d591badab6c60/src/tests/features/zone_combine.at

I suspect this is actually a firewall-config UI bug. `_load_store_append()` is suspicious as it always calls `getZoneByName()` even though it may be passed runtime zone names. I would guess it broke with commit `d81485cdf4c1 ("improvement(firewall-config): show zone description as tooltip")`. You could revert that commit to verify.

Issue #1486, opened 2025-06-09
Inconsistent and undocumented behavior when zones folder contain folder or filename with dot


ericwong3:
### What happened

When /etc/firewalld/zones contains folder, XML filenames with dot, or folder with dot in name, --check-config and --reload will behave strangely and return misleading message.

It appears firewalld attempts to take the folder name as the zone name, but validation is inconsistent.

**Base Good Case: XML with dot (e.g., public.1.xml)**
- `firewall-cmd --check-config` returns: "Error: INVALID_NAME: '.' is not allowed in 'public.1'" 
- `firewall-offline-cmd --check-config` returns: "Configuration error: INVALID_NAME: '.' is not allowed in 'public.1'" 
- `firewall-cmd --reload` returns: "Error: INVALID_NAME: '.' is not allowed in 'public.a'" 

**Test Case 1: Folder with dot, XML without dot (e.g., 250610.backup/public2.xml)**
- `firewall-cmd --check-config` returns: "success" ❌
- `firewall-offline-cmd --check-config` returns: "Configuration error: INVALID_NAME: '.' is not allowed in '250610.backup/public2'" 
- `firewall-cmd --reload` returns: "Error: INVALID_ZONE: public" 
Expectation: Load as public2 or do not load (ignore subfolder).

**Test Case 2: Folder without dot, XML with dot (e.g., archive/public.3.xml)**
- `firewall-cmd --check-config` returns: "success" ❌
- `firewall-offline-cmd --check-config` returns: "Configuration error: INVALID_NAME: '.' is not allowed in 'archive/public.3'" 
- `firewall-cmd --reload` returns: Error: INVALID_ZONE: public 
Expectation: Prompt error for "public.3" name or do not load (ignore subfolder).

**Test Case 3: Folder without dot, XML without dot (e.g., archive/public4.xml)**
- `firewall-cmd --check-config` returns: "success"
- `firewall-offline-cmd --check-config` returns: (Empty)
- `firewall-cmd --reload` returns: "success" => But zone is loaded as archive
Expectation: Load as public4 or do not load (ignore subfolder).

### What you expected to happen

The two ❌ cases should be fixed to return error and as subsequent reload will in fact return error.

[...]

PR #1550, opened 2026-04-13
Refactor zone combining code

AndersBlomdell:
core/fw_config.py:
        move combined zone logic to 'add_zone' method
core/fw.py:
        in '_start_copy_config_to_runtime' don't copy zones with '/' in name
        remove combined zone logic from '_loader_zones'
        use 'list_zone_files' to find what files to load
core/io/functions.py:
        add 'list_xml_files' function
        add 'list_zone_files' function
        rewrite 'check_on_disk_config' to use 'list_xml_files' and
                'list_zone_files'
server/config.py:
        don't return combined zones in 'getZoneNames'

### DIFF
--- a/src/firewall-offline-cmd.in
+++ b/src/firewall-offline-cmd.in
@@ -1659,7 +1659,7 @@ try:
 
     # zones
     elif a.get_zones:
-        zones = fw.config.get_zones()
+        zones = fw.config.get_zones(select=lambda z: not z.combined)
         cmd.print_and_exit(" ".join(zones))
 
     elif a.new_zone:
@@ -1696,7 +1696,7 @@ try:
         fw.config.load_zone_defaults(obj)
 
     elif a.info_zone:
-        zone = fw.config.get_zone(a.info_zone)
+        zone = fw.config.get_zone(a.info_zone, select=lambda z: not z.combined)
         settings = FirewallClientZoneSettings(fw.config.get_zone_config_dict(zone))
         cmd.print_zone_info(a.info_zone, settings, default_zone=fw.get_default_zone())
         sys.exit(0)
--- a/src/firewall/core/fw.py
+++ b/src/firewall/core/fw.py
@@ -34,12 +34,12 @@ from firewall.core.io.firewalld_conf import firewalld_conf
 from firewall.core.io.direct import Direct
 from firewall.core.io.service import service_reader
 from firewall.core.io.icmptype import icmptype_reader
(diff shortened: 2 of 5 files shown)


AndersBlomdell:
Fixes issue https://github.com/firewalld/firewalld/issues/1353

AndersBlomdell:
Would be nice if 'black' was run with '--diff', since black does not signal any errors on my machine.

erig0:
I don't follow this change. There is [testsuite coverage for combined zones for the CLI](https://github.com/firewalld/firewalld/blob/6cdb99268335b57ffd6028480e4d591badab6c60/src/tests/features/zone_combine.at). getZoneNames() is not returning the combined zone. It returns the individual pieces.

Are we sure this isn't actually a UI bug in firewall-config? e.g. it's getting a runtime zone list and trying to load permanent configuration based off that?

AndersBlomdell:
I made the decision to move the combining code, since combined zones were broken with profiles as well, and my somewhat largish fix made more sense to me (the zone combining code has been obscure since the beginning).

AndersBlomdell:
> I don't follow this change. There is [testsuite coverage for combined zones for the CLI](https://github.com/firewalld/firewalld/blob/6cdb99268335b57ffd6028480e4d591badab6c60/src/tests/features/zone_combine.at). getZoneNames() is not returning the combined zone. It returns the individual pieces.
> 
> Are we sure this isn't actually a UI bug in firewall-config? e.g. it's getting a runtime zone list and trying to load permanent configuration based off that?

Actually it gets a zone name set in NetworkManager and tries to find that zone (and hence reports an error).

erig0:
> Actually it gets a zone name set in NetworkManager and tries to find that zone (and hence reports an error).

NM only works against the runtime configuration. [See my comment in the Issue](https://github.com/firewalld/firewalld/issues/1353#issuecomment-4432244213). I think this is a bug in `_load_store_append()`.

erig0:
> > Actually it gets a zone name set in NetworkManager and tries to find that zone (and hence reports an error).
> 
> NM only works against the runtime configuration. [See my comment in the Issue](https://github.com/firewalld/firewalld/issues/1353#issuecomment-4432244213). I think this is a bug in `_load_store_append()`.

Confirmed. See #1559. I'll merge it ASAP.

erig0:
Hey @AndersBlomdell , Thanks for the PR! Sorry I didn't merge it. I chose to #1353 and #1486 in a different way.

PR #1563, opened 2026-05-15
fix: use config.VERSION for --version in firewall-cmd

### LINKED ISSUE #1562: firewall{,-offline}-cmd --version can only be executed by root
### Discussed in https://github.com/firewalld/firewalld/discussions/1412

<div type='discussions-op-text'>

<sup>Originally posted by **unknowndevQwQ** October 26, 2024</sup>
I don't think this is a reasonable restriction -- the `--help` option doesn't require root. If there is any special safety reason, please explain in detail.</div>

blut-agent:
## Summary

Fixes `firewall-cmd --version` requiring root access, which is inconsistent with `firewall-cmd --help` (no root needed) and `firewall-offline-cmd --version` (also no root needed).

The version string is a static compile-time constant (`config.VERSION`) available at build time. The current code calls `fw.get_property("version")` which requires a D-Bus connection to the firewalld daemon, which in turn requires root privileges.

This change makes `firewall-cmd --version` use `config.VERSION` directly, matching the approach already used by `firewall-offline-cmd`.

## Related

Closes #1562

### DIFF
--- a/src/firewall-cmd.in
+++ b/src/firewall-cmd.in
@@ -25,6 +25,7 @@ from firewall.client import (
     FirewallClientPolicySettings,
 )
 from firewall.errors import FirewallError
+from firewall import config
 from firewall import errors
 from firewall.functions import joinArgs, splitArgs, getPortRange
 from firewall.core.fw_nm import (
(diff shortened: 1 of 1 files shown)


erig0 (approved):
Thanks, lgtm.

unknowndevQwQ:
Unfortunately... the `--version` option in `offline-cmd` also requires root access, which I don't understand.

erig0:
My guess is that it's because the `Firewall()` object is created and `start()` is called which will read all the config files - which require root.

The fix would be to move the `a.version` handling to occur before this.

In fact, I bet `firewall-cmd` also has issues printing the version if the daemon is not running.

PR #1567, opened 2026-06-04
Add missing policy xml schema to package

Timdawson264:
Currently policy.xsd is not getting shipped in RPMs as its missing from the distribution.

Pretty please could this be backported into 2.3 and 2.4. aka fedora 43/44


erig0 (approved):
Thanks!

FYI. I tweaked the commit message.

PR #1576, opened 2026-07-02
feat(client): support automatic dark theme

1van1ka:
This change makes the firewall-config follow the GNOME desktop color scheme preference.
The theme preference is also applied during application startup so the initial appearance matches the current system setting.


erig0 (approved):
Thank you!!

I verified that this works as expected on RHEL-10. I'll merge after CI.

PR #1584, opened 2026-07-19
Tilde expansion support for --add-entries-from-file

olimah:
The following patches make firewall-cmd honor bash tilde expansion for the --add-entries-from-file option, so that one can run:

sudo firewall-cmd [--permanent] --ipset=ipset --add-entries-from-file=~/new-ips.txt

or even:

sudo firewall-cmd [--permanent] --ipset=ipset --add-entries-from-file=~user/new-ips.txt

### DIFF
--- a/src/firewall-cmd.in
+++ b/src/firewall-cmd.in
@@ -1676,6 +1676,7 @@ if a.permanent:
 
             for filename in a.add_entries_from_file:
                 try:
+                    filename = os.path.expanduser(filename)
                     entries = cmd.get_ipset_entries_from_file(filename)
                 except IOError as msg:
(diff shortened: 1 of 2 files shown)


erig0:
Hrm. I don't know. This only fixes `--add-entries-from-file` and doesn't address any of the other commands.

This works if you omit the `=` and let the shell do the expansion. This doesn't require changes in firewalld.

```
[root@rhel9-44819 01]# firewall-cmd --ipset foobar --add-entries-from-file=~/foo.txt
Failed to read file '~/foo.txt': [Errno 2] No such file or directory: '~/foo.txt'

[root@rhel9-44819 01]# firewall-cmd --ipset foobar --add-entries-from-file ~/foo.txt
success
```

PR #1602, opened 2026-09-06
docs(richlanguage): document burst attribute for limit

fts-fahmi:
The rich-rule `limit` element already accepts an optional `burst` attribute, but `firewalld.richlanguage(5)` only documented `limit value="rate/duration"`.

This documents `burst` in the Limit section, the related syntax lines, and one example.

Resolves: https://issues.redhat.com/browse/RHEL-50573

Verified on RHEL 8 (`firewalld-0.9.11-10.el8_10`):

```
$ man firewalld.richlanguage | grep -i burst
(no output)

$ firewall-cmd --add-rich-rule='rule family="ipv4" source address="192.0.2.1" limit value="1/m" burst="5" drop'
success

$ firewall-cmd --list-rich-rules
rule family="ipv4" source address="192.0.2.1" drop limit value="1/m" burst=5

$ firewall-cmd --remove-rich-rule='rule family="ipv4" source address="192.0.2.1" limit value="1/m" burst="5" drop'
success
```

Made with [Cursor](https://cursor.com)


doc/xml/firewalld.richlanguage.xml
```diff
@@ -375,6 +375,8 @@ limit value="rate/duration"
 	 The rate is a natural positive number [1, ..]
 	 The duration is of "s", "m", "h", "d". "s" means seconds, "m" minutes, "h" hours and "d" days.
 	 Maximum limit value is "2/d", which means at maximum two matches per day.
+	 The optional burst is a non-negative integer in the range 0 to 10000000.
```
erig0, line 378: 0 is invalid. It must be > 0.

Should also mention that the default value is dependent on the firewall backend. For nftables, value=5 at the time of writing.

doc/xml/firewalld.richlanguage.xml
```diff
@@ -375,6 +375,8 @@ limit value="rate/duration"
...
+	 It sets how many packets may match in excess of the given rate.
```
erig0, line 379: This is not totally accurate. For packets, `burst` defines the bucket size. `rate/duration` defines the _drain_ rate of the bucket. It's basically the leaky bucket algorithm: https://en.wikipedia.org/wiki/Leaky_bucket

Perhaps this phrasing is better.

> It defines the bucket size for accepted packets. Once the bucket is full the rule will not match until the bucket drains based on `rate/duration`.

erig0 (changes requested):

erig0 (changes requested):
Please also include these tags in your commit message:

```
Fixes: RHEL-50573
Fixes: 58dfdcafabaa ("improvement(rich): support "burst" attribute to limit in rich rules")
```

Issue #194, opened 2017-01-10
RFI: Please add comments to ipset entries


TriMoon:
It would be **much** appreciated if the firewalld UI, in particular (and possibly ``firewall-cmd``),  would allow us to add comments to entries in ipsets.

leonbloy:
Related https://serverfault.com/questions/893112/migrating-from-iptables-to-firewalld-commenting-rules

rozmansi:
I too am planning a migration from _/etc/sysconfig/iptables_ to FirewallD and would love the support for commenting: entries in ipsets, particular rich rules, etc.

Could `firewall-cmd` be upgraded to just copy the `<!-- -->` elements when updating the XML config files?

erig0:
> Could firewall-cmd be upgraded to just copy the  elements when updating the XML config files?

Possibly. I think it would be useful for all the XML files.

In the mean time you could use the ipset description to stash metadata. e.g.

    # firewall-cmd --permanent --ipset=foobar --set-description="10.1.1.0/24: lab -- 10.1.2.0/24: office"

```
# cat /etc/firewalld/ipsets/foobar.xml 
<?xml version="1.0" encoding="utf-8"?>
<ipset type="hash:ip">
  <description>10.1.1.0/24: lab -- 10.1.2.0/24: office</description>
  <entry>10.1.1.0/24</entry>
  <entry>10.1.2.0/24</entry>
</ipset>
```

FrenchFranck:
Accord the change log history, since V6.20 the comment are supported for each entry 
ipset create foo hash:ip comment
ipset add foo 192.168.1.1/24 comment "This is my comment 2"
ipset add foo 192.168.1.2/24 comment "This is my comment 2"

But some Linux distribution does not support this option (see https://bugzilla.redhat.com/show_bug.cgi?id=1496859)
On my Fedora 28 its ok, but with an error message appear whitout importance.

Use description is not sufficient. As SpareSimian suggested, it will be a good idea to define the comment as an attribute of each entry. Like this :
.......
<entry comment="This is my comment 1">192.168.1.1</entry>
<entry comment="This is my comment 2">192.168.10.5</entry>
......


But how to specify the option to indicate the ipset have to use the comment 
 ipset create foo hash:ip **comment**

judchuks:
www. aidenpromotions. com/articles

That is one way in which you could drop your articles online for free without charge. Go there and post your articles without paying a dime!

TriMoon:
[...]

PR #1603, opened 2026-09-06
feat(ipsets): support maintaining comments when reading/writing config xml

ArthurRM:
Addresses: https://github.com/firewalld/firewalld/issues/194

extended `core/io/io_object` code to support basic reading and writing of comments; and a dictionary wrapper to maintain comments linked to a corresponding element
extended `core/io/ipset` to process any read comments and include them back when writing

Supported with this change is what I would call _preceding_ and _closing_ comments. E.g.:

```
<ipset type="...">
  <!-- comment some option -->
  <option name="..." value="..."/>
  ...
  <!-- multi line
          comment
          for entry
  -->
  <!-- second comment for same entry -->
  <entry>...</entry>
  ...
  <!-- closing comment ipset -->
</ipset>

```
multiple comments preceding an element are supported; but when removing the element _all_ comments directly preceding it will be removed as well. Because in the chosen approach comments are always 'linked' to an element.

Not supported is what I would call _trailing_ comments. E.g.:

```
<ipset type="...">
  ...
  <entry>...</entry><!-- trailing comment not supported -->
</ipset>

```
the reason this is not supported is because the used XML reader by default just 'encounters' comments between the start and/or end of elements. So when between elements I have chosen to interpret it as preceding the next element, as I think that is much more common. 
(in theory it is possible to also support trailing comments but that would add substantial further complexity to the content handler)

Final notes: 
- if this PR were to be approved could add processing of comments in other places than ipsets (if/where desired); but let's first see if we can agree on the general approach
- this PR maintains existing comments; I have a [separate PR ready](https://github.com/ArthurRM/firewalld/pull/1/changes) for a `--with-comment` option that can be used together with `--add-entry` in `firewall-cmd` and `firewall-offline-cmd`

### DIFF
--- a/src/firewall/client.py
+++ b/src/firewall/client.py
@@ -1740,7 +1740,7 @@ class FirewallClientIPSetSettings:
         if settings:
             self.settings = settings
         else:
-            self.settings = ["", "", "", "", {}, []]
+            self.settings = ["", "", "", "", {}, [], {}]
 
     @handle_exceptions
     def __repr__(self):
--- a/src/firewall/core/io/io_object.py
+++ b/src/firewall/core/io/io_object.py
@@ -211,6 +211,41 @@ class IO_Object:
             )
 
 
+class IO_Object_CommentsDict(dict):
+    def __comment_key(self, el_name, el_id=""):
+        return "%s%s%s" % (el_name, "/" if el_id != "" else "", el_id)
+
+    def set_comments(self, el_name, el_id, preceding_comments, closing_comments):
+        key = self.__comment_key(el_name, el_id)
+        if not key in self:
+            self[key] = (preceding_comments, closing_comments)
+        else:
+            if not self[key][0]:
+                self[key][0] = preceding_comments
+            if not self[key][1]:
+                self[key][1] = closing_comments
+
+    def set_preceding_comments(self, el_name, el_id, preceding_comments):
+        key = self.__comment_key(el_name, el_id)
+        if not key in self:
+            self[key] = (preceding_comments, [])
+        elif not self[key][0]:
+            self[key][0] = preceding_comments
+
+    def set_closing_comments(self, el_name, el_id, closing_comments):
+        key = self.__comment_key(el_name, el_id)
+        if not key in self:
+            self[key] = ([], closing_comments)
+        elif not self[key][1]:
+            self[key][1] = closing_comments
+
+    def get_comments(self, cmt_pos, el_name, el_id=""):
+        key = self.__comment_key(el_name, el_id)
+        if key in self and self[key][cmt_pos]:
+            return self[key][cmt_pos]
+        return []
+
+
 # PARSER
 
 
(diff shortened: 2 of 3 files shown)


L34T:
Could you clarify how a few edge cases are handled?

1.
```python
    def comment(self, comment):
        self._write("<!-- " + comment + " -->")
```
What happens if a comment contains `--` (e.g. CLI flags or separators)?
Since `--` is illegal inside XML comments, will the written XML remain valid and parseable?

2.
```xml
<!-- comment for both entries -->
<entry>192.0.2.1</entry>
<entry>192.0.2.2</entry>
```
If we remove the first entry (`firewall-cmd --permanent --ipset=test --remove-entry=192.0.2.1`):
```xml
<entry>192.0.2.2</entry>
```
Will the comment disappear completely for the remaining `192.0.2.2`?

3.
If an entry is removed and later re-added to the same ipset:
1. `remove-entry 192.0.2.1` (the key remains in `self.comments`).
2. `add-entry 192.0.2.1` later for a different purpose.

Will it unintentionally restore the old comment from memory?

4.
```xml
  <!--
      Line 1
      Line 2
  -->
  <entry>192.0.2.3</entry>
```
With `data[1:]` / `data[:-1]` trimming, what happens to indentation and newlines across save cycles?
```xml
  <!--       Line 1
      Line 2  -->
```

5.
Will changing DBUS_SIGNATURE be a breaking change?

upd
6.
```python
    def set_preceding_comments(self, el_name, el_id, preceding_comments):
        key = self.__comment_key(el_name, el_id)
        if not key in self:
            self[key] = (preceding_comments, [])
        elif not self[key][0]:
            self[key][0] = preceding_comments
```
Since `self[key]` is stored as a tuple `(preceding, closing)`, won't modifying `self[key][0]` on an existing key raise a runtime ```TypeError ('tuple' object does not support item assignment)```?

ArthurRM:
> What happens if a comment contains `--` (e.g. CLI flags or separators)? Since `--` is illegal inside XML comments, will the written XML remain valid and parseable?
> 
With the current PR comments only get read from existing XML config. So there shouldn't be any illegal characters inside, otherwise the XML was already broken (and not parsable).

But, if in a subsequent step it would be allowed to provide a comment via the cmd, then it should indeed be escaped.


> If we remove the first entry (`firewall-cmd --permanent --ipset=test --remove-entry=192.0.2.1`):
> ...
> Will the comment disappear completely for the remaining `192.0.2.2`?
> 
In the current approach yes. The logic views comments as linked to an element. Since config files can also be edited via external tools I think there is no way to know for sure if a comment is meant for the next element or for the next series of elements (or even for a prior element for that matter).
Basically a choice has to be made on what would be the supported way of commenting (if any). 
The alternative here would be assuming a comment is meant for all subsequent elements until a new comment is encountered. But that would force users to have to provide comments for everything, which to me seemed a greater drawback. (I may be biased by how I personally apply comments of course)


> If an entry is removed and later re-added to the same ipset:
> ...
> Will it unintentionally restore the old comment from memory?
> 
Good point; and not entirely sure. Would have to study the code some more for this. I'm a long-time user of firewalld but the code is new to me. 
If the ipset data is indeed maintained in memory by the service for as long as it runs then you are probably right. Though it would be a relatively small change to fix that then.


> With `data[1:]` / `data[:-1]` trimming, what happens to indentation and newlines across save cycles?
> 
Yeah it was an attempt to somehow at least always have a nice opening and closing space before/after each comment; but you are right that in case of a newline or tab as very first/last character this will go wrong. Will remove this; and instead only add spaces when a comment gets added via the cmd.


> Will changing DBUS_SIGNATURE be a breaking change?
>
Know too little about dbus to judge I'm afraid..


> Since `self[key]` is stored as a tuple `(preceding, closing)`, won't modifying `self[key][0]` on an existing key raise a runtime `TypeError ('tuple' object does not support item assignment)`?
> 
Strangely I didn't encounter this while testing, but yes. Need to rewrite that to update tuple as a whole.

